PR #2, opened 2020-05-16
Define keymaps declaratively

a13:
This pull request allows action keymaps to be described declaratively with defcustoms, so it's much easier to define/customize them without changing the package source code (and easier to test/refactor).
The only downside is it should be done before package loading `use-package :custom`, which I use for configuration should work well, though. Maybe we should introduce something like `embark-regenerate-keymaps` interactive command for that, or make embark-keymap-alist accept functional values (besides keymaps).

P.S. I have an error message in Emacs 26.3
 ```
minibuffer-force-complete called with 3 arguments, but accepts only 0-2
```
It looks like this behaviour is the same for the trunk.

### CODE REVIEW

**Comment by oantolin** [comment]
I don't really like this idea. Keymaps are the standard interface for, well, binding keys to functions. If there is a standard Emacs interface for something, I'd rather use it than invent a new one. I think modifying keymaps is a solved problem in Emacs: solved in many different ways and everybody has their favorite solution: `define-key`, `bind-keys`, `use-package :bind`, `general.el`, etc.!

If we changed from keymaps to alists, somebody would file an issue [like this one](https://github.com/raxod502/selectrum/issues/71) rightly pointing out that keymaps are the standard way to associate keys to bindings in Emacs!

Also, are you saying that:

```elisp
(use-package embark
  :bind
  (:map minibuffer-local-completon-map
        ("M-o" . embark-act))
  :custom
  (embark-buffer-map-alist
   '(("k" . kill-buffer)
     ("s" . switch-to-buffer))))
```

is better and "more declarative" than this?

```elisp
(use-package embark
  :bind
  (:map minibuffer-local-completon-map
        ("M-o" . embark-act))
  (:map embark-buffer-map
        ("k" . kill-buffer)
        ("s" . switch-to-buffer)))
```

I'm not convinced at all!

Notice that in the alist method, besides the annoying downsides you already mentioned (needing to customize before loading, needing a "regenerate keymaps command"), when using `:custom` you'd have to give the entire alist. But the keymap way is more flexible: you can just list the differences. (In the example above I could remove the kill-buffer binding in the keymap method but not in the alist method.)

I'm sorry, but I don't want to merge this, I don't see any advantage to alists.

oantolin:
I'll look into the `minibuffer-force-complete` error. I normally rely on `package-lint-current-buffer` to tell me what version of Emacs is needed for my code, but it didn't spot this.

oantolin:
Now that I reread my comment, I feel I may have worded it too strongly. I'm very glad you wanted to help improve this package, and appreciate the effort that went into this PR, I just disagree that your suggestion really improved things. I may simply not be understanding the benefits and would be willing to discuss them some more, if you want to explain the benefits to me.

oantolin:
Also, I think I've fixed the `minibuffer-force-complete` error on older Emacsen in commit 63ffb11. Please test.

a13:
> Also, are you saying that … is better and "more declarative" than this?

No.

I'm saying that 

```
  '(("d" . delete-file)
    ("r" . rename-file)
    ("c" . copy-file))
```

is more declarative than

```
    (set-keymap-parent map embark-general-map)
    (define-key map "d" #'delete-file)
    (define-key map "r" #'rename-file)
    (define-key map "c" #'copy-file)
    map))
```    

The main point of this PR is not to invent a new way of configuring keymaps (it's just a side effect we get for free) but to get rid of (almost duplicate) imperative code. These alists are not trying to substitute keymaps (I prefer standard solutions too), but complement them. It's even possible to make them defvars or even defconsts, if you want to. After all, `embark--generate-map` is just a helper function which makes the code simpler and easier to support.

a13:
an example, where an approach like that gives significant simplification of code:
https://github.com/a13/point-im.el/blob/master/point-im.el#L443

I'm really tired of getting manually define-key dozens of keys, and copy/paste is a source of many errors

oantolin:
But the code you want to get rid of is inside the source code of this package where users never have to see it. If users want to configure the keymaps, they can use perfectly declarative code in their configuration, as I showed above, no? Are you thinking about a use case where people isntall this package and then start modifying the source code?

I think if you are just worried about the aesthetics of the source code (which users don't actually need to see), we could make a helper function but I would rather use it inline in the definition of the keymaps. That is I want to expose the keymaps to the user so they can use `general.el` of whatever to configure them. Given that I wouldn't want to _also_ expose a redudant alist variable _and_ have to explain that it's OK to modify it but you have to remember to do it before loading the package or to call a regenerate keymaps function.

I also wouldn't worry too much about the aesthetics of the source code: it uses the standard Emacs idiom for defining keymaps (take a look at any major mode that ships with Emacs).

oantolin:
How about adding the helper function but not making any alist defvars and instead doing something like:

```elisp
(defcustom embark-buffer-map
  (embark--keymap
   '(("k" . kill-buffer)
 ...)
```

That way we get the main benefit you are after, but without exposing a redundant variable with some restrictions on its use?

oantolin:
I just want to point out I edited the previous comment a bunch of times, in case you already read it when you got the notification.

a13:
> But the code you want to get rid of is inside the source code of this package where users never have to see it. If users want to configure the keymaps, they can use perfectly declarative code in their configuration, as I showed above, no?

Of course! As I said before, I usually prefer more or less standard solutions

> Are you thinking about a use case where people install this package and then start modifying the source code?

Yes. See above.

> I think if you are just worried about the aesthetics of the source code (which users don't actually need to see), we could make a helper function but I would rather use it inline in the definition of the keymaps.

Could you give an example, please?

> That is I want to expose the keymaps to the user so they can use general.el of whatever to configure them. 

agreed, there's absolutely no way to reinvent new wheels, the old ones aren't perfect, of course, but since we have instruments (like `:bind`) to fix - it's ok

> Given that I wouldn't want to also expose a redudant alist variable and have to explain that it's OK to modify it but you have to remember to do it before loading the package or to call a regenerate keymaps function.

I agree now. Having "private" `embark--*` vars is a better solution, than defcustoms.

> I also wouldn't worry too much about the aesthetics of the source code: it uses the standard Emacs idiom for defining keymaps

It's not about aesthetics (well, partly), I'm trying to make it more error-proof and simple.

> (take a look at any major mode that ships with Emacs).

I know, I know. But I treat `define-key` stuff like a low-level construction, we have to use it under the hood of course, but there's absolutely no reason to write everything by hands, we are using Lisp, after all.

oantolin:
> > we could make a helper function but I would rather use it inline in the definition of the keymaps.
> 
> Could you give an example, please?

I meant like in this [other comment](https://github.com/oantolin/embark/pull/2#issuecomment-629674748). But your other proposal of a private defvar is probably equally good. (I'm not too fond of a private variable or constant that only gets used in *one* place, I tend to inline the definition and if I later do want the variable in more than one place _then_ I refactor ---but I don't feel strongly about this and if you prefer a private variable that's OK.)

a13:
> How about adding the helper function but not making any alist defvars and instead doing something like:

Almost perfect, Mr.!

> (I'd rather rename the imperative embark--generate-keymap to the declarative embark--keymap,

+

> and make setting the parent to embark-general-map the default.

another thing I was thinking about is being able to chain `embark--keymap` (let's use the new name) calls

> We could even decide to export the helper function, for users that want to define other categories of action and need to define corresponding keymaps.)

+

> That way we get the main benefit you are after (easier maintenance), but without exposing a redundant variable with some tricky restrictions on its use. Even if the alists didn't have those rules regarding their use, I don't like having two different ways for users to configure the same thing.

+

>I meant like in this other comment. But your other proposal of a private defvar is probably equally good. (I'm not too fond of a private variable or constant that only gets used in one place, I tend to inline the definition and if I later do want the variable in more than one place then I refactor
> ---but I don't feel strongly about this and if you prefer a private variable that's OK.)

Me neither, I think inline definition would be ok here and we don't have to introduce new intities.


Great! I'm going to refresh the PR in some minutes! Thanks for the collaboration!

oantolin:
> Great! I'm going to refresh the PR in some minutes!

Excellent, thanks!

>  Thanks for the collaboration!

It's been good! (And again, sorry if I came off too opinionated at first, I'm really not that stubborn.)

oantolin:
Also, does my other change (63ffb11) intended to avoid minibuffer-force-complete work in your version of Emacs? I think it should but I'm on a  very space-limited Chromebook and don't want to install an older Emacs to test.

a13:
Updated, I'm still not sure about making `embark-general-map` the default, see the comment about chaining above, besides, we have to define `embark-general-map` before `embark-keymap` in that case. I'm ok with that but still prefer the current solution.

a13:
> It's been good! (And again, sorry if I came off too opinionated at first, I'm really not that stubborn.)

I'm a person who doesn't really care about the language, what makes sense is actual information which you are trying to tell. So, no worries about that.

a13:
> Also, does my other change ([63ffb11](https://github.com/oantolin/embark/commit/63ffb11fe810e2f470acacab77a3aadca38c543b)) intended to avoid minibuffer-force-complete work in your version of Emacs? I think it should but I'm on a very space-limited Chromebook and don't want to install an older Emacs to test.

looks like it works, thanks

The only thing is flycheck saying that 

`the function ‘with-minibuffer-selected-window’ is not known to be defined.`

a13:
A couple of questions more:

— would you like me to add PR-related info to README too?
— do you prefer commits to be squashed into one?

oantolin:
> the function ‘with-minibuffer-selected-window’ is not known to be defined.

Indeed it wasn't back then! It's a trivial convenience macro, I avoid it in commit 07adc6a.

> Updated, I'm still not sure about making embark-general-map the default, see the comment about chaining above, besides, we have to define embark-general-map before embark-keymap in that case. I'm ok with that but still prefer the current solution.

Yes, I was going to say the same thing, that upon further reflection and to make the function more generally useful, it's better to not make the parent `embark-general-map` by default.

> I'm a person who doesn't really care about the language, what makes sense is actual information which you are trying to tell. So, no worries about that.

Excellent! (No harm in me trying to be politer anyway, though.)

> — would you like me to add PR-related info to README too?

That would be amazing, thanks!

> — do you prefer commits to be squashed into one?

I'm not a sophisticated git user at all so I haven't formed an opinion on this. I'll defer to your judgement.

a13:
> That would be amazing, thanks!

I'm not good at writing documentation (neither English is my native language), but I think it's better than nothing :)

> I'm not a sophisticated git user at all so I haven't formed an opinion on this. I'll defer to your judgement.

OK, then I'm leaving it as is since there were a couple of merges

oantolin:
Perfect, thank you!

a13:
> Perfect, thank you!

Thank you!

PR #19, opened 2020-05-24
Allow deactivating embark indication

clemera:
As I use which key as a key menu prompt I don't need the indicator and would like to deactivate it so the prompt does not get pushed to the right.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -130,9 +130,12 @@ string or nil (to indicate it found no target)."
   :group 'embark)
 
 (defcustom embark-indicator (propertize "Act" 'face 'highlight)
-  "Echo area indicator the user is embarking upon an action.
-When set to the empty string no indication is shown."
-  :type 'string
+  "Indicator to use when embarking upon an action.
+If set to a string prepend it to the minibuffer prompt or to the
+message in the echo area when outside of the minibuffer. When set
+to a function it is called with no arguments to indicate the
+pending action itself. For nil no indication is shown."
+  :type '(choice function string nil)
   :group 'embark)
 
 (defcustom embark-setup-hook nil
@@ -373,17 +376,19 @@ This is used to keep the transient keymap active."
           negative-argument)))
 
 (defun embark--show-indicator ()
-  "Show pending action indicator either in the minibuffer or echo area."
-  (unless (string-empty-p embark-indicator)
-    (let ((mini (active-minibuffer-window)))
-      (if (not mini)
-          (message "%s on '%s'" embark-indicator embark--target)
-        (setq embark--overlay
-              (make-overlay (point-min)
-                            (minibuffer-prompt-end)
-                            (window-buffer mini)))
-        (overlay-put embark--overlay 'before-string
-                     (concat embark-indicator " "))))))
+  "Show pending action indicator accoring to `embark-indicator'."
+  (cond ((stringp embark-indicator)
+         (let ((mini (active-minibuffer-window)))
+           (if (not mini)
+               (message "%s on '%s'" embark-indicator embark--target)
+             (setq embark--overlay
+                   (make-overlay (point-min)
+                                 (minibuffer-prompt-end)
+                                 (window-buffer mini)))
+             (overlay-put embark--overlay 'before-string
+                          (concat embark-indicator " ")))))
+        ((functionp embark-indicator)
+         (funcall embark-indicator))))
 
 (defun embark--bind-actions (exitp)
   "Set transient keymap with bindings for type-specific actions.
@@ -419,8 +424,8 @@ argument), exit all minibuffers too."
   (embark--setup)
   (unless exitp
     (setq-local enable-recursive-minibuffers t))
-  (embark--show-indicator)
-  (embark--bind-actions exitp))
+  (embark--bind-actions exitp)
+  (embark--show-indicator))
 
 (defun embark-exit-and-act (&optional continuep)
   "Exit the minibuffer and embark upon an action.


protesilaos:
Maybe the docstring should read:

```diff
-When set to the empty string no indication is shown.
+When set to an empty string, no indication is shown.
```

 Or a direct like: "Set to an empty string to hide the indicator".

clemera:
Thanks for the suggestion. I would like the direct version better but it does deactivate the indication and not hide the indicator. Maybe "Set to an empty string to deactivate indication"?

clemera:
Thinking about it wouldn't it be better to have a function for this and switch the order of `embark--show-indicator` and `embark--bind-actions`. This way there wouldn't be the need for an advice [here](https://github.com/oantolin/embark/wiki/Additional-Configuration#use-which-key-like-a-key-menu-prompt)

protesilaos:
You are welcome!

> Maybe "Set to an empty string to deactivate indication"?

Looks good to me.

How about the docstring of `embark--show-indicator`?  Should it be updated accordingly to reference `embark-indicator`?

> This way there also wouldn't be the need for an advice here which has the purpose to use another indication mechanism.

That is for @oantolin!  I just noticed your pull request and thought I would comment :)

clemera:
> How about the docstring of embark--show-indicator? Should it be updated accordingly to reference embark-indicator?

Yes it probably should when it stays like it currently is. 

>  I just noticed your pull request and thought I would comment :)

I see PRs as just an alternative way to start discussions so you are welcome to comment of course.

oantolin:
Rather than the empty string indicate deactivation, wouldn't it be better to use `nil` for that?

clemera:
> Rather than the empty string indicate deactivation, wouldn't it be better to use nil for that?

I agree that would be better. I was unsure because of the `:type string` of the defcustom type. What do you think about the function suggestion? The reason I proposed this PR is because I want to change the indication to another mechanism for that I have to deactivate this and advice `embark--bind-actions`. See [this](https://github.com/oantolin/embark/pull/19#issuecomment-633200292) comment.

oantolin:
I say we allow, nil, strings and functions (and update the :type accordingly). I think allowing the which-key trick to be used without any advice is a worthy goal.

clemera:
Thank you!

oantolin:
No, thank you for implementing this!

oantolin:
I'll update the wiki to use the new setting, setting it to:

```elisp
(defun embark-which-key ()
  (let ((which-key-show-transient-maps t)
        (which-key-replacement-alist
         (cons '(("^[0-9-]\\|kp-[0-9]\\|kp-subtract\\|C-u$" . nil) . ignore)
               which-key-replacement-alist)))
    (which-key--update)))
```

oantolin:
If you want to change anything after I edit it, feel free to do so. In particular, you might object to removing the prefix argument commands from the display.

oantolin:
Wait, no, I should just put this in the README, instead.

clemera:
Perfect :) I'm fine with now showing the prefixes. BTW do you know if there is a way to stop which key from hiding when a prefix key is pressed? I can advice `universal-argument` and such to invoke it again immediately but I wonder if there is a better way.

clemera:
BTW in case you don't know, you can select from the dropdown to merge as a single commit so you don't get the noise in between in you project history.

oantolin:
> BTW in case you don't know, you can select from the dropdown to merge as a single commit so you don't get the noise in between in you project history.

I indeed didn't know, thanks!

oantolin:
> BTW do you know if there is a way to stop which key from hiding when a prefix key is pressed?

Huh. Here setting `embark-indicator` to that little `embark-which-key` function, keeps the display up even after the action has concluded (with `embark-act`, it does disappear with `embark-exit-and-act`).

Something else it making it disappear in your setup, probably selectrum. And for vanilla completion, we should figure out how to remove the indicator afterwards.

clemera:
Do you have `which-key-mode` enabled or do you always trigger it manually?

oantolin:
That was it, with `which-key-mode` the which key pop disappears even after just `C-u`, as you said.

Mmh. But the behavior with `which-key-mode` off suggests the function case of `embark-indicator` should actually allow **two** functions, one to show the indicator and one to remove it.

oantolin:
I attempted to allow a show/hide pair of functions, but the which-key popup reappears (with which-key-mode off). I'm leaving it in another branch (b8974ea) until I figure out what's wrong (help would be much appreciated!).

clemera:
Maybe we could use another approach. Users which would like to use which-key like a key prompt could set an user option `embark-wk-idle-delay` to 0. Internally we could set this buffer locally to `which-key-idle-delay` before the action map is made transient and reset afterwards.

clemera:
Here is a POC of the above idea, seems to work correctly:

```elisp
(defun embark--bind-actions (exitp)
  "Set transient keymap with bindings for type-specific actions.
If EXITP is non-nil, exit all minibuffers too."
  (setq-local which-key-idle-delay 0)
  (setq-local which-key-show-transient-maps t)
  (which-key--start-timer)
  (set-transient-map
   embark--keymap
   #'embark--prefix-argument-p
   (lambda ()
     (setq embark--keymap nil)
     (kill-local-variable 'which-key-idle-delay)
     (kill-local-variable 'which-key-show-transient-maps)
     (which-key--start-timer)
     (run-hooks 'embark-pre-action-hook)
     (when (and exitp
                (not (memq this-command
                           '(embark-cancel embark-undefined))))
       ;; schedule later rerun of this-command
       (run-at-time 0 nil
                    (lambda (cmd arg)
                      (setq inhibit-message nil)
                      (let ((this-command cmd)
                            (prefix-arg arg))
                        (command-execute cmd)))
                    this-command prefix-arg)
       ;; avoid the back to top level message
       (setq inhibit-message t)
       ;; and cancel current run of this-command
       (top-level)))))
```

clemera:
After experiments I figured out `embark-pre-action-hook` can be used for this:

```elisp
(setq embark-indicator
      (defun embark-which-key-setup ()
        (let ((which-key-show-transient-maps t)
              (which-key-replacement-alist
               (cons '(("^[0-9-]\\|kp-[0-9]\\|kp-subtract\\|C-u$" . nil) . ignore)
                     which-key-replacement-alist)))
          (which-key--update))))

(add-hook 'embark-pre-action-hook
          (defun embark-which-tear-down ()
            (unless which-key-persistent-popup
              (which-key--hide-popup))))
```

clemera:
After experimenting I figured out `embark-pre-action-hook` can be used for this:

```elisp
(setq embark-indicator
      (defun embark-which-key-setup ()
        (let ((which-key-show-transient-maps t)
              (which-key-replacement-alist
               (cons '(("^[0-9-]\\|kp-[0-9]\\|kp-subtract\\|C-u$" . nil) . ignore)
                     which-key-replacement-alist)))
          (which-key--update))))

(add-hook 'embark-pre-action-hook
          (defun embark-which-key-tear-down ()
            (unless which-key-persistent-popup
              (which-key--hide-popup))))
```

oantolin:
Excellent news! And I see you updated the wiki too, thanks!

PR #20, opened 2020-05-24
Set enable-recursive-minibuffers locally

clemera:
When using `embark-exit-and-act` the initial value of `enable-recursive-minibuffers` is not stored. Because `embark--old-erm` is `nil`,  `enable-recursive-minibuffers` will be set to `nil` on clean up even if it was `t` before.

Setting `embark--old-erm` could be moved outside the `unless exitp` clause to fix this but I have proposed another approach which is to set `enable-recursive-minibuffers` locally so the initial value doesn't have to be stored and the setting only affects the current minibuffer session.


oantolin:
Much better, thank you!

PR #21, opened 2020-05-24
Fix replace behaviour of embark-shell-command-on-buffer

clemera:
`shell-command-on-region` needs to get passed the output buffer to replace the contents, otherwise it only inserts the output. I also mention `shell-command-dont-erase-buffer` in the docstring which can be used to adjust the behaviour.


oantolin:
I did indeed miss the buffer argument, thanks for spotting that! But isn't the buffer the 4th argument and REPLACE the 5th? If I'm reading the docstring correctly, it should be:

```elisp
(shell-command-on-region (point-min) (point-max) command
                         (current-buffer) replace)
```

I don't think your fix should even work (in the replace case), because you aren't passing the 5th argument.

PR #22, opened 2020-05-27
Get button label with possible properties

clemera:
This preserves text properties that are possibly attached to the candidates (I have tested this with a custom grep command which doesn't work without this because it gets the target info from the properties of the search results).


clemera:
Oh I just noticed this breaks other commands and that default `completing-read` strips the properties anyway, I will close this for now.

PR #23, opened 2020-05-30
embark-occur adjustments

clemera:
This PR contains four changes for `embark-occur`:

1. The candidates where no more showing up for me after recent updates because `embark-occur--revert`  would reset `embark-occur-candidates` to `nil` (the minibuffer is still live but recomputing the candidates does not work after exit). Setting `embark-occur-from` to `nil` before exiting/display fixes this.

2. When there are no candidates show a message instead of displaying the empty occur buffer.

3. Run the mode hooks with the occur window selected.

4. Select the occur window after exit and control its display via a new option `embark-occur-display-action`.

I'm not sure adding an `embark-occur-display-action` is worth it myself. A mentioning of how to configure `display-buffer-alist` to detect the non live occur buffer in the docstring of `embark-occur` would also do, I guess.

Let me know if you have any suggestions or objections and I will adjust things accordingly.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -239,6 +239,18 @@ view for types not mentioned separately."
                  (const :tag "Grid view" grid))
   :group 'embark)
 
+(defcustom embark-occur-display-action nil
+  "The action used to display the occur buffer for `embark-occur'.
+
+The value of this option has the form (FUNCTION . ALIST), where
+FUNCTION is a function or a list of functions.  Each such
+function should accept two arguments: a buffer to display and an
+alist of the same form as ALIST.  See `display-buffer' for
+details."
+  :type '(cons (choice function (repeat :tag "Functions" function))
+               alist)
+  :group 'embark)
+
 (defcustom embark-exporters-alist
   '((buffer . embark-ibuffer)
     (file . embark-dired)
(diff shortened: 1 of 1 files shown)


oantolin:
This looks pretty good, let me just ask a few questions to make sure I understand:

1. This is only a problem if calling `embark-occur` from a minibuffer, right? I think in case `embark-occur` is called from a `dired` or `ibuffer` buffer we shouldn't set `embark-occur-from` to `nil.

2. OK. :)

3. Good idea, but do we really need both `with-selected-window occur-window` and `with-current-buffer occur-buffer`? I'm pretty sure it is enough to select the window, and that also selects the buffer.

4. Selecting the window after exit is a good idea, I'm kind of embarrased I wasn't already doing that. About `embark-occur-display-action`: isn't using it the same as adding an entry for `*Embark Occur*` in `display-buffer-alist`? If it is, then I'd rather take the docstring approach. And we should also mention the corresponding thing for `*Embark Live Occur*` in the docstring for `embark-live-occur`.

oantolin:
It bothers me to have the bug you pointed out in 1, so I'm fixing that right away (only for calls to `embark-occur` from a minibuffer), but don't take that as a sign that I'm unwilling to discuss 1 further.

oantolin:
Sorry, upon further consideration here's what I'll do: merge now, change 1 to only apply to calls from minibuffers, and remove `embark-occur-display-action`. But again, I'm open to further discussion on any of these points.

oantolin:
OK, I:

- Removed the redundant with-current-buffer (for 3), eb24b20
- Removed the display action option (bfe3556) and added to the docstrings of `embark-occur` and `embark-live-occur`, but as I said am willing to discuss it.
- Made the `(setq embark-occur-from nil)` fix only apply to calls from minibuffer (82133d1). I kind of like being able to revert an Embark Occur buffer that I made from dired, for example.

clemera:
Nice! Thanks for fixing all the problems with my PR right away!

oantolin:
I'm glad you approve of the changes. Thanks for this PR, the bug in 1 was embarrasing, and selecting the window for the hooks and after exit are great improvements.

PR #24, opened 2020-05-31
Propose changes for target based classification

clemera:
I have many completion commands which don't report the completion category and there might be other contexts where it is more useful to classify based on target. I noticed that we could distinguish between classification of buffer context and classification based on target. This also removes the doubling of ffap and symbol at point in classifiers and target finders. If the buffer context doesn't report a type we can fallback to determine the type using the current target.

I also made the functions in `embark-classifiers` position independent by moving `embark-cached-type` and `embark-target-type` into `embark-classify`. The reasoning is that the cached type is meant for internal use and should always be checked first and the target based classification should always be last. This makes user configuration easier because one has not to keep the order of those in mind.

Here are a two addition thoughts about classification that I think are worth noting: 

- There could be custom completion sessions which list targets with multiple types (for example my custom switch buffer command contains buffers and recent files). This makes occur problematic for those completion sessions. One possible solution I see would be to pick only those candidates which report the same target type if the class for occur  was determined by target classification. 

- In theory a target could have multiple types (the name could be buffer/file/symbol at the same time but those situations should be rare and we can probably ignore that for now). In the future we could maybe allow classifiers to report multiple types and then decide how to handle this generally if we think it's worth the effort.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -115,10 +115,23 @@
     embark-symbol-completion-type
     embark-dired-type
     embark-ibuffer-type)
-  "List of functions to classify according to current buffer context.
+  "List of functions to classify current buffer context.
 Each function should take no arguments and return the type
 symbol, or nil to indicate it could not determine the type in
-current context."
+current context. If the type is not determined by current buffer
+context fallback to `embark-target-classifiers'."
+  :type 'hook
+  :group 'embark)
+
+(defcustom embark-target-classifiers
+  '(embark-file-target-type
+    embark-symbol-target-type
+    embark-buffer-target-type)
+  "List of functions to classify current target.
+Each function takes the target as argument and returns the type
+symbol, or nil to indicate it could not determine the type of
+current target. If the type isn't determined by current target
+fallback to the `general' type."
   :type 'hook
   :group 'embark)
 
@@ -377,12 +390,22 @@ Always keep the non-local value equal to nil.")
   "Report type determined by target."
   (when-let ((target
               (run-hook-with-args-until-success 'embark-target-finders)))
-    (cond ((get-buffer target)
-           'buffer)
-          ((intern-soft target)
-           'symbol)
-          ((file-exists-p target)
-           'file))))
+    (run-hook-with-args-until-success 'embark-target-classifiers target)))
+
+(defun embark-file-target-type (cand)
+  "Report file type if CAND is a file."
+  (when (file-exists-p cand)
+    'file))
+
+(defun embark-symbol-target-type (cand)
+  "Report symbol type if CAND is a known symbol."
+  (when (intern-soft cand)
+    'symbol))
+
+(defun embark-buffer-target-type (cand)
+  "Remport buffer type if CAND is a buffer name."
+  (when (get-buffer cand)
+    'buffer))
 
 (defun embark-classify ()
   "Classify current context."
--- a/embark.el
+++ b/embark.el
@@ -385,7 +385,7 @@ Always keep the non-local value equal to nil.")
            'file))))
 
 (defun embark-classify ()
-  "Classify current minibuffer completion session."
+  "Classify current context."
   (or (embark-cached-type)
       (run-hook-with-args-until-success 'embark-classifiers)
       (embark-target-type)
--- a/embark.el
+++ b/embark.el
@@ -116,14 +116,15 @@
     embark-symbol-completion-type
     embark-dired-type
     embark-ibuffer-type
-    embark-ffap-type
-    embark-symbol-at-point-type)
-  "List of functions to classify the current completion session.
-Each function should take no arguments and return a symbol
-classifying the current minibuffer completion session, or nil to
-indicate it could not determine the type of completion."
+    embark-target-type)
+  "List of functions to classify according to current context.
+Each function should take no arguments and return the type
+symbol, or nil to indicate it could not determine the type in
+current context."
   :type 'hook
   :group 'embark)
+
+(autoload 'ffap-file-at-point "ffap")
 
 (defcustom embark-target-finders
   '(embark-top-minibuffer-completion
@@ -131,7 +132,7 @@ indicate it could not determine the type of completion."
     embark-completion-at-point
     ffap-file-at-point
     embark-symbol-at-point)
-  "List of functions to pick the target for actions.
+  "List of functions to determine the target in current context.
 Each function should take no arguments and return either a target
 string or nil (to indicate it found no target)."
   :type 'hook
@@ -374,17 +375,16 @@ Always keep the non-local value equal to nil.")
   "Report that ibuffer buffers yield buffer."
   (when (derived-mode-p 'ibuffer-mode) 'buffer))
 
-(autoload 'ffap-file-at-point "ffap")
-
-(defun embark-ffap-type ()
-  "If there is a file at point, report it."
-  (when (ffap-file-at-point) 'file))
-
-(autoload 'symbol-at-point "thingatpt")
-
-(defun embark-symbol-at-point-type ()
-  "If there is a file at point, report it."
-  (when (symbol-at-point) 'symbol))
+(defun embark-target-type ()
+  "Report type determined by target."
+  (when-let ((target
+              (run-hook-with-args-until-success 'embark-target-finders)))
+    (cond ((get-buffer target)
+           'buffer)
+          ((intern-soft target)
+           'symbol)
+          ((file-exists-p target)
+           'file))))
 
 (defun embark-classify ()
   "Classify current minibuffer completion session."
--- a/embark.el
+++ b/embark.el
@@ -110,14 +110,12 @@
   :group 'embark)
 
 (defcustom embark-classifiers
-  '(embark-cached-type
-    embark-category-type
+  '(embark-category-type
     embark-package-type
     embark-symbol-completion-type
     embark-dired-type
-    embark-ibuffer-type
-    embark-target-type)
-  "List of functions to classify according to current context.
+    embark-ibuffer-type)
+  "List of functions to classify according to current buffer context.
 Each function should take no arguments and return the type
 symbol, or nil to indicate it could not determine the type in
 current context."
@@ -388,7 +386,9 @@ Always keep the non-local value equal to nil.")
 
 (defun embark-classify ()
   "Classify current minibuffer completion session."
-  (or (run-hook-with-args-until-success 'embark-classifiers)
+  (or (embark-cached-type)
+      (run-hook-with-args-until-success 'embark-classifiers)
+      (embark-target-type)
       'general))
 
 (defun embark-target ()
--- a/embark.el
+++ b/embark.el
@@ -399,8 +399,12 @@ Always keep the non-local value equal to nil.")
 
 (defun embark-symbol-target-type (cand)
   "Report symbol type if CAND is a known symbol."
-  (when (intern-soft cand)
-    'symbol))
+  (let ((sym (intern-soft cand)))
+    (when (and sym
+               (or (boundp sym)
+                   (functionp sym)
+                   (facep sym)))
+    'symbol)))
 
 (defun embark-buffer-target-type (cand)
   "Remport buffer type if CAND is a buffer name."
--- a/embark.el
+++ b/embark.el
@@ -145,7 +145,8 @@ fallback to the `general' type."
     embark-symbol-at-point)
   "List of functions to determine the target in current context.
 Each function should take no arguments and return either a target
-string or nil (to indicate it found no target)."
+string or nil (to indicate it found no target). If the region is
+active the region content is used as current target."
   :type 'hook
   :group 'embark)
 
@@ -392,6 +393,15 @@ Always keep the non-local value equal to nil.")
               (run-hook-with-args-until-success 'embark-target-finders)))
     (run-hook-with-args-until-success 'embark-target-classifiers target)))
 
+(defun embark-active-region-type ()
+  "Report type of active region target."
+  (when-let ((target
+              (and (region-active-p)
+                   (buffer-substring (region-beginning)
+                                     (region-end)))))
+    (or (run-hook-with-args-until-success 'embark-target-classifiers target)
+        'general)))
+
 (defun embark-file-target-type (cand)
   "Report file type if CAND is a file."
   (when (file-exists-p cand)
@@ -413,7 +423,8 @@ Always keep the non-local value equal to nil.")
 
 (defun embark-classify ()
   "Classify current context."
-  (or (embark-cached-type)
+  (or (embark-active-region-type)
+      (embark-cached-type)
       (run-hook-with-args-until-success 'embark-classifiers)
       (embark-target-type)
       'general))


clemera:
I just added an additional idea:  If the region is active it should take precedence and the region contents should be used for classification. This way one can always intentionally select a target regardless of other context.

oantolin:
Wow, thanks for this! You've really been thinking about how to improve embark!

> This also removes the doubling of ffap and symbol at point in classifiers and target finders.

That bothered me when I added them, but I was too lazy to do anything about it. :)

oantolin:
For the problems you mentioned it might be a little tricky to decide what should happen. Note that there are two distinct uses for classification now:

1. `embark-act` uses the classification to decide which keymap to use for actions. This could use target-based classification exclusively.

2. `embark-export` uses the classification to decide what kind of buffer to export to. This is not about an individual target, so it makes sense for it to use the buffer-context-based classification.

> There could be custom completion sessions which list targets with multiple types (for example my custom switch buffer command contains buffers and recent files).

This would be solved for purposes of `embark-act` by not having this session report any context type, and letting the target-specific classification run the show. That way each target would get the appropriate keymap.

But it's not clear to me what `embark-export` should do for a mixed completion like this. Neither dired nor ibuffer is the right thing in this example. Maybe export should actually produce both an ibuffer buffer with the buffers and a dired buffer with the recent files? Or maybe `embark-export` should fall back on `embark-occur` but with no cached type for the occur buffer, so that the target classfiers run each time?

> In theory a target could have multiple types (the name could be buffer/file/symbol at the same time but those situations should be rare and we can probably ignore that for now).

Again, I think this complicates exporting more than acting.

For `embark-act` it makes sense to just join all the keymaps for the relevant type, maybe automatically putting each under a prefix key. What `xref-find-definitions` does on a symbol that names both a variable and a function is popup a buffer where you pick which meaning to use, and I never liked that.

For export, if you have a list of targets with multiple types each, what do you do? Again maybe only the embark-occur fallback makes sense.

oantolin:
> I just added an additional idea: If the region is active it should always take precedence and the region contents can be used for classification. This way one can always intentionally select a target regardless of other context.

I just realized you added a classifier for the active region but not a target finder! So the type of the active region takes precedence but then we don't act on the region. :) Fixed it: 05de61f

By the way, I'm not sure is `region-active-p` is better than `use-region-p` here. Maybe it should be `use-region-p`.

oantolin:
I agreed instintictively with giving precedence to the region if active, but now that I think about it more, I'm not sure I have useful examples in mind. All I've been able to come up with is selecting a portion of a symbol or file name which is again a symbol or file name in its own right. Did you have other uses in mind?

clemera:
> But it's not clear to me what embark-export should do for a mixed completion like this. Neither dired nor ibuffer is the right thing in this example. Maybe export should actually produce both an ibuffer buffer with the buffers and a dired buffer with the recent files? Or maybe embark-export should fall back on embark-occur but with no cached type for the occur buffer, so that the target classfiers run each time?


Maybe we could pick all the candidates which report the same target type (as the current one) to create the export/occur buffer in cases the buffer classification did not report something. So if the classification for occur/export happens by target type we could check all the candidates and filter them accordingly. In my mind this would work like a multisource collection in helm: if you are in another source we would show occur for the that source only. 

> For embark-act it makes sense to just join all the keymaps for the relevant type, maybe automatically putting each under a prefix key. 

> For export, if you have a list of targets with multiple types each, what do you do? Again maybe only the embark-occur fallback makes sense.

Joining the keymaps sounds like a good idea the user could then decide which action makes sense for the current target. For export I think you would only ever have a few candidates which report 
multiple types so if there is no buffer classification we could use the type reported by most of the others. But I'm really not sure if it's worth the effort at all, maybe we can wait and see if multi type candidates are relevant in practice for us.

> I just realized you added a classifier for the active region but not a target finder! So the type of the active region takes precedence but then we don't act on the region. :) Fixed it: 05de61f

Oops, thanks :)

> By the way, I'm not sure is region-active-p is better than use-region-p here. Maybe it should be use-region-p.

I was also thinking about that and wasn't sure but then thought it might be better to be consistent regardless if the region has content or not. But the only way this could be useful would probably be to be able to return target type for the empty string. 

> I agreed instintictively with giving precedence to the region if active, but now that I think about it more, I'm not sure I have useful examples in mind. All I've been able to come up with is selecting a portion of a symbol or file name which is again a symbol or file name in its own right. Did you have other uses in mind?

I was thinking about buffer occur or export buffers which might contain targets different from 
the buffer classification type. So maybe you have a file which is named `this-command` in dired, but you are right the usefulness is limited ;) But maybe we could expand the idea another way: I'm starting to think of embark actions as kind of a "right-click mechanism". Thinking that way right clicking on the active region would ideally show region commands (`upcase-region`, `downcase-region`...) and one could use those as actions on the active region. Right now `embark-active-region-type` falls back to `general', maybe we could introduce a `region' type as fallback.

oantolin:
> Maybe we could pick all the candidates which report the same target type (as the current one) to create the export/occur buffer in cases the buffer classification did not report something. So if the classification for occur/export happens by target type we could check all the candidates and filter them accordingly. In my mind this would work like a multisource collection in helm: if you are in a multi source collection we would show occur/export for the current source only.

I think there's an important difference between occur and export in this context: an occur buffer can naturally contain different types of entries and just let embark-act use target-based classificaction, but an export buffer is naturally geared towards just one type. So I like your idea (and the comparison with multisource collections in Helm), but I think we only need apply it to export, occur can keep all candidates.

> Joining the keymaps sounds like a good idea the user could then decide which action makes sense for the current target. For export I don't know what would make the most sense but maybe we can ignore this for now and wait if multiple types per candidate are actually relevant in practice for us.

Agreed for export, but maybe even for embark-act this case is not that important. Maybe we should just use embark a little more and see if we really need multi-type targets.

>> By the way, I'm not sure is region-active-p is better than use-region-p here. Maybe it should be use-region-p.

> I was also thinking about that and wasn't sure but then thought it might be better to be consistent regardless if the region has content or not. But I agree that it is probably useless to have it recognize the empty region.

I think I'll change it to `use-region-p`: it goes well with your change to not pop up an empty occur buffer but rather show an error message instead.

> I was thinking about occur/export buffers which might contain targets with a different classification type. So maybe you have a file which is named this-command in dired, but you are right that the usefulness of this feature is limited ;) Maybe we could remove that "feature" and expand the idea in another direction:

> I'm starting to think of embark actions as kind of a "right-click mechanism". Thinking that way right clicking on the active region would ideally show region commands (upcase-region, downcase-region...) and one could use those as actions on the active region.

Yes, I think that might be better. Let's give that a shot and we can see how it feels.

> Maybe this use case could be captured in an embark-act-dwim command: If the region is active offer region commands, if there is an active minibuffer call embark-exit-and-act and otherwise call embark-act.

I'm not sure we need a new command for that: how about simply having both embark-act and embark-exit-and-act offer region commands when the region is active? In that case embark-exit-and-act would behave similarly to your embark-act-dwim (because embark-exit-and-act already behaves the same as embark-act when there is no active minibuffer).

clemera:
>  but I think we only need apply it to export, occur can keep all candidates.

Yes that makes sense to me, too.

> Agreed for export, but maybe even for embark-act this case is not that important. Maybe we should just use embark a little more and see if we really need multi-type targets.

I agree, I think it probably isn't worth it.

> I think I'll change it to use-region-p: it goes well with your change to not pop up an empty occur buffer but rather show an error message instead.

Okay, I don't see any downsides to that.

> I'm not sure we need a new command for that: how about simply having both embark-act and embark-exit-and-act offer region commands when the region is active? In that case embark-exit-and-act would behave similarly to your embark-act-dwim (because embark-exit-and-act already behaves the same as embark-act when there is no active minibuffer).

If you are inside the minibuffer I think acting on the region and exit is probably not something that you would like to do? A possible dwim command could act like the following: When the region is active or there is no active minibuffer call `embark-act` and otherwise when the minibuffer is active and there is no active region `embark-exit-and-act`. I would use that as my default command and `embark-act`  would be an exceptional case when I would like to call an action in the minibuffer but keep it open. I think that would support the common case better but maybe that is just my way of doing things.

clemera:
Should the region commands be available under an additional prefix in general map when the region is active? So you can choose from regular actions and region commands.

oantolin:
> If you are inside the minibuffer I think acting on the region and exit is probably not something that you would like to do?

I ran a little experiment.

```elisp
(defun msg-reg ()
  (interactive)
  (message "%s" (if (region-active-p) "ACTIVE REGION" "region not active")))

(bind-key "C-`" #'msg-reg)
```

I marked a region in the scratch buffer, started `find-file` and ran `msg-reg` from the minibuffer. It says "no region active". This makes sense to me: the minibuffer doesn't contain an active region, the scratch buffer does. If I close the minibuffer, the region is still active and now `msg-reg` says "ACTIVE REGION".

So Emacs by `(region-active-p)` seems to mean "is the region in the current buffer (whether its a minibuffer or not) active?", which makes sense to me.

I think this means that `embark-exit-and-act` might actually still be what you want from `embark-act-dwim`, no?

> Should the region commands be available under an additional prefix in general map when the region is active?

What about the opposite: region commands directly available and type-specific ones under a prefix when the region is active? I think if you mark a region and run `embark-act` you most likely want the region commands.

Well, all this discussion is assuming we do add the region commands to embark. Is it really a good idea? Let's be explicit about why we would add them. I can see the following pros:

- It's a curated collection of region commands that you can get a reminder of (either with `C-h` or `which-key`).

- We might think of some region action which aren't simply an existing `*-region` Emacs command.

On the other hand: can't people just run the region commands directly? What's the advantage of going through embark? Is the advantange the reminder or am I missing something?

clemera:
> I think this means that embark-exit-and-act might actually still be what you want from embark-act-dwim, no?

I was thinking about the case when you mark something in the minibuffer input. I wouldn't want to run actions on regions outside the current buffer in that case I would switch to that buffer first. 

Another thing is that even if I know `embark-exit-and-act` and `embark-act` behave the the same outside the minibuffer I don't like the code to go through the extra hoops `embark-exit-and-act` does when not really needed, but that is just me maybe ;) For the general user the name might also be a bit confusing because there is no exit when not in the minibuffer. Maybe instead of introducing a new command `embark-exit-and-act` could be renamed and the definition changed to:

```elisp
(embark-act (active-minibuffer-window)))
```

> What about the opposite: region commands directly available and type-specific ones under a prefix when the region is active? I think if you mark a region and run embark-act you most likely want the region commands.

Yes that makes sense, I agree that would be better in most cases.

> On the other hand: can't people just run the region commands directly? What's the advantage of going through embark? Is the advantange the reminder or am I missing something?

The main advantage for me would be to have a single "right click prefix key" so to speak. I like the idea that I would have a single key that gives me the actions in current context. But I agree that is not necessarily embarks job to provide that convenience ;)

oantolin:
> The main advantage for me would be to have a single "right click prefix key" so to speak. I like the idea that I would have a single key that gives me the actions in current context. But I agree that is not necessarily embarks job to provide that convenience ;)

I think that's a compelling enough reason, let's do it!

I doubt I'd use embark for upcase-region and the like, since I already have muscle-memory for those, but I can see myself using it for other more exotic region commands.

> Another thing is that even if I know `embark-exit-and-act` and `embark-act` behave the the same outside the minibuffer I don't like the code to go through the extra hoops `embark-exit-and-act` does when not really needed, but that is just me maybe ;)

You are not alone, it would bother me too. :)

> For the general user the name might also be a bit confusing because there is no exit when not in the minibuffer. Maybe instead of introducing a new command `embark-exit-and-act` could be renamed and the definition changed to:

>     (embark-act (active-minibuffer-window)))

I like this suggestion a lot! But I can't think of a good name. :( Maybe `embark-final-act`? :)

Even if we can't think of a better name and keep the `embark-exit-and-act` name, I like this redefinition of it, and I'd reword the dosctring too, to make it clear the exiting applies to minibuffers but you can call it from anywhere.

If both `embark-act` and the redefined and possibly renamed `embark-exit-and-act` offered region actions when there is an active region (in the current buffer), then the new `embark-exit-and-act` would behave like you want `embark-act-dwim` to behave, right?

Or maybe we could rename `embark-exit-and-act` (with your proposed redefinition) to `embark-act` and rename the current `embark-act` to `embark-act-in-recursive-minibuffer` or `embark-act-and-continue`.

oantolin:
Mmh. For implementation reasons it would be easier to have the region commands under a prefix: the existing region commands wouldn't consume `embark--target`, and the prefix is a good opportunity to do that.

oantolin:
I made an initial version of region commands. Because it was easier I put the region commands behind a prefix. 72e5730

oantolin:
I flipped it: now with an active region the region commands are available without a prefix and the target-based actions depending on the contents of the region are under the prefix `RET`. be9a0e3

oantolin:
This commit d14ff5d makes it so the extra steps taken for `embark-exit-and-act` are only taken if they make a difference, i.e., only if called from a minibuffer. So now, I think maybe now only the issue of the function names remains. (Is that right? Or is there still some behavior to discuss?)

clemera:
Thanks a lot, this seems to work great! 

> Or maybe we could rename embark-exit-and-act (with your proposed redefinition) to embark-act and rename the current embark-act to embark-act-in-recursive-minibuffer or embark-act-and-continue.

I think that would be great, `embark-act-and-continue` sounds very good to me! I think the renaming would improve things for new users.

> So now, I think maybe now only the issue of the function names remains. (Is that right? Or is there still some behavior to discuss?)

I think there is one last thing worth thinking about: If you mark something within the minibuffer and apply a region command on it you would probably want the minibuffer to stay open because you want to continue with the changed text somehow. The region commands aren't actions on minibuffer candidates but on the contents so I think they shouldn't affect the minibuffer state. What do you think?

oantolin:
> I think that would be great, I think the renaming would improve the distinction for new users. `embark-act-and-continue` sounds already very good to me! Maybe `embark-minibuffer-act-and-continue` so the name tells the command is for usage with the minibuffer.

I think you are on to something with adding minibuffer to the name, so that it's clearer that that is the difference between the two commands. How about `embark-act` and `embark-act-keeping-minibuffer`? Another option is to keep `embark-act` as it is currently and call the other one `embark-act-exiting-minibuffer`. It could also be `keep` and `exit` without `ing`, but those don't read as well.

> I think there is one last thing worth thinking about: If you mark something within the minibuffer and apply a region command on the region you would probably want the minibuffer to stay open because you want to continue with the changed text somehow.

Agreed. Even if you called the supposedly exiting version of the command, right?
I'll make it so.

> When `embark-exit-and-act` would be renamed to `embark-act` there also wouldn't be a "name conflict" with this behaviour any more. What do you think?

I guess this speaks against my idea of `embark-act-exiting-minibuffer`. :)

So `embark-minibuffer-act-and-continue` or `embark-act-keeping-minibuffer` are OK, but I kind of don't want such long names... Here's a table of name options:

| Exiting             | Non-exiting                   |
|---------------------|-------------------------------|
| embark-exit-and-act | embark-act                    |
| embark-act          | embark-act-keeping-minibuffer |
| embark-act          | embark-act-keeping-minibuffer |
| embark-act          | embark-act*                   |
| embark-act!         | embark-act                    |
| embark-nact         | embark-act                    |
| embark-act          | embark-act-on                 |
| embark-act          | embark->act                   |

The idea for `embark-act!` and `embark-nact` is that they are the "destructive" versions of `embark-act` (because they destroy the minibuffer); the `!` is the Scheme convention for destructive functions, and `n` (like `nreverse` or `nconc`) is the Lisp convention. Maybe it could even be `embark-enact` as a pun on `embark-nact`. :)

I included the starred (`*`) name, but I think which of the two commands gets the star is a little arbitrary so it would make it hard to remember which is which.

The idea for `embark->act` is that the arrow indicates you keep going? I don't know, maybe that's weak.

clemera:
I like some of your suggestions in the table but I think I would prefer more descriptive names for the uninitiated ;) But I also think you are right that the names shouldn't be to long. What about `embark-act` for the exiting version and `embark-minibuffer-act` (which indicates it's a special command which can be used in the minibuffer and keeps the minibuffer) for the other one?

oantolin:
`embark-minibuffer-act` is OK, but it makes it sound like it is *for* the minibuffer, when you can use it anywhere. One more suggestion: `embark-acting`, because the "-ing" verb ending is used for continuing actions. :)

By the way, I just noticed a bug when trying to use region commands for an active region in the minibuffer: the indicator doesn't appear! Does that happen for you too?

(Side remark: that's with `embark-act`, `embark-exit-and-act` doesn't work at all because the minibuffer along with its active region is gone by the time the action takes place --but as you suggested above even the exiting command shouldn't exit in the region-in-the-minibuffer case and I haven't fixed that yet.)

clemera:
> embark-minibuffer-act is OK, but it makes it sound like it is for the minibuffer, when you can use it anywhere.

Is there a reason to use it outside the minibuffer when the exiting command acts the same outside the minibuffer? If not what about the combination `embark-minibuffer-acting`?

> By the way, I just noticed a bug when trying to use region commands for an active region in the minibuffer: the indicator doesn't appear! Does that happen for you too?

Yes it happens here, too. I did not notice it before because I use the following:

```elisp
(setq embark-indicator
      (defun embark-which-key-setup ()
        (let ((which-key-show-transient-maps t)
              (which-key-replacement-alist
               (cons '(("^[0-9-]\\|kp-[0-9]\\|kp-subtract\\|C-u$" . nil) . ignore)
                     which-key-replacement-alist)))
          (setq-local which-key-show-prefix nil)
          (setq-local which-key-persistent-popup t)
          (which-key--update))))

(add-hook 'embark-pre-action-hook
          (defun embark-which-key-tear-down ()
            (kill-local-variable 'which-key-persistent-popup)
            (kill-local-variable 'which-key-show-prefix)
            (unless which-key-persistent-popup
              (which-key--hide-popup))))
```

oantolin:
> Is there a reason to use it without the minibuffer when the exiting command acts the same in this case?

Well, no, but there's also no reason to use the non-exiting command outside of the minibuffer either, since you can use the other one. :)

By the way, I couldn't remember what Ivy calls the exiting and non-exiting action commands. Looking it up didn't really inspire me: the exiting command is `ivy-dispatching-done` and the non-exiting command `ivy-dispatching-call`. I think `done` is evocative but `call` isn't, really.

How about `embark-act-noexit` (there are functions in Emacs that create a buffer but don't select its window called `*-noselect`) or `embark-act-rec` (because the action occurs in a recursive minibuffer) or `embark-act-open` (because it keeps the minibuffer open)?

oantolin:
You know, I think for the exiting command is the main one, but for @protesilaos and me it's the non-exiting command. I think Protesilaos doesn't even bind the exiting command. I bind both in the minibuffer but I bind the non-exiting command to `;` which has no modifiers and the exiting command to `:` which I need to use shift for. (By the way, I haven't had to enter a single semi-colon or colon at a completing minibuffer prompt, so those keys were good choices.)

oantolin:
Oh, the indicator not appearing when a there's an active region in the minibuffer is mysterious: if I use edebug to step through `embark--show-indicator` it does appear!

clemera:
> How about embark-act-noexit (there are functions in Emacs that create a buffer but don't select its window called *-noselect) or embark-act-rec (because the action occurs in a recursive minibuffer) or embark-act-open (because it keeps the minibuffer open)?

The main reason I thought it may make sense to rename them was that I thought `embark-exit-and-act` may be confusing because there is no exit outside the minibuffer and you can call it there, too. But with `embark-act-noexit` the confusing one would just be flipped because what should noexit meant outside the minibuffer? I did not thought about the case that people would like to use the non-exiting command as the default but obviously I'm in the minority in out small sample :) 

In this case I think the suggestions in your table are better because their names don't imply a minibuffer at all. Or maybe we could have two commands `embark-act-and-exit` and `embark-act-noexit` which are meant to be called from the minibuffer only and a third one called `embark-act` (maybe just a defalias?) which is meant to be called outside the minibuffer only. This way the distinction would be clearer from a user view point and one would be able to globally bind `embark-act` and bind the other ones in `minibuffer-local-map` as they like without having to refer to documentation.

clemera:
Okay, I agree that is a good solution.

oantolin:
Since we both find `embark-act-noexit` acceptable and I want to lay this to rest, I've gone with that name. b8c9d1c

Thanks for the initial PR and the subsequent discussion of region commands and of the naming issue (of which I was initially skeptical, but became convinced that the name `embark-exit-and-act` did actually lie), all of which improved embark!

clemera:
Great! Thanks for the thoughtful discussion as always!

PR #30, opened 2020-06-16
Some fixes for embark-occur

clemera:
This PR addresses the following points:

- Because `embark-occur` runs after exit the metadata can't be retrieved any more when the buffer gets reverted. I added a buffer local variable `embark-occur-annotation-func` to remember the annotation function for the current occur buffer.
- Avoid adjusting `tabulated-list-revert-hook` globally by adding `embark-occur--revert` locally
- I noticed that embark occur buffers don't show the header line even if `tabulated-list-use-header-line` is non nil so I added a call to `tabulated-list-init-header` on display of the embark occur buffer (this is mentioned in the docstring of `tabulated-list-mode` for modes which derive from it). One can use `embark-occur-mode-hook` to change the setting for `tabulated-list-use-header-line`

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -825,7 +825,8 @@ To be used as an annotation function for symbols in `embark-occur'."
 
 (defun embark-annotation-function-metadatum (cand)
   "Use the `annotation-function' metadatum to annotate CAND."
-  (when-let ((annot-fn (or (completion-metadata-get (embark--metadata)
+  (when-let ((annot-fn (or embark-occur-annotation-func
+                           (completion-metadata-get (embark--metadata)
                                                     'annotation-function)
                            (plist-get completion-extra-properties
                                       :annotation-function)))
@@ -991,6 +992,9 @@ If you are using `embark-completing-read' as your
 (defvar-local embark-occur-linked-buffer nil
   "Buffer local variable indicating which Embark Buffer to update.")
 
+(defvar-local embark-occur-annotation-func nil
+  "Annotation function of minibuffer session for this occur.")
+
 (defvar-local embark--live-occur--timer nil
   "Timer scheduled to update Embark Live Occur buffer.")
 
@@ -1119,6 +1123,11 @@ Argument BUFFER-NAME specifies the name of the created buffer."
                                              ; buffer is displayed, so
                                              ; they can use the window
       (setq embark-occur-from from)
+      (setq embark-occur-annotation-func
+            (or (completion-metadata-get (embark--metadata)
+                                         'annotation-function)
+                (plist-get completion-extra-properties
+                           :annotation-function)))
       (add-hook 'tabulated-list-revert-hook #'embark-occur--revert)
       (setq embark-occur-view
             (or initial-view
--- a/embark.el
+++ b/embark.el
@@ -1128,7 +1128,7 @@ Argument BUFFER-NAME specifies the name of the created buffer."
                                          'annotation-function)
                 (plist-get completion-extra-properties
                            :annotation-function)))
-      (add-hook 'tabulated-list-revert-hook #'embark-occur--revert)
+      (add-hook 'tabulated-list-revert-hook #'embark-occur--revert  nil t)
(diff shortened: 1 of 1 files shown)


oantolin:
Excellent, thanks for this!

- I think I have a bit of a blindspot for the needs of non-live occur, I'll try to keep it in mind more.
- I confess I thought `tabulated-list-revert-hook` was declared with `defvar-local`, almost all the other `tabulated-list` variables are.
- I don't find the header particularly useful, but it is good to be able to have it, of course.

oantolin:
Oh, now I get the header line by default, I don't think I like that, specially since it doesn't update when you switch from grid to list view.

clemera:
Actually I think I don't want the header line by default, too. Maybe we could have extra command to toggle it on demand?

oantolin:
I have a plan and more weird complaints. I'll post something in a few minutes.

oantolin:
Well, maybe before that: if we had a command to toggle the header would you ever use it? Do you personally ever want the header? I don't. I worry we might have zero evidence anyone will want the header. Maybe we can just leave it out and wait to see if anyone requests it.

oantolin:
I think the header only serves two purposes:

1. To tell you what the columns mean. Here that's unnecessary, it's clear what they are.
2. To give you the ability to sort on a column. That doesn't make much sense for grid view. Maybe for list view it does, but currently the code doesn't mark those columns as sortable-by. (I could change that.) But I doubt people really want to sort alphabetically by annotation, maybe by candidate...

If that's all it does maybe we just don't need it for embark occur.

clemera:
I agree probably not worth it. I like to have the header in tabulated list modes as it indicates to me it's a tabulated list but that's all.

oantolin:
So I removed the call to `tabulated-list-init-header` for now, but am open to adding it later if anybody wants it, but there are a couple of things to consider if we do:

- It should be called from `embark-occur--list-view` and `embark-occur--grid-view` instead, so that the header updates when you switch views.

- Here's the weird complaint I hinted at earlier: I'm back on Emacs 28 and it seems that if you call `tabulated-list-init-header` you always get a header indepently of the value of `tabulated-list-use-header-line`! If that variable is `t` you get a nice normal looking header, but if the variable is `nil` you instead get a truly hideous "fake header" styled with this face:

```elisp
(defface tabulated-list-fake-header
  '((t :overline t :underline t :weight bold))
  "Face used on fake header lines."
  :version "27.1")
```

(Since there is a `:version` 27.1 there, I'm assuming this madness started in Emacs 27.1.) This design makes no sense to me, so I must be misunderstanding something. Maybe I'm missing some `tabulated-list-really-dont-use-header-line-not-even-a-fake-one` option. :D

If there really is no way to avoid a header in Emacs 27.1 if you call `tabulated-list-init-header`, and it turns out that some people do want the possiblity of having a header line in emabrk occur buffers, then what I would do is only call `tabulated-list-init-header` after checking that `tabulated-list-use-header-line` is set to `t`.

oantolin:
OK, I guess we don't have to wait for someone to ask for them if they are off by default. I did what I suggested above: check `tabulated-list-use-header-line` before calling `tabulated-list-init-header` and to do it from both `embark-occur--list-view` and `embark-occur--grid-view`, so that the header line updates when you switch view.

And, as you said, those who want the header can do:

```elisp
(add-hook 'embark-occur-mode-hook
          (defun use-header ()
            (setq tabulated-list-use-header-line t)))
```

oantolin:
Oh, and I didn't remember but in list view I did make the Candidate column a column you can sort by.

oantolin:
I added a command to toggle the header.

clemera:
Nice, does not harm to have it there.

oantolin:
> Nice, does not harm to have it there.

I guess not, but I am kind of tempted to rip it out, in case no-one ever uses it. :D

clemera:
You could add some telemetry to detect that ;)

PR #32, opened 2020-07-01
Add option to choose prompt style

clemera:
Hi Omar,

This PR aims to allow choosing embark commands via completion by default (like helm does). This is useful if you always want to use this prompt style and do not want the extra step of invoking `embark-keymap-help`. I also refactored the code a bit and fixed some clean up issues I had with `embark-keymap-help`.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -175,7 +175,8 @@ current input string or nil (to indicate it is not applicable)."
 The `default' style prompts for keys and makes use of
 `embark-action-indicator' and `embark-become-indicator'. There is
 also `completion' style which prompts with completion."
-  :type 'symbol
+  :type '(choice (const default)
+                 (const completion))
   :group 'embark)
 
 (defcustom embark-action-indicator (propertize "Act" 'face 'highlight)
--- a/embark.el
+++ b/embark.el
@@ -356,6 +356,14 @@ Possible values of this variable are:
                  (const :tag "Truncate docstrings" truncate))
   :group 'embark)
 
+(defcustom embark-keymap-help-format-function 'embark-keymap-help-format-default
+  "Function to format the keymap bindings.
+
+The function is called with two string arguments: The key and the
+command name. It should return the string used for completion."
+  :type 'function
+  :group 'embark)
+
 ;;; stashing information for actions in buffer local variables
 
 (defvar-local embark--type nil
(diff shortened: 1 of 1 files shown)


oantolin:
Sorry, Clemens! I hadn't seen this. (I naively thought GitHub notified me via email of PRs.) It sounds like a good idea. I'll give it a try and read the code and get back to you.

oantolin:
Looks good!

If I'm reading this correctly though, with the `completion` prompt style you are limited to the actions (and become targets) in the keymaps, right? You lose the ability to use arbitrary commands as actions (or become targets)?

I'm not saying there is anything wrong with that (people can opt to trade completion for open-endedness), and obviously the keymaps are meant to cover 99% of the command you'd want. (Well, I don't mean the keymaps _as they ship with Embark_, what I mean is that I would expect users to add commands until the keymaps do cover 99% of their usage.) I wonder if there is an easy way to add an escape hatch to the `completion` prompt style.

oantolin:
I'm trying out the `completion` style now and have a few observations:

- The defcustom could use `choice` to present the options in a friendlier way in custom buffers.
- Maybe people using `completion` don't need or want to see the keybindings for the actions.
- The `completion` prompt style doesn't currently respect the `allow/skip-edit` lists. This isn't just a matter of respecting user's preferences: some actions, like `shell-command` in `embark-file-map`, really do need further input from the user. (Currently using `shell-command` just leads to Emacs trying tu run the target as a shell command.)

(Of these I'd only call the last one a bug, but the first one is an easy improvement to make, and the second might be worth thinking about.)

clemera:
> Sorry, Clemens! I hadn't seen this. (I naively thought GitHub notified me via email of PRs.) 

No problem, I just assumed you might be busy with other things. You should be able to configure email notifications in your account settings.

 
>     * The defcustom could use `choice` to present the options in a friendlier way in custom buffers.
> 
>     * Maybe people using `completion` don't need or want to see the keybindings for the actions.
> 
>     * The `completion` prompt style doesn't currently respect the `allow/skip-edit` lists. This isn't just a matter of respecting user's preferences: some actions, like `shell-command` in `embark-file-map`, really do need further input from the user. (Currently using `shell-command` just leads to Emacs trying tu run the target as a shell command.)
> 
> 
> (Of these I'd only call the last one a bug, but the first one is an easy improvement to make, and the second might be worth thinking about.)

Thanks for the review, I will look into these points. Regarding the second one I think it is useful to see them so one can learn them potentially but maybe that should be optional?

oantolin:
> Regarding the second one I think it is useful to see them so one could potentially learn them but having an options for this is a good idea.

Mmh. But a `completion` prompt style user won't be able to _use_ the keybindings until they change the prompt style, right?

clemera:
If you use `completion` style by default you can still have a command which invokes `emark-act` with `default` style when you know the binding already. Another way this can be useful is that you can match against the binding using `^...` (which can be automatic) so you would just press the char and RET.

clemera:
I have worked on the first and last point. I just noticed passing the prefix arg is also a problem with completion style.

oantolin:
> Another way this can be useful is that you can match against the binding using `^...` (which can be automatic) so you would just press the char and RET.

It's funny I didn't think of that, since I've actually done it (with `embark-keymap-help`, before this PR). 😛

clemera:
`embark-act` has its own meaning for the prefix arg which conflicts with using it to pass it along to embark actions. I'm not sure how to proceed with this. Maybe `embark-act` could be changed not to use the prefix arg to toggle its exit behaviour? I haven't use it this way and prefer having separate commands instead but maybe you would prefer to keep it? There could also be a helm style behaviour (pressing the prefix before choosing the completion) but this would be non standard.

oantolin:
Feel free to change what the prefix argument for `embark-act` and `embark-act-noexit` means! I never really liked it (and I don't use it, you don't use it, ...). Passing it along to actions sounds much better.

clemera:
Things should work as expected now. I also added an option to configure the formatting of the bindings using `embark-keymap-help-format-function`.

oantolin:
Sorry, I've been pretty busy with work. Testing now.

oantolin:
Every looks great! Thanks for making the changes I requested and I am sorry it took me so long to get to this.

clemera:
No problem, thanks for your review and merge!

PR #35, opened 2020-07-28
Fix error when getting docs of alias for a not yet defined function

clemera:
I got an invalid-function error when invoking `embark-occur` and fixed it according to the comments in `describe-function-1`: 

```elisp
...
(condition-case nil
    ;; FIXME: Maybe `documentation' should return nil
    ;; for invalid functions i.s.o. signaling an error.
    (documentation function t)
  ;; E.g. an alias for a not yet defined function.
  ((invalid-function void-function) nil))
...
```

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -828,7 +828,11 @@ To be used as an annotation function for symbols in `embark-occur'."
   (when-let* ((symbol (intern name))
               (docstring
                (if (fboundp symbol)
-                   (documentation symbol)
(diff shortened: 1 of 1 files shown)


oantolin:
Thanks for this! I'm sorry I've been neglecting this repo, but I hope to get back on track soon.

clemera:
Thanks, no problem!

Issue #26, opened 2020-06-08
How to create a keymap for a custom function


thy6:
Please help me understand how to setup embark for custom functions. For example, I get "unknown embark action" when I try something like this:

    (let ((embark-custom-map))
      (embark-keymap
       '(("t" . my-embark))
       embark-custom-map)
      (completing-read "prompt: " '("red" "green" "blue")))
    
    (defun my-embark ()
      (interactive)
      (message "You chose %s" (embark-target)))

So that's definitely not the correct way, but then what is the right way for creating embark keymaps for custom functions?

oantolin:
Anothe thing I haven't documented very well, sorry. In my defense, `embark` is still in pretty active development.

You almost have it right, you need to tell embark what "type" of completion you are doing and to register your custom keymap as the keymap to use for that type. I'll do it, as you did, in a let binding, so that configuration isn't permament:

```elisp
(let ((embark--type 'my-custom-type)
      (embark-custom-keymap (embark-keymap
                             '(("t" . my-embark))
                             embark-general-map))
      (embark-keymap-alist
       (cons '(my-custom-type . embark-custom-keymap)
             embark-keymap-alist)))
  (completing-read "prompt: " '("red" "green" "blue")))

(defun my-embark ()
  (interactive)
  (message "You chose %s" (embark-target)))
```

The completion type is arbitrarily chosen to be the symbol `my-custom-type`, and the `let` form adds the pair `(my-custom-type . embark-custom-keymap)` to `embark-keymap-alist`, which is what embark consults to figure out which keymap to offer you.

Notice also that in your code you used `embark-keymap` slightly incorrectly. The last argument is the desired parent keymap, and the return value is the constructed keymap. In your code you throw away the keymap (!), and specified that the null variable `embark-custom-map` should be the parent (which isn't very useful). You'll usually want `embark-general-map` as the parent so you get `C-g`, `C-h`, `C-u` and the general actions `i` and `w`. And in my code you can see that the returned keymap is assigned to `embark-custom-map` instead of discarded.

I have to say, although it works, I'm not terribly happy with the code I wrote above. I'd say embark doesn't support this use-case (one-off commands with their own ad hoc action lists) very elegantly. Maybe I should think about adding way to directly say "trust me, I want *this* keymap", without having to make up a phony type and all. Of course this is Lisp and you can easily fix anything without waiting for me! You can hide that ugly code behind this macro:

```elisp

(defmacro with-embark-actions (actions &rest body)
[...]

PR #37, opened 2020-07-30
Allow to short circuit usual classification

clemera:
This adds two variables `embark-overriding-keymap` and `embark-overriding-export-function` which can be used
to dynamically specify the actions or exporter to use. This makes it easy to let bind those variables to enable specific actions or exporters adhoc.

One could also bind `embark-exporters-alist`  and `embark-keymap-alist` and this would work to some extend (it wouldn't work for the keymaps when the region is active).  But I think with the new variables it's more straight forward because it doesn't require the user to lookup how to create the values which would be: `((t . export-function))` and `((general . map))` and it also works correctly for the case when the region is active.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -110,6 +110,10 @@
   :type '(alist :key-type symbol :value-type variable)
   :group 'embark)
 
+(defvar embark-overriding-keymap nil
+  "Can be bound to short circuit `embark-keymap-alist'.
+Embark will set the parent of this map to `embark-general-map'.")
+
 (defcustom embark-classifiers
   '(embark-category-type
     embark-package-type
@@ -313,6 +317,11 @@ value indicates the default function to use for other types.  The
 default is `embark-occur'."
   :type '(alist :key-type symbol :value-type function)
   :group 'embark)
+
+(defvar embark-overriding-export-function nil
+  "Can be bound to short circuit `embark-exporters-alist'.
+The expected format is the same as for functions in
+`embark-exporters-alist'.")
 
 (defcustom embark-live-occur-delay 0.15
   "Wait this long for more input before updating Embark Live Occur buffer."
@@ -623,7 +632,11 @@ relative path."
 (defun embark--setup-action ()
   "Setup for next action."
   (setq embark--keymap
-        (symbol-value (alist-get (embark-classify) embark-keymap-alist))
+        (or (and embark-overriding-keymap
+                 (let ((map (copy-keymap embark-overriding-keymap)))
+                   (set-keymap-parent map embark-general-map)
+                   map))
+            (symbol-value (alist-get (embark-classify) embark-keymap-alist)))
         embark--target
         (if (use-region-p)
             (when embark--target-region-p
@@ -1345,7 +1358,8 @@ The variable `embark-exporters-alist' controls how to make the
 buffer for each type of completion."
   (interactive)
   (let* ((type (embark-classify))
-         (exporter (or (alist-get type embark-exporters-alist)
+         (exporter (or embark-overriding-export-function
+                       (alist-get type embark-exporters-alist)
                        (alist-get t embark-exporters-alist))))
     (if (eq exporter 'embark-occur)
         ;; let embark-occur gather the candidates


clemera:
As you can see in the code I have written it so that setting the parent map is handled automatically. Maybe this could also be done for the regular action keymaps? Other then the `embark-region-map` they all set `embark-general-map` as the parent. This would make it a bit more convenient to create new keymaps because you would do it like usual without having to use `embark-keymap` to construct them.

oantolin:
Hi Clemens! Sorry I've let this sit here for so long. It looks great, short and sweet. I suggested an ugly work around to do what this PR accomplishes much more directly in issue #26. This is much better, thanks.

oantolin:
And I now think you are right about automatically handling `embark-general-map`. I seem to recall that @a13's original idea was to have the inheritance be automatic but I misguidedly wanted to be more general or something.

clemera:
Thanks Omar! Nice to see you back around here!

clemera:
> And I now think you are right about automatically handling embark-general-map. I seem to recall that @a13's original idea was to have the inheritance be automatic but I misguidedly wanted to be more general or something.

Do you still want to do this?

oantolin:
Sure, except I guess for `embark-region-map`, since the general actions don't actually work.

clemera:
Okay, thanks I haven't follow the development of embark lately and I was wondering if this is still on your radar because removing the `embark-keymap` wrapper would be nice, I think.

oantolin:
If we remove `embark-keymap`, we'd what? Use the following idiom?

```elisp
(defvar embark-file-map
    (define-key map ...)
    ...))
```

clemera:
I would like that, yes. But this sounds like you aren't a big fan? ;)

minad:
I am not a fan of the embark macro as it is defined now, since it is not concise enough. And I think it is justified to have some kind of DSL to define keybindings if they are somehow special purpose, e.g., like hydra or https://github.com/minad/polyp/

minad:
But generally I am more interested in the larger structures ;) For example I proposed to split the core logic from the actions and keymaps, e.g., embark-core.el and embark-action.el to make it easier to dive in. But Embark already got more readable to my eyes - I think @oantolin put quite some work in it with the extraction of the classifiers/annotators and more refactorings.

clemera:
But the embark wrapper simply sets the parent map, which can be also be done dynamically. What I would I like about that is that users don't have to learn about some wrapper when they define their own keymaps for embark.

minad:
> But the embark wrapper simply sets the parent map, which can be also be done dynamically. What I would I like about that is that users don't have to learn about some wrapper when they define their own keymaps for embark.

I think both would be fine from my perspective. Either you have this very concise DSL or nothing at all. I would not worry about the intermediate layer. It does not add anything beyond what bind-key etc already does. And you can always macroexpand.

oantolin:
@clemera So you want `embark--setup-action` to add `embark-general-map` as a parent to all keymaps, and `embark-become` to set `embark-meta-map` as the parent for all its keymaps, right?

Then the only parent map we'd need to set by hand is `embark-meta-map` as parent of `embark-general-map`. Is that right?

clemera:
I think so but haven't looked into the code for a while.

oantolin:
Well, independently of who sets the parent, the point is you the user to be able to bring their own keymap, with no parent, and have embark treat it as having the appropriate parent, right?

clemera:
Yes, that is my intention :)

oantolin:
OK. I think I'll do that. And as for how to define the keymaps that come with embark: I don't want to use the `(let ((map (make-sparse-map))) (define-key map ...) ... map)` idiom, since I want the source to remain readable and I find that idiom pretty ugly.

I kind of like that `embark-keymap` is a function returning an anonymous keymap rather than a macro for defining named keymaps. I think of it as providing keymap literals. You said above, @clemera, that all it does is set the parent but I think that is very far from being true: it is much more readable than the let-idiom.

An `embark-define-keymap` macro would be OK too. I like the idea of "keymap literals", but we never use an anonymous one, so the macro would be fine too.

minad:
I think there is one setq literal keymap. I would actually define embark-define-keymap in terms of an embark-make-keymap.

oantolin:
Is that embark-make-keymap the same as the current embark-keymap?

oantolin:
> I think there is one setq literal keymap

Oh, that's right, `embark-occur-mode-map`: `define-derived-mode` declares the variable for you, and you need store the keymap in it. So then I think I'd just keep the `embark-keymap` function as is, but remove all the parents, since the code would enforce the parental relationships we need.

clemera:
> You said above, @clemera, that all it does is set the parent but I think that is very far from being true: it is much more readable than the let-idiom.

Okay, I hope I did not sound blunt. You are right but there already exists specialized tools to deal with this, like bind-key or general.

oantolin:
Sure, but I want to use such a tool in the embark source code and I do not want to require a dependency.

clemera:
Yes, I meant on the user side, the edit was to late...

oantolin:
I guess if `embark-keymap` were a macro and didn't need to worry about setting the parent map it could have a syntax like `(embark-keymap ("f" find-file) ("r" rename-file) ...)`

clemera:
I wanted to add that I did not want to say your `embark-keymap` is a bad idea. My intention was to remove it from user configuration if that is possible, so they can use their own tooling to deal with keymaps and also don't need to be aware of `embark-keymap`.

oantolin:
> I want to add that I did not meant to say `embark-keymap` is a bad idea. My intention was to remove it from user configuration if that is possible, so they can use their own tooling to deal with keymaps and also don't need to be aware of `embark-keymap`.

No, worries, I figured that's what you meant. I've implemented most of the changes and the code does look nicer now:

- Action maps need no longer have `embark-general-map` as their parent, this is arranged for them.
- Become maps need no longer have `embark-meta-map` as their parent, this is arranged for them.
- The maps that come with embark no longer set a parent (since it is not necessary anymore).
- I fixed a funny bug in `embark-occur-direct-action-minor-mode` in which things like numeric prefix argument command were also being set.
- The code is less imperative in many places.

The one thing I didn't do is turn `embark-keymap` into a macro. I still might, but it seems minor compared to the autoparenting.

clemera:
Thanks for the update, sounds great, I'm running it right now without problems!

PR #51, opened 2020-12-17
Remove unnecessary indirection, avoid using embark-target

minad:
These were some quick finds, but I am not yet very familiar with the code base, so take the changes with a grain of salt ;)

Is there a reason for `embark-describe-symbol`? Furthermore I think it makes sense to avoid embark-target and instead implement commands in a general-purpose way by using `(interactive ...)`. There are more of that kind, if it makes sense one could convert them one by one.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -76,7 +76,7 @@
 
 ;; You can read about the Embark GitHub project wiki:
 ;; https://github.com/oantolin/embark/wiki/Default-Actions
- 
+
 ;; Besides acting individually on targets, Embark lets you work
 ;; collectively on a set of target /candidates/. For example, while you are
 ;; in the minibuffer the candidates are simply the possible completions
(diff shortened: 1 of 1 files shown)


oantolin:
Good idea! I can't remember why some of those exist and use `embark-target`...

oantolin:
Some of the commands that use `embark-target` would feel a little silly prompting you, like `embark-eshell-in-directory` (why would it prompt your for a file, then ignore the actual file and only use the directory?). Other are just annoying to write as a proper interactive command, like `embark-browse-package-url`: I could copy the `interactive` form from `describe-package`, but it's enormous, and it felt weird to copy it, so I just went for `embark-target` instead ---which is a shame, because that one would make perfect sense as a general command that prompts for a package.

oantolin:
I guess the ones that use `embark--target-buffer` are justified in using `embark-target`.

minad:
Yes, I agree with what you say. Not every command makes sense in the general form.

PR #52, opened 2020-12-18
Generalize more commands, keymap macros

minad:
@oantolin I am going a bit over the code base, reading and reviewing things. I hope you agree with some of the changes.

* I think having generalized commands is useful and better style since things are decoupled from embark
* I prefer macros in this case since they allow more compile time checking. They look nicer and shorten the boilerplate a bit.

Some remarks:

* I wonder about the consumption of embark-target via `(ignore (embark-target))`. Why is this necessary if you are resetting the target in `embark--cleanup`?
* Why are actions not always executed in the context of the embark--target-buffer? This would allow to write more commands in the general form. Any reason why this is impossible?


oantolin:
I like all of the efforts to replace usage embark-target with general commands!

I'm not sure how I feel about recreating the earlier (slightly more complex, because it had the option to specify a parent) keymap definition system, except this time with a function and a macro instead of just one function. Is there some advantage to this more elaborate setup I'm not seeing? Is it just to save a little typing?

oantolin:
It probably doesn't make sense to generalize embark-insert-unicode-character, though, because that uses embark--target-buffer, which only makes sense in the context of embark actions.

minad:
> I'm not sure how I feel about recreating the earlier (slightly more complex, because it had the option to specify a parent) keymap definition system, except this time with a function and a macro instead of just one function. Is there some advantage to this more elaborate setup I'm not seeing? Is it just to save a little typing?

I see multiple advantages:

* Looks nicer
* Saves typing
* Allows expansion via macroexpand, which is nice to see what is executed in the end
* Slightly more efficient (not really relevant in this case, but using macros reduces startup speed if you have many of them in your init.el)
* Allows static checking! This one, I find important!

Downsides:

* More complicated than a function (but only marginally, macros are widespread in lisp, I think the distinct criterion of a lisp, so no complain if they are used, I guess)
* An additional intermediate layer (but you can macroexpand!)
* More?

minad:
> It probably doesn't make sense to generalize embark-insert-unicode-character, though, because that uses embark--target-buffer, which only makes sense in the context of embark actions.

I agree regarding this one - but I wonder about the question posed above:

> Why are actions not always executed in the context of the embark--target-buffer? This would allow to write more commands in the general form. Any reason why this is impossible?

oantolin:
OK, let's do a macro, but do need two macros (I earlier mistakenly called embark-make-keymap a function, sorry)? That is do we really need both embark-define-keymap and embark-make-keymap? Wouldn't just inlining the make version into the define version be enough?

oantolin:
I'll think about running all actions in embark--target-buffer that sounds like it either doesn't work for a simple reason or is a great idea. I still don't know which it is. 😄

minad:
> OK, let's do a macro, but do need two macros (I earlier mistakenly called embark-make-keymap a function, sorry)? That is do we really need both embark-define-keymap and embark-make-keymap? Wouldn't just inlining the make version into the define version be enough?

Yes, this is an artifact, since you had some keymaps defined as literals. But these are gone now. So we could also kick out the make-keymap and simply rename embark-define-keymap to embark-keymap if you feel like it.

> I'll think about running all actions in embark--target-buffer that sounds like it either doesn't work for a simple reason or is a great idea. I still don't know which it is. :smile: 

I am also not sure :grin: But I hope it would work out since it makes some sense to execute commands in the target buffer, this is the usual behavior if I execute "M-x command", right? Then the command is executed in the original buffer. And I hope we can move more and more to the point where embark actions are just normal commands. That would be ideal.

oantolin:
Could we have just one macro, and call it `embark-define-keymap`, please? You suggested calling the unique macro `embark-keymap`, but since it defines a variable I do want `define` in the name.

As for running actions in the `embark--target-buffer`, I'm very tempted to just do it and then we use embark for a couple of days to see if anything breaks. I think it should be fine and as you say. more actions could be normal commands then.

oantolin:
Eventually we might want to move some of these actions that have turned into normal commands into consult, right? (Though probably not all are a good fit.) And a little embark-consult package could bind those commands to keymaps.

minad:
I quickly tried the with target buffer idea and it breaks at least the occur buffer. But generally I think it is the way to go.

minad:
@oantolin The macro is called embark-define-keymap now. From my side this is ready.

oantolin:
Ok, I'll this a final read through and merge.

oantolin:
I actually trust your code more that I trust my own, but another set of eyes never hurts.

oantolin:
Mmh, `package.el` really should define a `read-package-name` function. I see in your browe-url function you just use symbol. Let's leave it like that for now, but I would ideally prefer package name completion.

oantolin:
Thanks a lot for this, @minad!

minad:
> Mmh, package.el really should define a read-package-name function. I see in your browe-url function you just use symbol. Let's leave it like that for now, but I would ideally prefer package name completion.

I agree. It is somehow half-baked like this and not really general purpose. But since it is mostly used as an action it is not that bad and sufficient for now.

minad:
Thanks for merging!

minad:
@oantolin I tried a bit around regarding the idea of executing actions always in the target buffer and I think it is not that easy if you use embark-act-noexit, since in that case you are simply executing the action. In the case of embark-act the transient map is left via top-level and then the action is executed in run-at-time/embark-after-exit. There is certainly some refactoring needed to make this work...

oantolin:
I just found a horrible downside to your `embark-define-keymap` macro, @minad: the keymaps are no longer listed in `imenu`... 😭

minad:
> I just found a horrible downside to your embark-define-keymap macro, @minad: the keymaps are no longer listed in imenu... sob

Well, there is an easy fix for this, just add it to imenu-generic-expression. I don't consider this a big downside, this happens for other things too, e.g. use-package.

minad:
@oantolin Now that I am thinking about it - it is actually an advantage! You can add embark keymaps to this imenu regexp and add embark keymaps to the consult imenu narrow such that you get a separate category for them :grimacing:

oantolin:
Right, it's not hard to fix, but it wasn't necessary before. It will be nice to have a special imenu category for them., though. 👍

PR #55, opened 2020-12-18
generalize embark-save

minad:
This one is weird, I tried to generalize/simplify embark-save too and suddenly embark behaves incorrectly on Selectrum. After executing the command with embark-act-noexit, the saved candidate is somehow entered and the candidate lists only shows this one candidate instead of the full candidate list which has been there before. But it does not happen on icomplete.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -1422,10 +1422,10 @@ This is whatever command opened the minibuffer in the first place."
   (with-current-buffer embark--target-buffer
     (insert (substring-no-properties (embark-target)))))
 
-(defun embark-save ()
-  "Save embark target in the kill ring."
-  (interactive)
-  (kill-new (substring-no-properties (embark-target))))
+(defun embark-save (str)
+  "Save STR in the kill ring."
+  (interactive "sStr: ")
+  (kill-new str))
 
 (defun embark-eshell-in-directory ()
   "Run eshell in directory of embark target."


minad:
@clemera Could you check this if you see this too?

clemera:
Why do you want remove the usage `embark-target` from that function? I haven't looked at recent changes but as far as I remember it is essential to how embark makes it magic work and it is expected that things break if you remove it.

minad:
@clemera Instead of using embark-target, I try to rewrite functions such that they obtain the argument via "interactive" and they embark injection mechanism. This way the commands become embark independent and generic and the embark logic is better decoupled from the rest. In this case, the benefits are marginal, but I think it generally is a good guideline to write the commands in the generic way if possible. And maybe in this case it helps uncover some bug or some misunderstanding from my side ;)

clemera:
Ah thanks now I get it, when called interactively the minibuffer is invoked and embark should inject it, thanks.

oantolin:
Wow, this is weird: that function looks like it should definitely work everywhere. I wonder what's going on with Selectrum?

minad:
Yes, we should fix this. But for this reason I did not include it into the PR #52.

clemera:
Should be fixed by [this](https://github.com/raxod502/selectrum/pull/293). We let bound our sate checking variable `selectrum-active-p` which made code run in `selectrum-set-selected-candidate` (used by embarks setup for selectrum) which wasn't supposed to run in non Selectrum sessions (like the `read-string` case in this example).

minad:
@clemera Thank you! This fixes the issue indeed. I added the generalization to #52.

clemera:
Great! Thanks for finding this, too!

Issue #54, opened 2020-12-18
Add bookmark and possibly register actions


minad:
@oantolin I can add those if you like.

oantolin:
Thanks for the bookmark actions!

oantolin:
Do the register actions really make sense? I don't think there are built-in commands that prompt with completing-read for a register. They all use `register-read-with-preview`, which only requires you to type one character. That means there is no chance to run an action, there is no moment before pressing `RET` when you have specified a register but haven't finished the command you started,

minad:
Yes, there is the register menu in consult though, but it does not make sense to add actions for them here.

oantolin:
> Yes, there is the register menu in consult though, but it does not make sense to add actions for them here.

Well, hold on a second. Why not add them here? They won't bother people who don't use consult, and might be useful for consult users. I do not oppose adding stuff that benefits consult users as long as it does not involve making consult a dependency. I already have default list view for the `line` and `kill-ring` categories that consult introduces.

oantolin:
But the things with registers is that other that inserting them or jumping to them, I can't really think of relevant actions, and the consult-register command combines both of those into one command already.

minad:
Ok, I guess there are a few actions which make sense in the consult menu, e.g. increment-register.

oantolin:
To be honest, I don't use `consult-register`, because it's a lot more typing than the built-in register commands, and from a quick glance didn't seem to offer any advantages. What am I missing?

minad:
It is nothing more than a reminder of what register values you have stored, but I am also not using it much. There has been some recent discussion in https://github.com/minad/consult/issues/6 if there is a possibility to make it more useful.

oantolin:
But the built-in commands also remind you of what you have stored, that's what `register-read-with-preview` does.

minad:
@oantolin That's right. But in that case the order of selection action is reversed. As I said, consult-register is of questionable usefulness. Maybe I should kick it out of consult again.

minad:
At least I am pretty sure we should not add actions in Embark for registers as of now. If something comes up we can still reconsider?

oantolin:
> But in that case the order of selection action is reversed.

Ah, gotcha! Yes. I wasn't noticing this because I don't actually use either `insert-register` or `jump-to-register`, instead I use this command I wrote, which is basically the same as `consult-register` but using the special `register-read-with-preview` instead of `completing-read`:

```elisp
[...]

PR #57, opened 2020-12-18
add bookmark actions

minad:
See #54

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -123,6 +123,7 @@
     (unicode-name . embark-unicode-name-map)
     (symbol . embark-symbol-map)
     (package . embark-package-map)
+    (bookmark . embark-bookmark-map)
     (region . embark-region-map))
   "Alist of action types and corresponding keymaps.
 For any type not listed here, `embark-act' will use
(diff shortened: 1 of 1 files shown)


oantolin:
We should add jump, jump-other-window (maybe even jump-other-frame) and relocate, possibly set too.

oantolin:
Oh, and bookmark-insert shadows the general insert command. Since they do different things, it probably makes sense to have both. In the package keymap I bound the general insert command to `I`, we can do the same thing here.

oantolin:
I'll merge and add those.

minad:
> Oh, and bookmark-insert shadows the general insert command. Since they do different things, it probably makes sense to have both. In the package keymap I bound the general insert command to I, we can do the same thing here.

I think I fixed this.

oantolin:
Don't worry, I'll push those changes in a minute. Thanks for the PR!

oantolin:
Done, 9bdbfc40e6f35254be19b3b2bc5230e08b6dc1de.

oantolin:
Huh, the other window and other frame commands were popping up a mouse menu! I fixed that too: d7a89eb2b797e551030989de74634372ddae37c0

Issue #15, opened 2020-05-22
Handle unbound key presses


clemera:
When using an unbound key in the transient map the transient map exits and the usually bound command gets executed. This also means that after such a mistyped key `embark--cleanup` doesn't run because `embark--target` is still set. This way you can end up in a state where you think you quitted `embark` but next time you open a minibuffer `embark--inject` does its thing.

I think any wrong pressed key should probably call a command like `embark-undefined` which would be like `embark-cancel` but showing an additional message to inform the user that the key is undefined. One way I can think of implementing this would be to use `[t]` to bind this command. This is a feature of keymaps to use this command as the default for any unbound key.

clemera:
Looking into it seems adding the following as the last line to `embark-keymap` would do:

    (make-composed-keymap map '(keymap (t . embark-undefined)))

oantolin:
The idea is that you can use any command as an embark a Rion, it doesn't need to be bound in an `embark-something-map`, or even bound at all: you can use `M-x some-command` as an action too.

I wouldn't want users to be limited to actions they had the foresight to bind in the `embark-something-maps`.

clemera:
> The idea is that you can use any command as an embark action, it doesn't need to be bound in an embark-something-map, or even bound at all: you can use M-x some-command as an action too.

How would that work? Don't you need to write a special command to use it as an action?

oantolin:
But there should be a visible indicator of whether an action is pending or not and an easy way to cancel it. I guess `embark-cancel` needs a binding not just in the transient map.

clemera:
Okay I understand. But the "bug" is still a problem no?

clemera:
I at least was bitten by that behaviour several times. I wouldn't like that to ever happen and that is more important to me than having the freedom to be able to use any command as an action adhoc which I don't see me doing. But you are free to disagree of course.

oantolin:
What e, actly is the bug, though? Because injecting the target at the next available prompt is embark doing what it is supposed to do.

I guess if the indicator "Act" disappears before the injection, that's a bug. 

And if there is no way to cancel the action before it happens, that's a bug.

By the way, some people will probably *want* to limit themselves to actions in `embark-*-maps`. So it would be nice to add your `embark-undefined` command to embark, and tell people looking to curtail their own freedom voluntarily to bind it to t in `embark-general-map`.

oantolin:
[...]

PR #59, opened 2020-12-19
remove embark-undefined

minad:
* Check if key is present in embark--keymap.
* Print error message if key is not present.
* The behavior is more strict than before, every keypress must be handled properly by embark
* Do you want to exit the transient map or not if an non-action key is pressed?
  In contrast to the current behavior, the transient map does not exit anymore.
  I prefer that in comparison to the current behavior for non-action keys.
  You can always quit explicitly via C-g.
* Minor doc fixes via package-lint

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -680,6 +680,8 @@ BODY."
        (top-level))))
 
 (defun embark--setup-action (continuep)
+  "Setup Embark action when exiting the transient keymap.
+If CONTINUEP is nil exit all minibuffers."
   (run-hooks 'embark-pre-action-hook)
   (setq embark--action this-command)
   ;; Only set prefix if it was given, prefix can still be added after calling
@@ -1462,7 +1464,7 @@ This is whatever command opened the minibuffer in the first place."
     (message "No homepage found for `%s'" pkg)))
 
 (defun embark-insert-relative-path (file)
-  "Insert relative path to embark target.
+  "Insert relative path to FILE.
 The insert path is relative to `default-directory'."
   (interactive "FFile: ")
   (insert (file-relative-name (substitute-in-file-name file))))
--- a/embark.el
+++ b/embark.el
@@ -691,8 +691,7 @@ If CONTINUEP is nil exit all minibuffers."
     ;; embark-act and cleanup is scheduled then
     (advice-add this-command :after #'embark--cleanup))
   (let ((want-current-buffer
-         (memq this-command '(ignore embark-undefined embark-occur
-                                     embark-live-occur embark-export))))
+         (memq this-command '(ignore embark-occur embark-live-occur embark-export))))
     (setf (buffer-local-value 'embark--command embark--target-buffer)
           embark--command)
     (unless want-current-buffer (set-buffer embark--target-buffer))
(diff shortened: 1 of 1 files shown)


oantolin:
I haven't tested this or read the code, only your description here. But if I understand correctly I am very much opposed philosophically to this change. You want to make embark into something like helm or ivy actions, where you can only use a command as an action if you planned far ahead and put it into an action keymap. For me, a key point of embark is that any command at all can be used as an action: the action keymaps are only for convenience, not necessary at all! You can use *all* your keybindings and even `M-x` to choose a command.

oantolin:
I actually do make use of this freedom occasionally, though of course the actions maps probably account for more than 90% of my actions.

minad:
@oantolin Okay, that makes sense. I did not see it in that way since I am more familiar with these fixed keymaps (also in the context of hydra). However what you are currently doing seems to be a bit fragile. For example if I press ESC ESC ESC while being in Embark, the embark injection mechanism is not reset correctly and this is what worries me when executing arbitrary actions. This is certainly a special case since it jumps out. Besides that, I am not very fond of the self-insert-command remapping and would prefer if these keys are rather handled in the keep-alive handler, such that when inspecting the keymaps, they are not polluted with embark-undefined. But this is clearly a matter of taste.

oantolin:
> However what you are currently doing seems to be a bit fragile.

Yes, it is! If there is a way to make it more robust without losing the freedom to use all key bindings and `M-x` to specify a command, I'm all for it.

The `self-insert-command` rebinding is a compromise. You are not the first person to suggest a radical alteration of embark's philosophy, and this is not the first time I resist! 😄 @clemera once tried to force our poor unsuspecting users into a prison of predefined keymaps, just like you. That time he came up with this trick  of rebinding `self-insert-command` to catch the common case in which you type the wrong letter. That way the target doesn't remained unconsumed ready to pounce on you the next time there is a prompt. You can read issue #15 for details of that discussion.

I'm definitely open to changing this aspect of Embark, as long as I don't loose the freedom to use any command as an action. Somehow nobody but me seems to care about that freedom, but I'm quite reluctant to give it up. 🤷

oantolin:
In particular, I hadn't thought of using the keep alive handler for this. That's a good idea. I definitely don't like the non-action stuff in the keymaps.

minad:
Okay, I agree with your philosophy, it is better to not enforce things. The only thing I am unhappy is this self-insert-command hack, I can rewrite that and update this PR.

minad:
See here https://github.com/oantolin/embark/pull/59/commits/365474b9093bf6b69a1032486149044aceb075f2.

Regarding the keyboard-escape-quit problem, it may make sense to also handle this in the keep-alive handler - just set this-command to ignore. But it is not very important I guess. There is always the question what should go to the keymap and what to the handler. But this self-insert-command check is better added to the handler I think.

oantolin:
Merged! Thanks for this, it is a little nicer.

oantolin:
I spoke too soon: this breaks freedom! Using `M-x` to specify an action doesn't work after this. I'm reverting.

oantolin:
I think I figured out what was wrong with your patch and commit a freedom-loving version.

minad:
Sorry! Thank you for fixing! I believe I had a the same or a similar version you are having at some point too and then over-restricted it without proper checking :/ I am glad it works though - this is better than having the this embark-undefined in the keymap.

minad:
I tried this and it works, but I stumbled a few times now over the issue that the embark target is not properly reset when leaving embark. I am still looking for a good example to reproduce, as mentioned before ESC ESC ESC does it. But it also happens from time to time when doing the freedom M-x and then jumping out again at some point with C-g. Then when opening another minibuffer, e.g. C-x b the target is still there and another buffer opens. I wonder if there is a way to make this more robust or maybe it makes sense to introduce an embark-prison-modus for people who are not careful enough to be free... X-)

oantolin:
I'm trying hard to keep you out of prison of your own making, @minad! 😄 I'm trying a different strategy for scheduling cleanup which I think completes the clean up succesfully in more cases (including `ESC ESC ESC`!). Please test: 32b0976cbc27101b06582ab7fdc265d30f87c9d2

minad:
That's a big relief :joy: 

I quickly tried it and it seems to be work better now! I will let you know if I stumble over more problematic edge cases.

PR #91, opened 2021-01-07
Fix package name to conform with libary header conventions

vinszent:
Small fix that I noticed when trying to install Embark using Nix as it was being installed as `Embark` (capitalized). This was causing it not be able to be loaded properly due to case sensitivity. The official library header conventions state that this should be the base name of the file, so I have corrected it.


oantolin:
I made that uppercase by accident, probably a stray `M-c`. I had already noticed and fixed it in a branch I'm working on and will merge soon, so instead of merging this PR I'll just wait until that branch is ready.

Issue #123, opened 2021-01-24
embark-quit during recursive edits


jakanakaevangeli:
embark-quit's behaviour is incorrect when quitting from a recursive minibuffer
or a recursive edit inside a minibuffer.

To demonstrate this, enable the option `enable-recursive-minibuffers` and press
`M-x` twice. Quitting with `C-g` (bound to `abort-recursive-edit` by default)
takes you back to the outer minibuffer, but quitting with `emabrk-quit` takes
you to the main window and leaves the minibuffer window in an inconsistent
state.

Similarly if you get into a recursive edit in the minibuffer, quitting with
`C-g` will simply abort the recursive edit, but quitting with `embark-quit`
inappropriately runs `minibuffer-exit-hook` and takes you to the main window.

In practise this can happen if you want to run `epa-insert-keys` while inside
the minibuffer for some reason. `epa-insert-keys` enters a recursive edit. If
then you switch to the minibuffer and change your mind about inserting the
keys, you can simply press `C-g` (`abort-recursive-edit`), and you're back to
the minibuffer. Quitting with `embark-quit` will however run the exit hooks and
take you the main window.

oantolin:
I should definitely fix the inconsistent state it leaves the minibuffer in, and should make quitting from an recursive minibuffer leave you in the outer minibuffer.

But you also say I should not run the minibuffer-exit-hook? Why not?

jakanakaevangeli:
> But you also say I should not run the minibuffer-exit-hook? Why not?

Seems like abort-recursive-edit already leads up to the exit hooks being
run, see src/minibuf.c line 645. Us running it as well in embark-quit
means that it runs twice.

abort-recursive-edit also does the right thing of not running it if it
doesn't actually quit a minibuffer, which is the case during
M-x recursive-edit.

However, unlike embark-quit, it isn't guaranteed that they are run with
the minibuffer being the current buffer. Perhaps it is the
responsibility of the hook functions to not assume the minibuffer being
the current buffer and switch to it if needed.

jakanakaevangeli:
I have narrowed the problem down to `(current-window-configuration)` saving the
window configuration too precisely. During the second M-x, mini-window shows
` *Minibuf-2*` and the saved window configuration remembers that. Restoration
of this configuration insists on setting mini-window's buffer to ` *Minibuf-2*`
despite it already being inactive due to `abort-recursive-edit`. Here is the
patch that clumsily tries to address this.

```diff
[...]

PR #125, opened 2021-01-25
Fix embark-quit during recursive minibuffers

jakanakaevangeli:
Fixes https://github.com/oantolin/embark/issues/123.
I have improved on the patch proposed in the issue.


oantolin:
Ah, this looks much nicer than the `(format " *Minibuf-%s*" ...)` trick in the other patch. Thank you!

Issue #121, opened 2021-01-23
Using embark recursively during embark--act doesn't work properly


jakanakaevangeli:
To reproduce with emacs -q, run:
```elisp
(load-file "~/src/emacs/embark/embark.el")
(define-key minibuffer-local-map (kbd "M-o") 'embark-act)
```
press `C-x C-f` and then `M-o c` to copy a file. You are now in a minibuffer
window inside `embark--act`. Using embark now (by pressing `M-o o` for example)
will not quit the first minibuffer properly.

Quitting from the inner minnibuffer with `C-g` will also fail to quit the
outer one, but I personally find this very convenient.

Here is my ugly attempt at fixing this. Using a simple unwind-protect turns out
to not be sufficient because the outermost recursive edit would win at
restoring the incorrect window configuration. It would also break the
previously mentioned `C-g` functionality which i like.

```diff
diff --git a/embark.el b/embark.el
index c91e485..35ee212 100644
--- a/embark.el
+++ b/embark.el
@@ -755,6 +755,8 @@ (defun embark--target ()
         (funcall transformer target)
       (cons type target))))
 
+(defvar embark-act--should-unwind nil)
+
 ;;;###autoload
 (defun embark-act (&optional arg)
   "Prompt the user for an action and perform it.
@@ -779,9 +781,18 @@ (defun embark-act (&optional arg)
                                                target)))
     (if (null action)
         (minibuffer-message "Canceled")
-      (embark--act action target)
-      (when (if embark-quit-after-action (not arg) arg)
-        (embark-quit)))))
+      (let ((should-quit
+             (and (minibufferp) (if embark-quit-after-action (not arg) arg))))
+        (setq embark-act--should-unwind nil)
+        (unwind-protect (embark--act action target)
+          (when (and embark-act--should-unwind should-quit)
+            ;; We don't use `embark-quit' here because only the innermost quit
+            ;; sees the correct window configuration to be restored
+            (abort-recursive-edit)))
+        (when should-quit
+          (setq embark-act--should-unwind t)
+          (run-at-time 0 nil #'set 'embark-act--should-unwind nil)
+          (embark-quit))))))
 
 (defun embark--become-keymap ()
   "Return keymap of commands to become for current command."
```

oantolin:
I think this issue is included in the more general issue you reported in #123, right?

jakanakaevangeli:
As I see it, the issues seem to be independent.

This issue is concerned with embark-act and the way it quits from the
[...]

PR #126, opened 2021-01-25
Fix recursive use of embark-act, second solution

jakanakaevangeli:
Fixes https://github.com/oantolin/embark/issues/121.
This solution avoids `run-at-time` by using throw instead. It improves on my
second second patch, posted in the issue. It uses a static variable in favor
of a global one and improves behaviour of heavy recursive minibuffer usage.


minad:
I did not really follow the discussion, but I am sorry to say that, this looks quite ugly. I've never seen static variables being used somewhere else, is this common practice? I don't see a huge advantage to a global variable. But maybe it makes sense to reconsider the old top-level behavior if the new minibuffer recursion behavior complicates everything in such a way?

PR #127, opened 2021-01-25
 Fix recursive use of embark-act, first solution

jakanakaevangeli:
Fixes https://github.com/oantolin/embark/issues/121.
This solution improves on my first patch from the issue by using a static
variable in favor of a global.
It feels a bit uglier to me than https://github.com/oantolin/embark/pull/126
because it uses `run-at-time` and relies on setting rather than binding the
static variable, but I have posted it anyway because the code from the other
pull request became quite convoluted in order to handle recursion into
non-embark-act minibuffers.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -758,11 +758,9 @@ the type, it is called with the initial target, and must return a
         (funcall transformer target)
       (cons type target))))
 
-(defvar embark-act--should-unwind nil)
-
-;;;###autoload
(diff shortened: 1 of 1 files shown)


oantolin:
The byte compiler doesn't like the `(let (...) (defun ...))` pattern. I rewrote some code of mine that used to avoid the warnings.

PR #129, opened 2021-01-25
Fix recursive use of embark-act

### LINKED ISSUE #121: Using embark recursively during embark--act doesn't work properly

jakanakaevangeli:
Fixes  #121.
Closed all other pull request as they are pretty much an embarrassment compared
to this one.
It simply replaces `run-at-time` with an ephemeral `post-command-hook`. Perhaps
it would be worth to think about all the other uses of `run-at-time` with an
ephemeral post-command-hook. I'm personally not really a fan run-at-time
because it breaks keyboard macros. I know that embark is currently not usable
for kmacros because quitting aborts a macro definition but perhaps we should
make it ready for a future emacs releases where quitting might not abort a
macro definition any more.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -785,9 +785,11 @@ minibuffer, and if you use \\[universal-argument] it will do the opposite."
                                                target)))
     (if (null action)
         (minibuffer-message "Canceled")
-      (embark--act action target)
-      (when (if embark-quit-after-action (not arg) arg)
-        (embark-quit)))))
+      (unwind-protect
+          (embark--act action target)
+        (and (if embark-quit-after-action (not arg) arg)
+             (minibufferp)
+             (embark-quit))))))
 
 (defun embark--become-keymap ()
   "Return keymap of commands to become for current command."
--- a/embark.el
+++ b/embark.el
@@ -664,18 +664,21 @@ Collect buffers."
                (overlay-get minibuffer-message-overlay 'after-string))))
         (when (and msg (string-match-p "\\` *\\[.+\\]\\'" msg))
           (setq msg (substring (string-trim msg) 1 -1)))
-        (run-at-time
-         0 nil
-         (lambda ()
-           (let* ((new-miniwin (active-minibuffer-window))
-                  (minibuf (when new-miniwin (window-buffer new-miniwin))))
-             (set-window-configuration wincfg)
-             (if new-miniwin
-                 (set-window-buffer miniwin minibuf)
-               (when (minibufferp)
-                 (select-window (get-mru-window)))))
-           (message msg)))
-        (abort-recursive-edit)))))
+        (let ((hook))
+          (setq
+           hook
+           (lambda ()
+             (remove-hook 'post-command-hook hook)
+             (let* ((new-miniwin (active-minibuffer-window))
+                    (minibuf (when new-miniwin (window-buffer new-miniwin))))
+               (set-window-configuration wincfg)
+               (if new-miniwin
+                   (set-window-buffer miniwin minibuf)
+                 (when (minibufferp)
+                   (select-window (get-mru-window)))))
+             (message msg)))
+          (add-hook 'post-command-hook hook))))
+    (abort-recursive-edit)))
 
 (defun embark--act (action target)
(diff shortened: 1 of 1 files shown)


oantolin:
This is very nice! I already removed `embark-quit`, and believe that I also fixed the recursive embark-act problem. We can consider reverting my commit and doing this instead, but I'm not so sure I want the window keeping properties of `embark-quit` anymore.

jakanakaevangeli:
Yes, I actually refreshed my mail after submitting this, only to find
the most recent development and discussion. I'm personally all for
removing embark-quit.

In fact, it caused a weird bug that I couldn't reliably reproduce where
it would horizontally scroll my inactive miniwindow so that the eldoc
signatures would only be half visible, haha. I'm sure there would be
plenty of other nasty bugs creeping in. So if you ask me, we shouldn't
revert your commit.

I would definitely consider using post-command-hook in place of
run-at-time though. I'm currently testing with
a simple find and replace of all "run-at-time 0 nil" with
"embark-run-after-command".

    (defun embark-run-after-command (function &rest args)
      (let (hook)
        (setq hook
              (lambda ()
                (remove-hook 'post-command-hook hook)
                (condition-case _
                    (apply function args)
                  (quit))))
        (add-hook 'post-command-hook hook)))

oantolin:
> Yes, I actually refreshed my mail after submitting this, only to find the most recent development and discussion. I'm personally all for removing embark-quit.

Excellent! I'm really sorry that you wasted your time on trying to fix my buggy `embark-quit`, but if you are in favor of its removal, then it's gone!

> I would definitely consider using post-command-hook in place of run-at-time though. I'm currently testing with a simple find and replace of all "run-at-time 0 nil" with "embark-run-after-command".

I have no real reason to prefer `run-at-time`, so if a self-destructing `post-command-hook` works equally well I'd accept use it instead.

So, this removal of `embark-quit` does fix the recursive case of `embark-act`, right? Can we close that issue, @jakanakaevangeli?

jakanakaevangeli:
Yes the issue is fixed now, thanks.

Issue #114, opened 2021-01-16
Proliferation of completions' buffers


protesilaos:
Hello Omar!  I have a reproducible recipe for unintentionally expanding the number of Embark's completion buffers.

1. On `emacs -Q` evaluate those:

```elisp
(load-file "/path/to/embark.el")
(require 'embark)
(add-hook 'minibuffer-setup-hook #'embark-collect-completions-after-delay)
```

2. `C-x b` and then switch to the completions' buffer.
3. Instead of selecting a candidate, do `C-]`.
4. Repeat steps 2 and 3 a few times.
4. `C-x b` again and notice that there are multiple completions' buffers to choose from.

* * *

What I have for the time being is the following, though I assume you will provide a better way of handling this (**bad for recursive minibuffers**):

```elisp
(defun prot-embark-clear-live-buffers ()
  "Remove lingering Embark Collect Completions' buffers.
Add this to `minibuffer-exit-hook'."
  (let* ((buffers (buffer-list))
         (case-fold-search nil)
         (completions
          (cl-remove-if-not (lambda (buf)
                              (string-match "\\*Embark.*Completions.*"
                                            (format "%s" buf)))
                            buffers)))
    (mapc #'kill-buffer completions)))
```

EDIT: A simpler alternative:

```elisp
(defun prot-embark--live-buffer-p ()
  "Determine presence of a linked live occur buffer."
  (let ((buf-link embark-collect-linked-buffer))
    (when buf-link
      (window-live-p (get-buffer-window buf-link)))))

(defun prot-embark-keyboard-quit ()
  "Control the exit behaviour for Embark collect buffers.

If in a live Embark collect/completions buffer, run
`abort-recursive-edit'.  Otherwise run `keyboard-quit'.

This is meant to be bound in `embark-collect-mode-map'."
  (interactive)
  (if (and (derived-mode-p 'embark-collect-mode)
           (prot-embark--live-buffer-p))
      (progn
        (kill-buffer)
        (abort-recursive-edit))
    (keyboard-quit)))
```

oantolin:
I kill the collect completions buffer in the minibuffer-exit-hook, and `abort-recursive-edit` called from outside the minibuffer closes the minibuffer without running the exit hook (if `abort-recursive-edit` is called from the minibuffer, then the exit hook *is* run).

oantolin:
[...]

PR #134, opened 2021-01-26
Kill completion-buffer more reliably

jakanakaevangeli:
Fixes https://github.com/oantolin/embark/issues/114.


oantolin:
Amazing, thanks!

@protesilaos, now `C-]` will close the minibuffer and the collect completions buffer if called from either the minibuffer or from the collect completions buffer!

protesilaos:
Looks good!  I will remove my workarounds and if something goes awry I will let you know.

PR #136, opened 2021-01-27
Use an ephemeral post-command-hook in favor of run-at-time

jakanakaevangeli:
During kmacro execution, functions delayed with `(run-at-time 0 nil #'func)`
will only be called after the whole kmacro. That is why it is preferable to
delay with post-command-hook, making it possible to use non-exiting embark-act
during keyboard macros.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -676,15 +676,25 @@ minibuffer."
     cmd))
 
 (defun embark-run-after-command (function &rest args)
-  "Call FUNCTION with ARGS after currently running command."
-  (cl-labels
-      ((hook ()
-             (remove-hook 'post-command-hook #'hook)
-             (with-demoted-errors "embark-run-after-command PCH: %S"
-               (condition-case _
-                   (apply function args)
-                 (quit (message "Quit"))))))
-    (add-hook 'post-command-hook #'hook)))
+  "Call FUNCTION with ARGS on next command loop iteration."
+  (let ((hook (make-symbol "embark--run-after-command")))
+    (put hook 'embark--run-after-command t)
+    (fset hook
+          (lambda ()
+            (remove-hook 'post-command-hook hook)
+            (with-demoted-errors "embark-run-after-command PCH: %S"
+              (condition-case _
+                  (apply function args)
+                (quit (message "Quit"))))
+            ;; `embark-run-after-command' may have been used inside `function'
+            ;; to delay another function, we should call it here
+            (run-hook-wrapped 'post-command-hook
+                              (lambda (h)
+                                (when (and (symbolp h)
+                                           (get h 'embark--run-after-command))
+                                  (funcall h)
+                                  t)))))
+    (add-hook 'post-command-hook hook)))
 
 (defun embark--act (action target &optional quit)
   "Perform ACTION injecting the TARGET.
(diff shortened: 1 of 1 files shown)


oantolin:
You changed from `run-at-time` to `post-command-hook` in several places, I'll comment on each separately:

- In the `inject` function in `embark--act`: this one is great! Now non-quitting `embark--act` works in keyboard macros!

- In the quitting case of `embark--act`: this gives recursive quitting `embark-act` very odd behavior. Try your old example: in `find-file` use `embark-act` to attempt to copy a file but instead use `embark-act` to `find-file-other-window` some other file. With `run-at-time`, this correctly shows the other window and cancels the copy-file. With `post-command-hook` it still cancels the `copy-file` but does not immediately show the other window, it waits for one more command before showing it (try an arrow key or something).

- In `embark-become`: also odd behavior. I tried a simple `find-file` --> `switch-to-buffer` --> `find-file` sequence. The first become does work, except that it has the wrong default value: with `run-at-time` if I start typing `mess` in `find-file` and become `switch-to-buffer`, the prompt says the default buffer is `*Messages*`, the only one that matches my input. With `post-command-hook` it gave a different default. This is not a big deal, I guess. But if after becoming `switch-to-buffer` I try becoming `find-file` again, it confusingly looks like it quits the minibuffer, but after one more command (try an arrow key) it jumps to `find-file`.

- In `embark--eval-prep`: as far as I can tell it behaves the same way with `run-at-time` and `post-command-hook`.

Maybe it makes sense to use it only for `inject`?

oantolin:
Sorry for the force push, I commited something by mistake to your branch.

oantolin:
Ignore the stuff about the default value in embark-become, I think that may have been due to some other difference in my testing setups. But the point about using embark-become twice remains.

jakanakaevangeli:
I have fixed embark-run-after-command so that it works from within
delayed functions. Tell me if it's too complicated, and we could
consider only using it for injection inside embark--act.
Though personally, I want to try as hard as I can to avoid run-at-time
at least for the quitting case of embark--act, because a future Emacs
version might allow quitting during kmacros, making it possible to use
both quitting and non-quitting versions of embark-act during macros.

oantolin:
It does seems a little complicated (although all the complication is concentrated in a single function, which can be redefined to just call `run-at-time 0 nil` if there is any problem). And since `abort-recursive-edit` stops keyboard macros anyway, how about that for now I just make the `inject` and `embark--eval-prep` functions use `(add-hook 'post-command-hook #'exit-minibuffer nil t)`. Those run in a fresh minibuffer, so a local hook is good enough, we don't need to make it self destruct or anything.

oantolin:
If in the future Emacs does change so that there is a way of quitting a minibuffer without stopping a recording, then we can fix the other cases.

jakanakaevangeli:
> how about that for now I just make the `inject` and
> `embark--eval-prep` functions use
> (add-hook 'post-command-hook #'exit-minibuffer nil t).

Fine with me. I'll leave the branch lying around for the possible
future.

oantolin:
OK, done. 562c6c8541702beac711768f2d78fb0027e3f78e

oantolin:
And again, thanks for all your work on this! I wasn't even aware of that `run-at-time 0 nil` ran after the entire keyboard macro finished replaying. I'm happy to be able to use at least not-quitting actions in keyboard macros now.

PR #138, opened 2021-01-29
Replace consult-buffer with consult-multi category

minad:
See https://github.com/minad/consult/commit/1864b6a942182ea74e9e4b959bea21886e48370e and https://github.com/minad/marginalia/commit/218881ea80476723cb12795ad744867b0d2a3c03. This is a step towards commands with multiple categories (https://github.com/minad/consult/issues/135).

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -200,34 +200,28 @@ The elements of LINES are assumed to be values of category `consult-line'."
 (setf (alist-get 'consult-location embark-exporters-alist)
       'embark-consult-export-occur)
 
-;;; support for consult-buffer
+;;; support for consult-multi
 
-(defun embark-consult-refine-buffer-type (target)
-  "Refine `consult-buffer' TARGET to its real type.
+(defun embark-consult-refine-multi-type (target)
+  "Refine `consult-multi' TARGET to its real type.
 
-This function takes a target of type `consult-buffer' (from
-Consult's `consult-buffer' command) and transforms it to its
-actual type, whether `buffer', `file' or `bookmark', and also
-removes its prefix typing character."
-  (let ((first (- (aref target 0) #x100000)))
-    (if (<= 0 first ?z)
-        (cons (pcase first
-                ((or ?b ?h ?p) 'buffer)
-                ((or ?f ?q) 'file)
-                (?m 'bookmark)
-                (_ 'general))
-              (substring target 1))
-      ;; new buffer case, don't remove first char
-      (cons 'buffer target))))
+This function takes a target of type `consult-multi' (from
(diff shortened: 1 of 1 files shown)


oantolin:
I like the idea of this `consult-multi`, and I haven't tested this code yet, but I thought when we discussed this earlier we said that we cannot rely on text-properties in `embark-act` from the minibuffer under default completion or icomplete. (I think `embark-act` in the embark collect buffer or in the minibuffer but under Selectrum, do work).

minad:
I think it is reliable since we only apply the transformer to the raw candidates obtained from the completion system in `embark--target`. It is the same in Marginalia. There we also access the candidates as is and can use the properties.

oantolin:
Well, I'll test, but I'm pretty sure we already went over this and realized that default completion and icomplete do not preserve the text properties.

oantolin:
Here is where we [discussed this before](https://codeberg.org/jao/elibs/issues/1#issuecomment-168957).

oantolin:
You know, if all embark needs from `consult-multi` is the category name it can be encoded in more tofu.

minad:
Well, it works, I tried it with both icomplete and selectrum :) But please recheck, maybe I missed something or made a mistake.
Maybe I didn't differentiate all the cases properly in the prior discussion - or I relaxed my opinion on text properties by now.

From my understanding, the only point where we are not allowed to access text properties is when accessing the result of completing-read. This happens also when injecting candidates in the minibuffer for example.

But in this case, the transformer is only used here:

~~~ elisp
(defun embark--target ()
  (pcase-let* ((`(,type . ,target)
                (run-hook-with-args-until-success 'embark-target-finders))
               (transformer (alist-get type embark-transformer-alist)))
    (if transformer
        (funcall transformer target)
      (cons type target))))
~~~

This calls `selectrum--get-current-candidate` and this returns the candidate as is. And the same applies to the top-completion target finder. The point is that we are accessing the internals of the completion system in the target finder and there the full properties are present.

If there are problems we could encode the category as tofus. If not, I would try to continue using this approach.

minad:
There is the one scenario where the properties are lost - this is when the completion system for some reason reports a candidate typed in manually by the user. This only happens when we have `:require-match=nil`. But in this case tofus would not be present as well.

oantolin:
You're right! I managed to confuse myself! As the earlier, non-confused version of me explains in that link the moment the properties are lost is when the action receives the argument from `completing-read`. An embark transformer runs before then and does have access to the text properties. I'm sorry for the noise.

minad:
No problem. Thanks or checking again!

PR #148, opened 2021-02-07
Fix #147

### LINKED ISSUE #147: embark--target-window calls display-buffer with nil
Now that #146 is fixed (thanks!) I more consistently see this the first time I perform the default action on an embark collect snapshot of `consult-ripgrep`. If I redo the ripgrep and the snapshot, it works fine.

```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  string-match("^CAPTURE.*\\.org$" nil nil)
  display-buffer-assq-regexp(nil (("^CAPTURE.*\\.org$" (+popup-buffer) (actions) (side . bottom) (size . 0.25) (window-width . 40) (window-height . 0.16) (slot) (vslot) (window-parameters ... ... ... ... ...))  ... #<bytecode 0x1c12ad98950b5225>) #<killed buffer>)
  apply(ad-Advice-display-buffer #f(compiled-function (buffer-or-name &optional action frame) ... #<bytecode 0x1c12ad98950b5225>) #<killed buffer>)
  display-buffer(#<killed buffer>)
  embark--target-window(t)
  embark--act(embark-goto-location "config.org:5005:  (aj-defadvice aj--disable-markdo...")
  embark-collect-choose(#<marker (moves after insertion) at 6075 in *Embark Collect*>)
  button-activate(#<marker (moves after insertion) at 6075 in *Embark Collect*> nil)
  push-button(6075)
  funcall-interactively(push-button 6075)
  call-interactively(push-button nil nil)
  command-execute(push-button)
```

minad:
I propose this fix for #147. Additionally I fixed a package linting complaint about symbol-name and improved the goto-location code, such that it does not print a message and use next-error which does not require the declare-function. I had this goto-location "fix" laying around due to some other experiments.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -326,9 +326,10 @@ window if necessary."
     (minibuffer-selected-window))
    ((and embark--target-window
          (window-live-p embark--target-window)
-         (eq (window-buffer embark--target-window) embark--target-buffer))
+         (or (not (buffer-live-p embark--target-buffer))
+             (eq (window-buffer embark--target-window) embark--target-buffer)))
     embark--target-window)
-   (embark--target-buffer
+   ((and embark--target-buffer (buffer-live-p embark--target-buffer))
     (or (get-buffer-window embark--target-buffer)
         (when display (display-buffer embark--target-buffer))))
    (display (selected-window))))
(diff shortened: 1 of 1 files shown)


oantolin:
Thanks!

oantolin:
I remember trying `next-error` first, but abandoning it for some reason. But I can't remember the reason.

minad:
Okay, I experimented with your version and then switched to next-error for simplicity, see https://github.com/minad/consult-fuzzy/blob/c36b6b21ec2423e90886f663bde5b45727f0ba75/consult-fuzzy.el#L168. I think it should work since next-error just delegates to the function provided by the mode.

oantolin:
The `next-error` version seems to work just fine. I must have done something wrong, mistakenly thought there was a problem, and then switched to `compile-goto-error`.

PR #152, opened 2021-02-10
Avoid with-selected-window for quitting embark-act

jakanakaevangeli:
What do you think?

I believe that `abort-recursive-edit` is pretty accurate at selecting the
correct window and we don't have to help with our own `with-selected-window`.
This will fix `switch-to-buffer-other-window` and `find-file-other-window` as
embark actions (currently, they fail to select the window they display).

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -722,7 +722,6 @@ minibuffer before executing the action."
       (command-execute action)
     (let* ((command embark--command)
            (prefix prefix-arg)
-           (action-window (embark--target-window t))
            (setup-hook (or (alist-get action embark-setup-overrides)
                            embark-setup-hook))
            (allow-edit (if embark-allow-edit-default
@@ -740,24 +739,23 @@ minibuffer before executing the action."
            (run-action (if (commandp action)
                            (lambda ()
                              (minibuffer-with-setup-hook inject
-                               (with-selected-window action-window
-                                 (run-hooks 'embark-pre-action-hook)
-                                 (let ((enable-recursive-minibuffers t)
-                                       (embark--command command)
-                                       (this-command action)
-                                       (prefix-arg prefix)
-                                       ;; the next two avoid mouse dialogs
-                                       (use-dialog-box nil)
-                                       (last-nonmenu-event 13))
-                                   (command-execute action))
-                                 (run-hooks 'embark-post-action-hook))))
+                               (run-hooks 'embark-pre-action-hook)
+                               (let ((enable-recursive-minibuffers t)
+                                     (embark--command command)
+                                     (this-command action)
+                                     (prefix-arg prefix)
+                                     ;; the next two avoid mouse dialogs
+                                     (use-dialog-box nil)
+                                     (last-nonmenu-event 13))
+                                 (command-execute action))
+                               (run-hooks 'embark-post-action-hook)))
                          (lambda ()
-                           (with-selected-window action-window
-                             (run-hooks 'embark-pre-action-hook)
-                             (funcall action target)
-                             (run-hooks 'embark-post-action-hook))))))
+                           (run-hooks 'embark-pre-action-hook)
+                           (funcall action target)
+                           (run-hooks 'embark-post-action-hook)))))
       (if (not (and quit (minibufferp)))
-          (funcall run-action)
+          (with-selected-window (embark--target-window t)
+            (funcall run-action))
         (embark--quit-and-run run-action)))))
 
 (defun embark-refine-symbol-type (target)


minad:
I would prefer if this is done consistently in the quit/non-quit cases.

oantolin:
Sorry, I'm not accepting changes in the semantics of actions, since those are huge incompatible changes. Specially changes that make the semantics different in the quitting and non-quitting cases!

Actions are supposed to run in a window that is displaying the so-called target buffer, which is:
- in case of minibuffer targets, the buffer selected prior to the minibuffer,
- in case of targets from regular text buffers, that buffer,
- in case of targets from `*Completions*` or embark collect buffers, whatever the target buffer was for the targets that wound being listed in the the `*Completions*` or collect buffer.

oantolin:
I think my reaction was a little too visceral. The point is though that which buffer is current when an action runs is an important detail in defining what an action does. To a lesser extent the same is true of which window is selected.

Initially  Embark left this up to chance, but then you'd get differences between running an action from the minibuffer or from an Embark collect buffer (known as occur back then). So at some point I decided to keep track of that so-called target buffer and always run actions in it. Then `consult-line` came along and it didn't like having the current buffer not be displayed in the selected window, so @minad talked me into always ensuring that the selected window displays the target buffer. This went through a few iterations to get the details comfortable (basically trying to reuse the original window displaying the target buffer when sensible).

And I think now the behavior is nice and consistent, actions whether quitting or not, always run in some window showing the target buffer, and if possible this is the window that was selected at the time the target buffer became the target buffer.

But maybe actions are more varied than that, and which buffer & window they run in should be configurable somehow? Already an exception is made for `embark-collect-*` and `embark-export` which really don't make much sense as actions, but there usage as action is supported so you don't need to bind them separately. This exception is hard coded. Maybe other actions deserve the chance to say "I don't care where I run" (like `embark-save`) or "I just need the target buffer, don't care about the window", or even "the target buffer/window isn't important for me, use this function to pick a buffer/window".

minad:
This sounds like some Embark display rules. Probably not a bad idea. But from which actions would it be needed?

oantolin:
> But for which actions would it be needed?

Well, *needed* is a strong word, I am more or less satisfied with things as they are now, but if things were tweakable, I might change the configuration that comes with Embark a bit.

First, I might make the default to be only to use the target buffer and not mess with windows at all. I think that basically we only have the target window for the consult search commands: prior to them embark only set the current buffer and it seemed fine to me. Since people really don't like it when you mess with their window layout, probably just using the target buffer is a less annoying default. (I don't feel too strongly about this.)

If the default did change then, of course, I'd configure the consult search commands to use the target window.

Many actions only need the target and don't need even the target buffer. Configuring those to not use the target buffer makes them usable even after you kill the buffer. Particularly saving on the kill ring, and basically all the file and buffer actions come to mind. This would also fix the window selection problem that @jakanakaevangeli mentioned for `switch-to-buffer-other-window` and `find-file-other-window` used as quitting actions.

For `embark-insert` I don't actually like the target buffer in every case. It's right for when you use that action from the minibuffer or from an embark collect buffer, but from a regular text buffer I think it would be better to insert in the buffer of the previously selected window.

I am not completely sure it is worth making this configurable, but maybe. I guess a nice interface would be an alist associating actions to functions that return either nil (meaning just run the action, don't worry about where), a buffer or a window. In most cases you want `nil`, `embark--target-buffer` or `(lambda () (embark--target-window t))` (I guess I'd make `t` the default so people could use `embark--target-window`.)

jakanakaevangeli:
> For embark-insert I don't actually like the target buffer in every case
This is surprisingly already configurable:
```elisp
(define-key embark-general-map "i" 'embark-insert-other-window)

(defun embark-insert-other-window (string)
  (interactive "sInsert: ")
  (with-selected-window (other-window-for-scrolling)
    (embark-insert string)))

(defvar embark-general-minibuffer-map
  (make-composed-keymap (make-sparse-keymap) embark-general-map))
(define-key embark-general-minibuffer-map "i" 'embark-insert)

(defun embark-general-minibuffer-map-h ()
  (setq-local embark-general-map embark-general-minibuffer-map))
(add-hook 'minibuffer-setup-hook #'embark-general-minibuffer-map-h)
(add-hook 'embark-collect-mode-hook #'embark-general-minibuffer-map-h)
```

What would you think about the following less intrusive fix for the
`*-other-window` problem I reported: instead of `with-selected-window`, we
could use a similar but weaker macro that would restore the selected window,
but only if it detects that its body didn't switch to another window. That way,
the behaviour wouldn't change for most commands, only commands that explicitly
select another window.

The downside to think about here is that a lot of commands such as `shell` or
`dictionary-search` use `pop-to-buffer` which usually does select another
window and it would be annoying to have these commands take focus away from the
minibuffer in the non-quitting embark-act. Perhaps this weaker macro should be
only used for the quitting case.

oantolin:
I think I like your proposed weaker macro. And I'm not sure it would really be annoying for cases like `shell` or `dictionary-search` when used as non-quitting actions. Take `shell` for example: why would you use it as a non-quitting action? Probably you intend to run a command or two right away and return to the minibuffer afterwards, right? I guess you could be just running it to open a shell that you'll use later, but this seems less likely since the window configuration is restored when you quit the minibuffer, so while the shell buffer would be open, you'd need to go hunt for it ---and in that case, why not just wait to run shell until after you are done with the minibuffer.

I think I'll try that and see how comfortable it is.

oantolin:
I've implemented you suggestion of not restoring the orginal window if actions select a different window, @jakanakaevangeli. 659df4bb9b9ac9abbb1afddcae35564137be0d9d I'm liking it so far. (The case that really annoyed me was `ediff`, it's much better this way.) We'll see if it causes any trouble down the line.

jakanakaevangeli:
Thanks you!

PR #153, opened 2021-02-11
Add embark-toggle-highlight

minad:


minad:
I just observed a weird embark exiting issue, when executing `C-. M-x embark-toggle-highlight`, before I had bound `embark-toggle-highlight` to `H`. It seems `embark-toggle-highlight` is executed in the minibuffer then?

oantolin:
It doesn't toggle for me! It behaves just like `highlight-symbol-at-point`. On the other hand, running from `M-x` does seem to work here (I mean, it doesn't toggle, but it does highlight new symbols). I thought for a second I was running `highlight-symbol-at-point`, but did carefully check it was `embark-toggle-highlight`.

I'll do some debugging.

oantolin:
Ah, in my Emacs the keys of `hi-lock-interactive-patterns` are not regexps but rather closures that look for the regexp.

```
hi-lock-interactive-patterns is a variable defined in ‘hi-lock.el’.
Its value is
((#f(compiled-function
     (limit)
     #<bytecode 0x1783bf5d2181864d>)
    (0 'hi-yellow prepend)))
Local in buffer embark.el; global value is nil
```

And the disassembly of the closure:

```
byte code:
  doc:   ...
  args: (arg1)
0       constant  t
1       constant  nil
2       varbind   search-spaces-regexp
3       varbind   case-fold-search
4       constant  re-search-forward
5       constant  ("\\_<hi-lock-interactive-patterns\\_>")
6       car-safe  
7       stack-ref 2
8       constant  t
9       call      3
10      unbind    2
11      return    
```

(This is a little meta: I ran `embark-toggle-highlight` on the variable `hi-lock-interactive-patterns`, which is why it appears in the regexp in the disassembly.)

minad:
👀 maybe something has changed in Emacs 28?

oantolin:
I think so. I'm using (an outdated) 28.

minad:
Independent from the Emacs 28 change, for me `C-. M-x highlight-symbol-at-point` does not work, where `C-.` is `embark-act`.

minad:
Emacs 28:

~~~ elisp
        (completing-read (format-prompt "Regexp to unhighlight" (car defaults))
                         (mapcar (lambda (pattern)
                                   (cons (or (car (rassq pattern hi-lock-interactive-lighters))
                                             (car pattern))
                                         (cdr pattern)))
                                 hi-lock-interactive-patterns)
			 nil t nil nil defaults))))))
~~~

Emacs 27.1:

~~~ elisp
        (completing-read (if (null defaults)
                             "Regexp to unhighlight: "
                           (format "Regexp to unhighlight (default %s): "
                                   (car defaults)))
                         hi-lock-interactive-patterns
			 nil t nil nil defaults))))))
~~~

:roll_eyes:

oantolin:
Do you have M-x rebound? To smex maybe?

minad:
No I have a rather bare bone setup, no smex nothing (smex/amx is obsolete by prescient and consult-mode-command btw ;) - I am also not using prescient.

I suggest you revert my broken PR, and I find a better solution. And I will try to investigate the M-x issue a bit.

minad:
Ouch, this hi-lock thing does not seem to be particularly trustworthy. It uses an exclusion list for untrusted modes, this does not instill confidence :cold_sweat: 

~~~ elisp
(defcustom hi-lock-exclude-modes
  '(rmail-mode mime/viewer-mode gnus-article-mode)
  "List of major modes in which hi-lock will not run.
For security reasons since font lock patterns can specify function
calls."
  :type '(repeat symbol)
  :group 'hi-lock)
~~~

oantolin:
Do you have a variable `hi-lock-interactive-lighters`? I have it and just replacing `hi-lock-interactive-patterns` with it make the toggle work here.

minad:
No I don't have this variable.

oantolin:
I'll just use use the `-lighters` variable if available and `-patterns` if not: I bet before `-lighters` was introduced the `-patterns` variable did have regexp keys.

oantolin:
OK, done. 558df851d6b9e67607e5892dce8237af01275eb6

oantolin:
The `M-x` problem you see is weird. It doesn't happen here.

minad:
I will investigate this soon.

oantolin:
I also wonder why `highlight-symbol-at-point` doesn't do toggling itself, that would be a good improvement.

oantolin:
Don't know why I used `(cl-find t ... :test #'boundp)` instead of `(cl-find-if #'boundp ...)`. 😆 f0c63c957cc7013ff75fba99700e3fb85b89d300

minad:
> I also wonder why highlight-symbol-at-point doesn't do toggling itself, that would be a good improvement.

Agree, this should be proposed on the emacs issue tracker. Maybe it could also be a separate command. But for Embark we would still need the detection of the legacy function for a while.

PR #156, opened 2021-02-11
embark-consult--strip-prefix: use consult--tofu-char

minad:
+ minor docstring corrects to make package lint happy

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -137,7 +137,9 @@ associated to an active minibuffer for a Consult command."
 (defun embark-consult--strip-prefix (string)
   "Remove the unicode prefix from a consult-location STRING."
   (let ((i 0) (l (length string)))
-    (while (and (< i l) (<= #x100000 (aref string i) #x10fffd))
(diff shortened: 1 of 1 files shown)


oantolin:
Funny, I could have sworn I did use consult--tofu-char

minad:
Yes you used it at another place.

oantolin:
Ah, that must be it. Thanks for spotting this one.

minad:
It just occurred to me that this function can even go away since the transformer has access to text properties. This means we can strip the consult-location property and the invisible property in the case of consult-isearch. Maybe that would be preferable?

oantolin:
Oh, you mean, change the transformer so instead of looking for initial tofu it looks specifically for characters with either a `consult-location` or `invisible` property? I guess that's a little better. Make the property names part of the API instead of reserving a chunk of unicode as part of the API.

oantolin:
Is this what you had in mind, @minad? 34e9070140f63627929a88503659fe82bef75d46

Issue #146, opened 2021-02-07
Default action for consult-*grep doesn't work if you bind directly to `embark-collect-snapshot`


aaronjensen:
Similar issue to #80, though it only reproduces if you have a key bound to `embark-collect-snapshot` and use it, rather than `embark-act`. 

I'm guessing it's because `embark-act` invokes `embark--target`, which sets `embark--command`. Is binding to `embark-collect-snapshot` not supported or is this a bug?

oantolin:
Calling `embark-collect-snapshot` from a key binding definitely *is* supported. I recently changed some things in embark-consult, and must have broken this then. I'll look into it. in the meantime, as a workaround, maybe you can use `embark-export` instead.

aaronjensen:
Great, good to know. For now I'm just using:

```elisp
(defun aj/embark-collect-snapshot ()
    "Embark collect snapshot without act"
    (interactive)
    (embark--target)
    (command-execute #'embark-collect-snapshot))
```

oantolin:
Oh, I didn't screw this up in the recent changes, I must have screwed this up a while ago. :D

If you use `embark-act` and then `RET`, the default action does work. It only fails if you use `RET` on its own, because I forgot to call the functions from `embark-transformer-alist` in that case.

oantolin:
> I'm guessing it's because `embark-act` invokes `embark--target`, which sets `embark--command`

By the way, that wasn't exactly why `embark-act` worked around the bug. It's not `embark-act` that sets `embark--command`, that is set in a minibuffer setup hook, because `embark--command` is whatever command opened the minibuffer, i.e., the value of `this-command` when the minibuffer opens.

So `embark--command` was correctly begin set to `consult-grep`. What the default action does is run `embark--command` on the target, i.e., feed the target back into the command you got it from. But the `consult-grep` commands are very peculiar: the completion candidates they produce are *not* valid inputs for the command, the candidates are in `grep`'s `file:lineno:text` format, not in the `#args#filter` format you type at the minibuffer.

They are in fact the only commands I use that offer completions that are not valid input! So, for the `consult-grep` family I have a target transformer that sets `embark--default-action` to something different from `embark--command`.

Calling that transfomer via `embark--target` is the reason your `aj/embark-collect-snapshot` function worked around the bug.

minad:
> They are in fact the only commands I use that offer completions that are not valid input! So, for the consult-grep family I have a target transformer that sets embark--default-action to something different from embark--command.

[...]

PR #158, opened 2021-02-19
Fix embark-goto-location with previous next-error

aaronjensen:
In cases where `next-error` was used in another buffer prior to doing an
`embark-collect-snapshopt`, that buffer will be preferred when doing
`next-error`. By setting the `next-error-find-buffer-function`, we override that
behavior

Fixes the issue I mentioned here: https://github.com/oantolin/embark/issues/146#issuecomment-779989307

Another approach is to set `next-error-last-buffer`, but this is more direct and `next-error-find-buffer-function` has higher precedence.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -1895,7 +1895,9 @@ Return the category metadatum as the type of the target."
       (insert location "\n")
       (grep-mode)
       (goto-char (point-min))
-      (let ((inhibit-message t))
+      (let ((inhibit-message t)
+            (next-error-find-buffer-function
+             (lambda (&rest _args) (current-buffer))))
         (next-error 0 t)))))
 
 (defalias 'embark-execute-command


aaronjensen:
Also, I understand better now what this is doing and it's definitely inferior to a regular grep mode buffer. Specifically, because it uses a temp buffer with only a single line, one of `next-error`'s greatest features (imo) is unable to be leveraged. That is, you cannot just go to the next match with a key from any other buffer.

Ironically, this is the same feature that caused this issue in the first place.

So, I think I'll be just using export when I run a grep so I can get that behavior natively.

minad:
@oantolin Go back to the old compile mode specific function instead? I guess this was the problem you had observed?

oantolin:
> Also, I understand better now what this is doing and it's definitely inferior to a regular grep mode buffer. Specifically, because it uses a temp buffer with only a single line, one of `next-error`'s greatest features (imo) is unable to be leveraged. That is, you cannot just go to the next match with a key from any other buffer.
 
> So, I think I'll be just using export when I run a grep so I can get that behavior natively.

Absolutely, @aaronjensen! The embark collect buffers are not supposed to reimplement all the features of the corresponding embark export buffers, they are just supposed to be some least common denominator base functionality that works for every type of targets. It is expected than whenever there is an exporter for a particular type, the export buffers are more featureful than the collect buffers.

Basically collect gives functionality that can be made to work uniformly independently of the type of targets; while export gives access to functionality that is perfectly adapted to one specific type of targets.

I'll compare this patch to @minad's suggestion of returning to my old implementation and pick one or the other.

aaronjensen:
Awesome, thanks.

It looks like the only way to detect if the current minibuffer command is a grep command is to actually attempt the export, is that right? I'm probably missing something here. I was trying to write something that'd detect if I'm in a grep and export, otherwise collect snapshot. I think actually I may just always export and see how that goes for a while.

oantolin:
> It looks like the only way to detect if the current minibuffer command is a grep command is to actually attempt the export

You can test for the `xref-location` category, which is the one the grep exporter is used for. Use this expression:
`(eq (completion-metadata-get 'category (embark--metadata)) 'xref-location)`

> I think actually I may just always export and see how that goes for a while.

Export falls back to collect if no specific exporter is available, so I'd actually recommend this if you always want the maximum functionality. Of course, if you, for example, don't like dired for some reason, and never want to use it, you can just remove the dired exporter from `embark-exporters-alist`.

oantolin:
I decided to go with the old implementation using `compile-goto-error`, since it seems more direct. 9b9aaff174a55eba76eb1b199dfdc4a52400c91e Please test, @aaronjensen.

aaronjensen:
Hm, it works, but it opens two windows when I press enter for some reason.

oantolin:
I saw the two window problem! But when I restarted Emacs to debug from a clean state, I couldn't reproduce the problem any more. I still can't reproduce it. So, maybe restart Emacs? It might also cure the problem for you. I know that's terrible advice, and I should figure out what caused it instead, but since I can't reproduce it figuring it out seems difficult.

aaronjensen:
It reproduces for me after a restart

aaronjensen:
Yeah, I can reproduce it pretty consistently. The embark snapshot window cannot be the only window. It must be the right window, I believe.

oantolin:
OK, thanks for the tip, I'll look into it.

oantolin:
Yep, that reproduces the problem, @aaronjensen! I'll see if I can figure it out.

oantolin:
I think this does it. 040eb317bc422477ddb4eeff4a565d1e3c9024b6

aaronjensen:
Hmm, I can still reproduce it in the same way even with that commit.

oantolin:
Oh, I really thought I had it, @aaronjensen. Here's what I saw happen with my config: first, before that 040eb31 commit, if I followed your recipe, that is, if I opened two windows, ran `consult-grep` and then `embark-collect-snapshot`, and then pressed `RET` in the collect buffer, I would see the other window split in two and *both* show the result I pressed `RET` on. After that commit, however, I see the non-collect window split in two, only *one* of which shows the result I pressed `RET` on. And I think that's correct behavior, no?

You still see both windows showing the search result?

aaronjensen:
Yes, I see both windows show the search result, but only if my emacs window is big enough... if it's small, it doesn't do it it seems. In the attached gif, I first do it with a smaller emacs window and then I do it with a full size (on my 16" laptop) window.


CleanShot 2021-02-20 at 20 25 17

oantolin:
Thanks for the gif. At some point in the past, `embark-goto-location` always displayed it's search result in the target window, which seemed (a) consistent and predictable, (b) sensible, since usually the target window is incidental to a command that searches many files. I think that was @minad's idea, but I removed it, probably without much thought. I brought that feature back now, and hope that finally takes care of this problem. c23bb53054baab8b927eae7d818d446b39a8e92c

aaronjensen:
Looks like it's working for me now, thanks!

oantolin:
> Looks like it's working for me now, thanks!

Great! And thanks for your patience, @aaronjensen.

PR #167, opened 2021-02-23
thin separator lines

minad:
Draft for thin line separators. Either replace zebra or add a toggler which allows to toggle between nothing, zebra and lines? But if zebra is kept it would be nice to improve it to fix the background precendence issue.


minad:
Note that this is just a draft, I wanted to see if it works nicely with the overlays.

Issue #168, opened 2021-02-25
Comments on Embark Key Bindings


hmelman:
From comments in [Consult Issue #210](https://github.com/minad/consult/issues/210) I took a look at the default embark bindings.  Here are some comments, though I don't think any of these are particularly important, they're just things I thought of that might be worth considering.

- I agree that the find-definition commands should move from "d" to "."
- I think that `delete-file` should stay on "d" as other delete commands are on "d" and it's natural.
- The package and bookmark maps move `embark-insert` to "I" so maybe it should just be that in the general map.  Then the three other maps that bind commands to "I" can bind them to the more natural "i" and "I" can be more consistent.
- The region map binds `fill-region` to "f".  That's fine, but the other obvious choice is "q" since M-q is `fill-paragraph` and I think q is less likely to have a future conflict than f.
- Maybe bind "*" in the region map to `calc-grab-region`.  I'd love it if `calc-grab-sum-down` and `calc-grab-sum-across` were bound to ":" and "_" but they're not autoloaded and little known.  They would be more convenient from embark than their existing bindings on C-x * :.  But they're probably better in personal configs, maybe it's with a comment in a readme or wiki.
- It's fine that `embark-browse-package-url` is on "u" but FWIW paradox puts the equivalent on "v" for visit homepage

oantolin:
`calc-grab-region` should automatically detect whether the region is a rectangle and if so act as `calc-grab-rectangle`. 😠

oantolin:
I think you are right about moving `embark-insert` to `I`, to free up `i` which several other commands seem to want. There are a couple of things giving me pause:

- Should I then also move `embark-save` to `w` for consistency? I think probably the answer is no: I think `embark-save` is probably much more common an action than `embark-insert` so it's OK if it gets a prized lowercase binding while `embark-insert` needs a modifier.

- In the file map there are actions to insert or save a relative path bound to `W` and `I`. If `embark-insert` moved to `I`, would I bind `embark-insert-relative-path` to `i`? That would leave it flipped with respect to `w/W`, which sounds like a bad idea.

Any thoughts?

hmelman:
Assuming `embark-insert` moves to `I`.   `embark-save` could move for consistency, that could the relative problem (both on lowercase).  Or leave save on `w` and use something else like `r` or `C-i` for file's relative version.

hmelman:
[...]

PR #177, opened 2021-03-03
embark-find-definition: push markers

leungbk:
This pushes an appropriate entry onto the xref marker stack, allowing `pop-tag-mark` to work.

`xref--push-markers` additionally pushes something onto the mark ring, though I'm not sure what the effect is supposed to be; `M-x pop-global-mark` works either way:

https://github.com/emacs-straight/xref/blob/a3b455273d751dbe80abc8f0e9bb5a8c78abb746/xref.el#L1148-L1151

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -1921,6 +1921,7 @@ When called with a prefix argument, open dired in another window."
 (defun embark-find-definition (symbol)
   "Find definition of SYMBOL."
   (interactive "SSymbol: ")
+  (xref-push-marker-stack)
   (cond
    ((fboundp symbol) (find-function symbol))
    ((boundp symbol) (find-variable symbol))))


oantolin:
Great idea, thanks!

oantolin:
Man, I really wanted this without knowing I wanted it. Thanks again!

By the way, what do you think about the suggestion (in #168) to move the `embark-find-definition` action to `.` instead of the current `d`?

leungbk:
I'm fine with that.

minad:
Can someone explain why xref uses its own marker stack? There is the global mark stack and we also have pop-global-mark and consult-global-mark. Should I also add consult-xref-mark? Or maybe the consult-global-mark should show both rings?

minad:
I've also found some discussion about org using its own mark ring, see https://lists.gnu.org/archive/html/emacs-orgmode/2018-02/msg00324.html "Could we get rid of Org specific mark ring".

Issue #186, opened 2021-03-22
Embark keymap prompter needs a few improvements


minad:
Since I am trying to use your Embark prompter now as general prefix command help command, there are a few issues. This is not unexpected I guess, since the Embark keymaps are somehow well-formed and now I am throwing more weird stuff at it.

* Simple menu items are not supported (desc . cmd). Throws an error as of now. This would be nice for Consult Narrowing.
* Complex menu items are probably also not supported (I haven't tested this)
* Consult narrowing keys use these which-key pseudo keys, the Embark prompter thows an error. These bindings should be ignored somehow.

EDIT: I can also provide patches to fix this. I just want to document the things somewhere which do not yet work for me.

oantolin:
Yeah, let's make this robust and general purpose. As you said, needing to work only for Embark keymaps means lots of edge cases had never been tested before.

oantolin:
Ah, I figured out how you are testing this. You have a consult narrow key set! (I do too, it just took a while to click.)

minad:
Using this patch works for me:

~~~~ diff
diff --git a/embark.el b/embark.el
index 25fc7be..db7e557 100644
--- a/embark.el
+++ b/embark.el
@@ -634,7 +634,8 @@ first line of the documentation string; otherwise use the word
-   (or
-    (when (symbolp cmd) (symbol-name cmd))
-    (when-let ((doc (documentation cmd)))
-      (save-match-data
-        (when (string-match "^\\(.*\\)$" doc)
-          (match-string 1 doc))))
-    "<unnamed>")))
+   (cond
+    ((stringp (car-safe cmd)) (car cmd))
+    ((symbolp cmd) (symbol-name cmd))
+    ((when-let (doc (and (functionp cmd) (documentation cmd)))
+       (save-match-data
+         (when (string-match "^\\(.*\\)$" doc)
+           (match-string 1 doc)))))
+    (t "<unnamed>"))))
@@ -653,0 +655,3 @@ first line of the documentation string; otherwise use the word
+                   ;; Filter which-key pseudo keys
+                   ;; TODO more general filter?
+                   unless (eq (car-safe cmd) 'which-key)
~~~

oantolin:
Excellent, thanks a lot! I was on my way to basically the same patch so I'll just use yours.

oantolin:
[...]

PR #190, opened 2021-03-23
Completing read prompter

minad:
A few more steps towards #186. Unfortunately no Marginalia annotation and no non-general Embark actions for now.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -383,6 +383,9 @@ This function is meant to be added to `minibuffer-setup-hook'."
 
 ;;; Internal variables
 
+(defvar embark--prompter-history nil
+  "History used by the `embark-completing-read-prompter'.")
+
 (defvar-local embark-collect--kind nil
   "Kind of current collect buffer.
 
@@ -689,7 +692,7 @@ first line of the documentation string; otherwise use the word
                   (if (eq action 'metadata)
                       `(metadata (category . command))
                     (complete-with-action action commands string predicate)))
-                nil t))
+                nil t nil 'embark--prompter-history))
              commands))
       (`(,cmd ,key ,_desc)
        (setq last-command-event (seq-elt key (1- (length key))))
--- a/embark.el
+++ b/embark.el
@@ -645,8 +645,7 @@ first line of the documentation string; otherwise use the word
 
 (defun embark-completing-read-prompter (keymap)
   "Prompt via completion for a command bound in KEYMAP."
-  (let* ((def (lookup-key keymap [13]))
-         (commands
+  (let* ((commands
           (cl-loop for (key . cmd) in (embark--all-bindings keymap)
                    for name = (embark--command-name cmd)
                    ;; Filter which-key pseudo keys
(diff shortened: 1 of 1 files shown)


embark.el
```diff
@@ -692,7 +693,8 @@ first line of the documentation string; otherwise use the word
                   (if (eq action 'metadata)
                       `(metadata (category . command))
                     (complete-with-action action commands string predicate)))
-                nil t nil 'embark--prompter-history))
+                nil t nil 'embark--prompter-history
+                (and def (symbol-name def))))
```
oantolin: I think it would be better to use embark--comand-name rather than symbol-name, to allow for lambdas, simple menu items, etc.

oantolin: Ah, this is obsoleted my a later commit, forget it.

oantolin:
It would be a tiny bit of code duplication between here and Marginalia, but I think maybe we should handle the annotations right here (specially since it's simpler than the Marginalia annotator, because we wouldn't include the keybinding).

Then we'd add an transformer that chops off the keybinding and reports type `command`, and we'd be set.

oantolin:
I can merge this and add the annotation and transformer myself.

minad:
> It would be a tiny bit of code duplication between here and Marginalia, but I think maybe we should handle the annotations right here (specially since it's simpler than the Marginalia annotator, because we wouldn't include the keybinding).

I think it is better to handle it in Marginalia, since the function `marginalia--function-doc` is non-trivial.

> Then we'd add an transformer that chops off the keybinding and reports type command, and we'd be set.

Yes. We should do this to regain the actions. Use 'symbol or better use 'command?

minad:
Instead of chopping of the keybinding I would attach a symbol 'embark-command to the formatted string. Then we can use that in the annotator and the transformer.

oantolin:
> I think it is better to handle it in Marginalia, since the function `marginalia--function-doc` is non-trivial.

I had forgotten about skipping the lines of the docstring reporting advice. I agree it would be better to do this in Marginalia then.
 
> Yes. We should do this to regain the actions. Use 'symbol or better use 'command?

Surely `command` is better since these are commands! The `command` category gives you access to all the symbol actions plus binding to a key locally/globally and `Info-goto-emacs-command-node`.

oantolin:
> Instead of chopping of the keybinding I would attach a symbol 'embark-command to the formatted string. Then we can use that in the annotator and the transformer.

OK, yes, that would be better.

minad:
> Instead of chopping of the keybinding I would attach a symbol 'embark-command to the formatted string. Then we can use that in the annotator and the transformer.

When attaching the 'embark-command we should also handle menu items better, such that we do not confuse the desc with the cmd of `(desc . cmd)`.

oantolin:
Do you want to do it before I merge? Or should I merge and then do the rest myself? Whatever you want.

oantolin:
> When attaching the 'embark-command we should also handle menu items better, such that we do not confuse the desc with the cmd of `(desc . cmd)`. Then the `consult-narrow` bindings also get a Marginalia annotation.

What do you mean exactly. I imagined the `embark-command property being set to `cmd` in the `(desc . cmd)` case.

minad:
> What do you mean exactly? I imagined the embark-command property being set to cmdin the(desc . cmd)` case.

Yes, I mean exactly that.

> Do you want to do it before I merge? Or should I merge and then do the rest myself? Whatever you want.

If you have time now, feel free to take over.

oantolin:
> If you have time now, feel free to take over.

OK, will do! Thanks for the changes you made! One tiny thing I liked is not taking the `cdr` of the result of `assoc`, might as well reduce indentation and handle it in the `pcase`. I keep the lines pretty short in Embark because I often don't maximize my Emacs.

minad:
I feel the embark completing read prompter is really a neat missing piece. We often had these requests before for `consult-descbinds` and I never understood why one would want this (I still don't), but using this as `prefix-help-command` is just the perfect fit. Maybe we should still rename the function to `embark-bindings`, make it a command and pass the current keymap as default to cater for the `consult-descbinds` needs?

oantolin:
I added the transformer. a103cfdf311167835145b236be4c9b40d700af6c And will now add the annotator to Marginalia.

oantolin:
Do you think in Marginalia the annotator should be considered light or heavy?

minad:
> Do you think in Marginalia the annotator should be considered light or heavy?

Hmm, I think it is a heavy one, since for commands we generally don't add docstrings. 

Maybe the light/heavy idea was a bad one, it seems everyone is configuring Marginalia to just use the heavy annotators :laughing:

oantolin:
I wouldn't be against removing the light/heavy distinction. Do people cycle? Maybe that whole thing is overkill. Maybe there should be a single alist of annotators. If people want cycling, they can write a command that swaps out the entire alist.

oantolin:
> Hmm, I think it is a heavy one, since for commands we don't add docstrings in the light variant.

The keybindings buffer doesn't have docstrings either, and this for a replacement of it, so I guess that also points to it being heavy.

oantolin:
`when-let` can handle a single set of parens, `(when-let (a b) ...)` rather than `(when-let ((a b)) ...)`? Weird. I'll use it to blend in.

minad:
> Do people cycle? 

Well, I cycle - when testing the heavy annotators since I am stubbornly still using the light annotators. But after I added `x-group-function` the heavy annotators are less annoying since the ugly misalignment in `consult-buffer` is nicely separated by the grouping lines.

> Maybe there should be a single alist of annotators. If people want cycling, they can write a command that swaps out the entire alist.

Yes, I am considering this.

oantolin:
OK, done. minad/marginalia@f3eaa1a0372c54295a3ae9b800527d58c991ff3d.

oantolin:
You've been handling Marginalia much more than I have, feel free to change names or whatever, if I didn't get the style right.

oantolin:
> I feel the embark completing read prompter is really a neat missing piece. We often had these requests before for `consult-descbinds` and I never understood why one would want this (I still don't), but using this as `prefix-help-command` is just the perfect fit. Maybe we should rename the `embark-prefix-help-command` to `embark-bindings` and support the case where the command is called without prefix to cater for the `consult-descbinds` needs.

Yes, it's great as a value for `prefix-help-command`. I'm not sure I'd ever run it from top level, it seems that that's what `M-x` is for... 😆 

I'd be OK with the renaming, we'd just need to figure out a good way to tell if it has been called from a prefix or not.

minad:
> Yes, it's great as a value for prefix-help-command. I'm not sure I'd ever run it from top level, it seems that that's what M-x is for... laughing

I think the point is to explore keybindings. That's valuable additionally to M-x.

oantolin:
Say your M-x shows keybindings (maybe you're a Marginalia user or you use the `*Completions*` buffer). Then the value in a descbinds command is that it skips commands that are unbound? Oh, and that you can match against the bindings (which seems like a weird thing to want to do). I'm not sure I see the point, but OK.

minad:
I thought a bit about the light/heavy annotators. One point is also that you may want to use Embark *without annotators*. This only works if you set `marginalia-annotators` to `'(nil ...)`. So there is some value in the indirection. 

The light/heavy distinction could be considered kind of unnecessary however. But I must also say that I never missed heavy annotations when running `consult-buffer`, `find-file` or `M-x`, since for these commands I am usually doing a quick action, and not some kind of exploration. I think the situation could change if we get https://github.com/oantolin/orderless/issues/30, such that I could for example match all elisp buffers via `@elisp`. Then I may want to see the buffer metadata always.

minad:
> Then the value in a descbinds command is that it skips commands that are unbound? Oh

Yes.

> Oh, and that you can match against the bindings (which seems like a weird thing to want to do).

Yes.

> I'm not sure I see the point, but OK.

I also don't see it. In the context of Consult I argued that one should rather use `describe-bindings` + `consult-focus-lines` to achieve a similar filtering. But in combination with the Marginalia annotators we get a little more.

For me the main point now is that we get the functionality basically for free since we already have the `embark-prefix-help-command`. At this point I think it is just worth to have this since the cost is so low. Having it as a separate consult command didn't make sense. But here we already have all the pieces!

oantolin:
> At this point I think it is just worth to have this since the cost is so low. Having it as a separate consult command didn't make sense. But here we already have all the pieces!

Agreed.

PR #191, opened 2021-03-23
Add embark-bindings command

minad:
As discussed in #186 and #190. We get the command in 10 lines of code. I think that's worth it :)

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -644,8 +644,9 @@ first line of the documentation string; otherwise use the word
            (match-string 1 doc)))))
     (t "<unnamed>"))))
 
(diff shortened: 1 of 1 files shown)


oantolin:
I was thinking of how to do this with a single command, but I guess this much easier. The PR looks good to me now, is there anything else you wanted to do first?

minad:
I have a few minor improvements. I ping you again, when I think it is ready.

oantolin:
OK.

minad:
Okay, I think it is good like this now.

oantolin:
Great, thanks!

PR #192, opened 2021-03-24
Prompter config and BUG

minad:
I would like to have the @ configurable via a defcustom. 

Unfortunately I just found that out that the @ prompter is broken since we switched to the formatted candidates. :cry: It also cannot not work with narrowing since we have the disambiguation problem again. it seems the prompter protocol must be changed such that the prompter returns a `(cons key cmd)` instead of only the `cmd`? I haven't done that yet here. How would you like to repair this?

### DIFF
--- a/README.org
+++ b/README.org
@@ -241,8 +241,23 @@ reasonable starting configuration:
 
   (use-package embark
     :ensure t
+
     :bind
-    ("C-S-a" . embark-act))              ; pick some comfortable binding
+    (("C-S-a" . embark-act)       ;; pick some comfortable binding
+     ("C-h B" . embark-bindings)) ;; alternative for `describe-bindings'
+
+    :init
+
+    ;; Optionally replace the key help with a completing-read interface
+    (setq prefix-help-command #'embark-prefix-help-command)
+
+    :config
+
+    ;; Hide the mode line of the Embark live/completions buffers
+    (add-to-list 'display-buffer-alist
+                 '("\\`\\*Embark Collect \\(Live\\|Completions\\)\\*"
+                   nil
+                   (window-parameters (mode-line-format . none)))))
 
   ;; Consult users will also want the embark-consult package.
   (use-package embark-consult
--- a/embark.el
+++ b/embark.el
@@ -184,6 +184,13 @@ prompts for an action with completion."
                         embark-completing-read-prompter)
                  (function :tag "Other")))
 
+(defcustom embark-keymap-prompter-key "@"
+  "Key to switch to the keymap prompter from `embark-completing-read-prompter'.
+
+The key must be either a string or a vector.
+This is the key representation accepted by `define-key'."
+  :type '(choice key-sequence (const nil)))
+
 (defface embark-keybinding '((t :inherit success))
   "Face used to display key bindings.
 Used by `embark-completing-read-prompter' and `embark-keymap-help'.")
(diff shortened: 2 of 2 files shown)


oantolin:
The @ stuff only broke because we changed the format of the completion candidates, I'll just merge this and then update it to the new format.

oantolin:
Instead of changing the return type of the prompters I decided to handle the key lookup locally. I think it's a less disruptive change ---and even with `(cons key cmd)` I would have need to construct the formatted candidate or make more changes.

oantolin:
Looks like everything is now working like it did before making the keybindings part of the candidate. Please test.

oantolin:
Thanks for making the key configurable, by the way. That's a very sensible thing to do.

jaor:
On Wed, Mar 24 2021, Omar Antolín Camarena wrote:

> Thanks for making the key configurable, by the way. That's a very
> sensible thing to do.

A feature request, if i may: it'd be great if there was an option to
customize it to "nothing", in the sense that then the prompter would
start automatically listening to keys instead of completing.  if i'm not
mistaken, right now setting it to nil just makes access by key
unavailable.

oantolin:
I'm a little confused, @jaor, are you suggesting that when you set `embark-keymap-prompter-key` to nil, then `embark-key-map-prompter` should not use `completing-read` at all and instead behaves like `embark-keymap-prompter`. I must be misunderstanding, because that sounds like a pretty odd request.

oantolin:
I'm a little confused, @jaor, are you suggesting that when you set `embark-keymap-prompter-key` to nil, then `embark-completing-read-prompter` should not use `completing-read` at all and instead behave like `embark-keymap-prompter`? I must be misunderstanding something, because that sounds like a pretty odd request.

jaor:
On Wed, Mar 24 2021, Omar Antolín Camarena wrote:

> I'm a little confused, @jaor, are you suggesting that when you set
> embark-keymap-prompter-key to nil, then
> embark-completing-read-prompter should not use completing-read at all
> and instead behave like embark-keymap-prompter? I must be
> misunderstanding something, because that sounds like a pretty odd
> request.

mostly, yes.  so, what i want is a prompter showing me the keys to press
with their meanings and allowing me to press, say, "w" instead of "@ w"
to run the corresponding action.  if it's okay
for embark-completing-read-prompter to give me access to an action via
<Prefix>+key, why wouldn't it be legit to let <Prefix> -> nil and giving
me access via key?  i'm just taking the limit; to me, it sounds
completly natural.

jaor:
just to clarify my previous answer: i'm not asking in that limit
embark-completing-read-prompter behaves exactly embark-keymap-prompter:
the key difference would be that the former is still showing the list of
available keybindings and their descriptions/annotations, but letting me
"complete" by the action key.

oantolin:
Ah, OK. Sure, I can do that. As I'm sure you know, @jaor, this function doesn't do any candidate display itself, it just calls `completing-read` and the completion UI may or may not display the candidates automatically. And if it does display them how you scroll through the list varies as well. The bindings used to display the candidates (in completion UIs that don't display them automatically) or to scroll through the candidate list may be shadowed in the keymap being prompted for. To overcome this I guess I might add another overriding keymap where you can put whatever keys your completion UI needs.

jaor:
On Thu, Mar 25 2021, Omar Antolín Camarena wrote:

> Ah, OK. Sure, I can do that. As I'm sure you know, @jaor, this
> function doesn't do any candidate display itself, it just calls
> completing-read and the completion UI may or may not display the
> candidates automatically. And if it does display them how you scroll
> through the list varies as well. The bindings used to display the
> candidates (in completion UIs that don't display them automatically)
> or to scroll through the candidate list may be shadowed in the keymap
> being prompted for. To overcome this I guess I might add another
> overriding keymap where you can put whatever keys your completion UI
> needs.

Right.  I hadn't thought of the possibility of the completion UI not
showing candidates, in which case this behaviour might be a bit awkward.
But since it is an opt-in and most people would just be using selectrum
i suppose it's not that bad :)

PR #195, opened 2021-03-28
Change `selectrum-active-p' -> `selectrum-is-active'

leungbk:
The latest rev of Selectrum does not define `selectrum-active-p'.


oantolin:
Thanks! Clemens told me about this but I forgot to change it.

minad:
Ups, sorry. I thought that this had already been fixed here. Thank you for the quick resolution!

ashton314:
Ugh; maybe the right solution is for me to regularly update my config from the README. Bleeding edge software… well… bleeds, but it sure is a lot of fun to use!

PR #196, opened 2021-03-28
Add "ext:" to declare-function invocations for external fns

leungbk:
Per the Help page for declare-function, "A FILE with an 'ext:' prefix is an external file.
check-declare will check such files if they are found, and skip
them without error if they are not."


oantolin:
I didn't know about this, thank you!

PR #199, opened 2021-04-09
Integration nitpicks

minad:
- Make the functions private. I argue that the integration code is an internal detail.
- Don't use bound-and-true-p

You may want to pull this or you may want to push your own version ;)

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -1907,51 +1907,50 @@ buffer for each type of completion."
 (defvar vertico--input)
 (defvar vertico--candidates)
 
-(defun embark-target-vertico-selection ()
+(defun embark--vertico-selected ()
   "Target the currently selected item in Vertico.
 Return the category metadatum as the type of the target."
-  (when (bound-and-true-p vertico--input)
+  (when vertico--input
     (cons (completion-metadata-get (embark--metadata) 'category)
           (vertico--candidate))))
 
-(defun embark-vertico-candidates ()
+(defun embark--vertico-candidates ()
   "Collect the current Vertico candidates.
 Return the category metadatum as the type of the candidates."
-  (when (bound-and-true-p vertico--input)
+  (when vertico--input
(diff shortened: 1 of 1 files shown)


oantolin:
So integration with "incompatible completion systems" should be private, but it's OK to keep the integration with default completion and icomplete  (which is compatible in the sense that it doesn't need separate functions) public? I guess that's sort of reasonable.

minad:
Ops, sorry. Icomplete and default should also be private.

minad:
But well, they are already in the default list somehow. I don't know :shrug:

oantolin:
But then integration with completion UIs would be private and all others target finders public, that sounds weird too. You don't want all the target finder to be private, do you?

oantolin:
I think I now agree with making the integration with incompatible completion UIs private, and keeping integration with built-in Emacs features public. I'll merge.

minad:
I object to the notion of an "incompatible completion UI" for Vertico. I claim full compatibility! :grin:

oantolin:
Well, obviously I say "incompatible" tongue in cheek. But for purposes of Embark, Vertico is "incompatible" in that the minibuffer contents and location of point are not the entire state, it has "hidden" state in the selection index.

minad:
Yes, that is right. It is also not using the `completion-all-sorted-completions` state. I had actually considered using that and performing the rotations like Icomplete is doing and just replacing the UI, but I guess this would have resulted in a more brittle and more complicated code base. And we already have Icomplete/Icomplete-vertical. For these reasons I think it was the right decision to implement it as I have it. Since the other minibuffer state does not interfere with Vertico you can use all the `minibuffer-complete` commands as usual (see https://github.com/minad/vertico#keymap).

oantolin:
I think Vertico is great as is, please don't take my "incompatible" joke seriously.

minad:
Thanks, I understood how you meant it :)

PR #200, opened 2021-04-09
Introduce embark--minibuffer-point

minad:
`(- (point) (minibuffer-prompt-end))` can be negative

I had problems with this in Marginalia (https://github.com/minad/marginalia/commit/2d1949778213ea5fa5f43be426e0c625ff419dd8) and Vertico (https://github.com/minad/vertico/commit/c72e8de81ea8898bd1646e1b8841b1dc86915e56).

----

Then I have a question about Embark metadata access

~~~ elisp
(defun embark--metadata ()
  "Return current minibuffer completion metadata."
  (completion-metadata
   (buffer-substring-no-properties (field-beginning) (point))
   minibuffer-completion-table
   minibuffer-completion-predicate))
~~~

Icomplete and Icomplete-vertical uses this:

~~~ elisp
(completion--field-metadata (icomplete--field-beg))
~~~

which is equivalent to

~~~ elisp
  (completion-metadata
   (buffer-substring-no-properties (minibuffer-prompt-end) (point))
   minibuffer-completion-table
   minibuffer-completion-predicate))
~~~

I wonder about `(point) < (minibuffer-prompt-end)`.

In Vertico I am now using this, which is probably wrong.

~~~ elisp
  (let* ((content (minibuffer-contents-no-properties))
         (metadata (completion-metadata content
                                        minibuffer-completion-table
                                        minibuffer-completion-predicate))
~~~

See also https://github.com/oantolin/icomplete-vertical/issues/27, where I have been confused for the first time.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -465,7 +465,9 @@ There are three kinds:
 (defun embark--metadata ()
   "Return current minibuffer completion metadata."
   (completion-metadata
-   (buffer-substring-no-properties (field-beginning) (point))
+   (buffer-substring-no-properties
+    (minibuffer-prompt-end)
+    (max (minibuffer-prompt-end) (point)))
    minibuffer-completion-table
    minibuffer-completion-predicate))
 
--- a/embark.el
+++ b/embark.el
@@ -318,11 +318,15 @@ window should only be used if it displays `embark--target-buffer'.")
 (defvar-local embark--command nil
   "Command that started the completion session.")
 
+(defun embark--minibuffer-point ()
+  "Return length of minibuffer contents."
+  (max 0 (- (point) (minibuffer-prompt-end))))
+
 (defun embark--default-directory ()
   "Guess a reasonable default directory for the current candidates."
   (if (and (minibufferp) minibuffer-completing-file-name)
-      (let* ((end (minibuffer-prompt-end))
-             (idx (- (point) end)))
+      (let ((end (minibuffer-prompt-end))
+            (idx (embark--minibuffer-point)))
         (expand-file-name
          (buffer-substring
           end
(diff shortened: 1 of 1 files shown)


minad:
I suspect all of the above methods to obtain the metadata are actually wrong. The correct way to obtain the metadata may be this:

~~~ elisp
(completion-metadata (buffer-substring-no-properties (minibuffer-prompt-end)
                                                     (max (minibuffer-prompt-end) (point)))
                     minibuffer-completion-table
                     minibuffer-completion-predicate)
~~~

minad:
Okay, I actually solved this in a simpler way now in Vertico. I simply do not update the UI when the point is outside the legal region, see https://github.com/minad/vertico/commit/b0f5699452b80a34dd137b79a474ea586cd56a6c.

However I think I keep the fix in Marginalia (https://github.com/minad/marginalia/commit/2d1949778213ea5fa5f43be426e0c625ff419dd8) for robustness.

And in Embark the problem is still there if you move the point to the prompt and run `embark-act`. Then you get out of range errors.

oantolin:
People should not be moving the point into the prompt! What's wrong with these people? 😆

I'm not completely convinced this really needs to be fixed, but I agree that the best fix is probably what you did.

minad:
Yes, it is very wrong :-P

1. I would have made the prompt an overlay.
2. I have this cursor-sensor-mode activated in order to prevent me from doing this

But there is a but. If you move the point right after the prompt I still had an issue in Vertico where I got -1, maybe because the sensor mode got triggered to late or because of the overlays. I have no idea. Then I thought it is better to fix this such that it works even for the people who like to move into the prompt :D

oantolin:
Merged! I was tempted instead to document that calling embark commands from within the minibuffer prompt is undefined behavior, and to add code to embark so that it uninstalls itself if you do that.

minad:
In any case it is good to merge this. Now that I made this `embark--minibuffer-point` function you have a central place where you can define the undefined and insert `(package-delete "embark")`.

PR #201, opened 2021-04-11
Adjust embark-consult to handle suffix tofus

minad:
See https://github.com/minad/consult/commit/4d0a8b9b5a7f76e7055726fb79ed73d01a49aa01. Please give it a check!

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -134,31 +134,21 @@ associated to an active minibuffer for a Consult command."
 
 ;;; Support for consult-location
 
-(defun embark-consult--strip (string property)
-  "Strip substrings with PROPERTY from STRING."
-  (if (text-property-not-all 0 (length string) property nil string)
+(defun embark-consult--strip (string)
+  "Strip substrings marked with the `consult-strip' property from STRING."
+  (if (text-property-not-all 0 (length string) 'consult-strip nil string)
       (let ((end (length string)) (pos 0) (chunks))
         (while (< pos end)
-          (let ((next (next-single-property-change pos property string end)))
-            (unless (get-text-property pos property string)
+          (let ((next (next-single-property-change pos 'consult-strip string end)))
+            (unless (get-text-property pos 'consult-strip string)
               (push (substring string pos next) chunks))
             (setq pos next)))
         (apply #'concat (nreverse chunks)))
     string))
 
-(defun embark-consult--strip-suffix (string property)
-  "Remove the final characters of STRING with the given PROPERTY."
-  (let ((len (length string)))
-    (if (and (> len 0) (get-text-property (- len 1) property string))
-        (substring string 0 (or (previous-single-property-change len property string) len))
-      string)))
-
 (defun embark-consult--location-transform (target)
   "Remove the unicode suffix character from a `consult-location' TARGET."
-  (cons 'consult-location
-        (embark-consult--strip
-         (embark-consult--strip-suffix target 'invisible)
-         'consult-location-strip)))
+  (cons 'consult-location (embark-consult--strip target)))
 
 (setf (alist-get 'consult-location embark-transformer-alist)
       #'embark-consult--location-transform)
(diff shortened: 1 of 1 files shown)


oantolin:
Looks good. I'll give it a test when I'm at the computer and then merge.

embark-consult.el
```diff
@@ -137,7 +137,9 @@ associated to an active minibuffer for a Consult command."
 (defun embark-consult--strip-suffix (string property)
   "Remove the final characters of STRING with the given PROPERTY."
   (let ((len (length string)))
-    (substring string 0 (or (previous-single-property-change len property string) len))))
+    (if (and (> len 0) (get-text-property (- len 1) property string))
+        (substring string 0 (or (previous-single-property-change len property string) len))
+      string)))
```
oantolin, line 142: Saves an unnecessary copy, I like it. This is the sort of optimization that in Vertico you'd have to weight against bumping up the line count by two... 😛

minad: I object to that! Vertico is highly optimized, this is what recently bumped the line count to 498 lines (https://github.com/minad/vertico/commit/a962a8b2b682bb99304aa022bb20d66a838fd1ea). @manuel-uberti observes the line count closely. I have to act wisely with the remaining two lines if I want to keep the users.

oantolin: Is deferring the highlighting really noticeable? When Clemens asked for the skip-highlighting variable in orderless I remember testing with Selectrum and not noticing any difference.

minad:
You can merge this as is.

But there is one small defect - `consult-global-mark` also uses `consult-location`. Before the suffix change the `consult-location` property was used to mark the part which should be stripped, in the case of `consult-global-mark` the `file:line:` location prefix. This will now break the embark occur export for `consult-global-mark`, which nobody uses.

Two alternatives:

1. Introduce a `consult-global-mark` category which will just be treated the same as the `consult-isearch` category, no occur export, only supports stripping the tofu suffix for `embark-save`.
2. Introduce another property `consult-location-strip` property which should be stripped by the `embark-consult--location-transform`.

I am inclined to go with alternative 1, what do you think?

oantolin:
A third option is to go back to using `consult-location` to mark the parts that need to be stripped, but have the stripping function check both the beginning and the end of the string for it. But I agree that export for consult-global-mark is unlikely to be used.

minad:
We wrote at the same time. See my edit above. I think option 2 is more general, so probably better in the long term. The problem with the option you propose is that I am always marking the beginning of the string with `consult-location`, even if the part should not be stripped.

oantolin:
Export to occur mode with multiple buffers already worked, you can test it with `consult-global-mark`!

minad:
Yes, I assumed that. Never tried and never used it, for obvious reasons :laughing: I will push a change for option 2 then!

oantolin:
I say we go with your option 2, for the sake of a future `consult-line-multi-buffer`, and let's keep the export working for `consult-global-mark`, why not?

minad:
See e21856cdfc88566cf873cf68c513d65b94bda4ca for the general version.

oantolin:
That does look nicer, don't you think? The only downside is that in consult you have to mark something as both invisible and `consult-strip`, but that doesn't sound too bad to me.

minad:
> That does look nicer, don't you think? The only downside is that in consult you have to mark something as both invisible and consult-strip, but that doesn't sound too bad to me.

Yes, it is better. And we can also use this in the future if we need the additional flexibility.

jaor:
On Sun, Apr 11 2021, Daniel Mendler wrote:

> @minad commented on this pull request.

> I have to act wisely with the remaining two lines if I want to keep
> the users.

you have. i don't say much, but i'm watching ;) (vertico is working like
a charm for me so far, and switching to the native compilation branch
really made it faster, for some reason).

oantolin:
I believe this is ready. Anything else you wanted to do before I merge, @minad?

minad:
@oantolin It is ready! @jaor Thanks :)

oantolin:
Merged. Thanks a lot for doing this in the first place, and for making the changes I suggested here and in Consult.

manuel-uberti:
@minad `vertico` is awesome and I'm willing to bear the tragic loss of the 500 lines limit if it will eventually happen. ;)

minad:
@manuel-uberti Thank you :D 

But staying minimal with Vertico is really good since it allows me to explore the limits of what the builtin completion infrastructure provides. There is a lot in there, and Vertico just uses it without inventing anything of its own. It performs a little bit of its own state management in contrast to Icomplete, but this is architecturally cleaner, simpler and more robust.

PR #202, opened 2021-04-13
Implement affixation-function support

minad:
I implemented affixation-function support. But I wonder if somehow the export could be improved such that the annotations are actually applied during export/snapshot. This would make fewer problems and you would also get annotations if the minibuffer is closed (See separate issue #203).

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -1452,18 +1452,22 @@ key binding for it.  Or alternatively you might want to enable
               (with-current-buffer (embark--target-buffer)
                 (let ((default-directory dir)) ; for marginalia's file annotator
                   (mapcar
-                   (lambda (cand)
-                     (let* ((annotation (or (funcall annotator cand) ""))
-                            (length (length annotation))
+                   (lambda (ann-cand)
+                     (let* ((prefix nil) (suffix "")
+                            (cand (pcase ann-cand
+                                    (`(,c ,s) (setq suffix s) c)
+                                    (`(,c ,p ,s) (setq prefix p suffix s) c)
+                                    (c c)))
+                            (length (length suffix))
                             (facesp (text-property-not-all
-                                     0 length 'face nil annotation)))
+                                     0 length 'face nil suffix)))
                        (when facesp (add-face-text-property
-                                     0 length 'default t annotation))
-                       `(,cand [(,cand type embark-collect-entry)
-                                (,annotation
+                                     0 length 'default t suffix))
+                       `(,cand [(,(if prefix (concat prefix cand) cand) type embark-collect-entry)
+                                (,suffix
                                  ,@(unless facesp
                                      '(face embark-collect-annotation)))])))
-                   candidates))))
+                   (funcall annotator candidates)))))
           (mapcar
            (lambda (cand)
              `(,cand [(,cand type embark-collect-entry)]))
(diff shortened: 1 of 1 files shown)


oantolin:
I think I'll implement this a little differently, adding an extra column for the prefix instead. I'm pretty sure this implementation breaks actions because suddenly the candidate all have a weird prefix.

minad:
I changed it to use 'line-prefix, then it continues to work. But for me the patch is mostly useless now, due to #203 and snapshot not working.

oantolin:
line-prefix almost certainly fixes the actions. This is also a valid strategy.

oantolin:
What do you mean by snapshot not working? (Sorry if I'm a little slow, it's late here and I'm a little sleepy.)

minad:
Snapshot works - I meant one does not get annotations when the minibuffer is closed and I don't get them with snapshot since the annotators run after closing.

> (Sorry if I'm a little slow, it's late here and I'm a little sleepy.)

Here it is quite early and I am still sleepy too. Otherwise the patch should have worked right away ;)

oantolin:
I think there might still be an advantage to use a third column: that way we don't have to trust the person who wrote the affixation function to do alignment. (I like for example that with a narrow frame embark collect list view aligns the annotations much more neatly than Marginalia does in Selectrum, Icomplete-vertical or Vertico.)

minad:
You are right, a third column is better. But we discussed that maybe it makes sense to stop using tabulated-list-mode? Why was that again, I don't find the discussion now.

oantolin:
Well, I wasn't suggesting abandoning tabulated-list-mode (although maybe I should have). We were talking about inserting the candidates in batches on a timer. I was thinking more of just replacing the function that inserts them, using everything else about tabulated-list-mode.

oantolin:
> I meant one does not get annotations when the minibuffer is closed and I don't get them with snapshot since the annotators run after closing.

Oh, OK. I though I did capture the annotation function before closing the minibuffer. But I guess I'd need to also capture the affixation function. As you say, the alternative would be to capture the results of the affixation/annotation function instead. I'll look into it. Ideally the affixation/annotation function would not run for export (since its results would just be discarded), but your proposal of running it right after gathering candidates for collect buffers is only wasteful for grid view, and I think I can live with that.

minad:
>  As you say, the alternative would be to capture the results of the affixation/annotation function instead. I'll look into it. Ideally the affixation/annotation function would not run for export (since its results would just be discarded), but your proposal of running it right after gathering candidates for collect buffers is only wasteful for grid view, and I think I can live with that.

Yes, and we could still optimize this and avoid annotating in that case for live grid.

oantolin:
I finally did this, and now that I did I'm not sure why. 🤣 I guess line numbers in `consult-line` are moderately useful. I do like seeing them.

In the end I decided to go with your `line-prefix` approach, because with a third column it was hard to get things perfect:

- `tabulated-list-mode` adds a space between columns, but it seems prefixes are meant to be adjacent to candidates with no space in between. I tried lying to `tabulated-list-mode`, but it didn't want to believe me when I said the column had width -1 in the case of no prefixes.
- point is put on the prefix by default, which is not ideal. I didn't want to forcibly move point to the candidate. To make actions convenient, I decided the candidate would be targeted even if you are on the prefix or suffix/annotation. But `line-prefix` does the right thing with respect to point without any fuss, so why fight it?

If I started get annoyed by misaligned prefixes I'll probably go back to 3 columns, but actually my impression is that all is was for nothing! 🤣 Besides line numbers in `consult-line`, what else uses prefixes?

oantolin:
By the way, the only reason I didn't merge your PR is that some things in the code had changed in the meantime and it seemed simpler to me to rewrite it. But basically the only thing I did differently is assume a stronger type on the affixation-function: I assumed all the elements of the return value are 3 element strings. Your code handled mixed lists of string and 2- and 3- element lists. But the stronger assumption should hold, right?

minad:
> By the way, the only reason I didn't merge your PR is that some things in the code had changed in the meantime and it seemed simpler to me to rewrite it.

No worries about my patch. I like that we got support now. Thank you very much!

> I assumed all the elements of the return value are 3 element strings. Your code handled mixed lists of string and 2- and 3- element lists. But the stronger assumption should hold, right?

Yes, the stronger assumption (always 3 elements) should hold. I think in Vertico I only handle this case.

oantolin:
What other completion tables use affixation functions with prefixes? I think so far I've only identified consult-line with its line numbers.

minad:
> What other completion tables use affixation functions with prefixes? I think so far I've only identified consult-line with its line numbers.

I think the built-in symbol affixation in help-fns.el. These are not aligned as far as I know, but please check.

oantolin:
I was going crazy trying to see those annotations until I realized you need to set `completions-detailed=t`. 😆 I think the prefixes in the symbol affixation function are always two characters (the second is a space), so they are aligned.

minad:
> I think the prefixes in the symbol affixation function are always two characters (the second is a space), so they are aligned.

I am not sure that this is still the case. There have been some changes to Emacs master recently. Did you check the current master?

oantolin:
Oh, no, I didn't check the current master. And you're right, I definitely remember discussion of using more than one character. I'll pospone the alignment issue until it actually bothers me or someone else, which, since most embark users use marginalia instead of the built-in affixations, isn't likely to happen soon.

PR #204, opened 2021-04-14
embark--display-width: Emacs 28 can compute `string-width' of substrings

minad:
See https://github.com/minad/consult/commit/48502b71e596b6532102cf2efcfc290122272aa1. This is a small optimization.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -1408,6 +1408,11 @@ key binding for it.  Or alternatively you might want to enable
 `embark-collect-direct-action-minor-mode' in
 `embark-collect-mode-hook'.")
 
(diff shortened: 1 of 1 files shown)


oantolin:
I'll take your word for the fact that some Emacsen have a 3 argument `string-width`.

minad:
I have a very recent Emacs 28 now since I started to push patches upstream. And so far I have not seen issues with the same code in Consult (But I didn't test this before, only with mockup functions) :sweat_smile:

oantolin:
I also checked that `func-arity` is in 26.1, which is the required version for Embark now (since you suggested moving up from 25, I think).

PR #208, opened 2021-04-16
Don't let-bind prefix-arg

jakanakaevangeli:
This fixes the following problem:
Press `C-u C-.` for non-quitting embark-act and then press `C-u i` to insert
the target.
(`C-u` has no effect on insertion in this case. It could have, if embark-insert
supported a prefix argument, but this isn't important for this issue.)
Now press `SPC`. 4 spaces will be inserted because the prefix-arg wasn't
cleared.


oantolin:
Thank you! I hadn't thought of the consequences of using `let` for this.

PR #219, opened 2021-05-09
Keymap grouping

minad:
Based on #218. Some WIP to try it out.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -644,7 +644,7 @@ first line of the documentation string; otherwise use the word
   "Prompt via completion for a command bound in KEYMAP.
 If NO-DEFAULT is t, no default value is passed to `completing-read'."
   (let* ((commands
-          (cl-loop for (key . cmd) in (embark--all-bindings keymap)
+          (cl-loop for (key cmd . kname) in (embark--all-bindings keymap)
                    for name = (embark--command-name cmd)
                    unless (or
                            ;; skip which-key pseudo keys and other invalid pairs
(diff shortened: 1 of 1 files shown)


embark.el
```diff
@@ -1238,19 +1238,16 @@ Returns the name of the command."
...
-                       ((null def))
-                       ((keymapp def)
-                        (dolist (bind (embark--all-bindings def))
-                          (push (cons (vconcat (vector key) (car bind))
-                                      (cdr bind))
-                                bindings)))
-                       (t (push (cons (vector key) def) bindings))))
-                    keymap)))
-      (gather (keymap-canonicalize keymap)))
+    (map-keymap
```
oantolin, line 1294: Haha! Thanks. My code is a little ridiculous. It's an editing mistake. I was considering both your approach, and a separate less consy approach that arrange for everything to be pushed onto a single list variable (here each time you run `embark--all-bindings` you get a fresh variable `bindings`, of course). Somehow I wound up with an unholy mix of both.

minad: Yes, you can pull the simplification from #218, or you rewrite everything with cl-labels. I am not a cl fan :shrug:

oantolin: Merged #218 (I hadn't noticed you also had a separate PR with just that). I am a fan of cl-lib, at least of cl-loop (which is often shorter and more readable than a sequence of mapcars, seq-filters, se-find, etc.) and, occasionally, cl-labels. :)

minad: cl-loop is in particular what I find probematic. I rather use the composable building blocks, maps, filters, etc. However there is a point in cl-loops since if one overuses the maps and filters since one ends up with very badly performing code. But then, a few times I rewrote some cl-loops in plain imperative style with while/dolist/if/let etc and the result was not significantly longer and more readable ("python style") for beginners who are not familar with cl-loop. There is also a minor syntactic argument, maybe I would be more happy with something like [iterate](https://common-lisp.net/project/iterate/).

PR #221, opened 2021-05-15
Simplify consult preview slightly

minad:
Remove embark-consult-preview--last-entry. No observable difference for me.

See https://github.com/minad/consult/blob/c839b82ce3e4db3faaecf3e9e5e2d98ad010eede/consult.el#L1004 for the Consult last candidate check.


oantolin:
I was about to do this, thanks for saving me the trouble.

PR #224, opened 2021-05-16
Deprecate embark-consult-preview in favor of consult-preview-at-point

minad:
See https://github.com/minad/consult/commit/cdc097d65e7b6d95daa187c9a7497679347c75bc. Simplifies the Embark-Consult integration a bit and seems like the cleaner solution overall. Please check if it works alright for you.


oantolin:
Works great, thanks!

PR #216, opened 2021-05-17
Export to another window

iyefrat:
Most of the time when I use `embark-export`, I don't want the exported buffer to replace the one I'm using, and would rather it open in another window. What are your thoughts on a config option that sets this? If you approve I'll send in a PR that implements this, although it seems to me that each export command needs to be edited individually for this to work, which doesn't seem ideal.

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -147,7 +147,7 @@ The elements of LINES are assumed to be values of category `consult-line'."
           (insert (concat lineno contents nl))))
       (goto-char (point-min))
       (occur-mode))
-    (switch-to-buffer buf)))
+    (pop-to-buffer buf)))
 
 (setf (alist-get 'consult-location embark-collect-initial-view-alist)
       'list)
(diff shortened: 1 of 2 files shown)


iyefrat:
Whoops, I always forget that the magit creating a pull request from an issue converts the issue instead of making a new pull request referencing it. At any rate this PR does two things:

1.  convert some tabs to spaces in `embark-consult.el`, since the whole project uses spaces
2. use `pop-to-buffer` on all export functions. Most of these were `switch-to-buffer` -> `pop-to-buffer` replacements (and some were already using `pop-to-buffer`, but `embark-export-dired` was a tad more involved.

oantolin:
Great, thanks! Even with the default `display-buffer-alist`, I like the behavior much more. This a nice improvement I didn't even know I wanted.

oantolin:
> magit creating a pull request

Wait? Magit can do that? Did you mean forge instead of magit? (I don't use forge yet, but have been tempted to learn it.)

iyefrat:
Oh right, forge. I highly recommend it! Makes dealing with github and the like much nicer.

PR #236, opened 2021-06-05
Add embark-target-bug-reference-at-point

minad:
These references occur mainly in elisp files, changelogs and Emacs C files. But I have activated the target finder only for elisp files here, since this seems the most relevant use case. At least I like to reference Emacs bugs in my packages and the Elisp standard seems to be "bug#1234".

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -143,6 +143,7 @@ Embark will set the parent of this map to `embark-general-map'.")
     embark-target-active-region
     embark-target-collect-candidate
     embark-target-completion-at-point
+    embark-target-bug-reference-at-point
     embark-target-url-at-point
     embark-target-file-at-point
     embark-target-custom-variable-at-point
@@ -460,6 +461,12 @@ There are three kinds:
                        (dired-get-filename t 'no-error-if-not-filep)
                      (ffap-file-at-point))))
     (cons 'file (abbreviate-file-name file))))
+
+(defun embark-target-bug-reference-at-point ()
+  "Target a bug reference at point."
+  (when-let ((url (seq-some (lambda (o) (overlay-get o 'bug-reference-url))
+                            (overlays-at (point)))))
+    (cons 'url url)))
 
 (defun embark-target-url-at-point ()
   "Target the URL at point."


hmelman:
This would also be useful when reading the emacs mailing lists via gnus or other reader. 😃 

I had done this in the 90s with my `mouseme.el` package which was similar to embark though enabled a context menu on a mouse click.  Clicking on our internal bug numbers in code, messages and documents was *extremely* useful.  

I don't use it myself, but those that use [debbugs.el](https://elpa.gnu.org/packages/debbugs.html) might want to use it to view the bug instead of opening an url.

It's not clear to me that emacs bug number handling should be in the default embark configuration, though the point that they are included in emacs source code is a good argument to do so.  It would perhaps be nice if there were a configuration option to recognize one or more type of bug string and associate them with an url or function to "open" them.  I had also configured such handling for RFCs and java JSRs.  To really bikeshed this I could see tying it into some expansion of thing-at-point to recognize various things.

minad:
Yes, there is also `bug-reference-mode-push-button`. I would like to have this integrated with Embark, but maybe one could do better than what I have proposed here. There exists `bug-reference-bug-regexp`. Also one has to set `bug-reference-url-format` per file.

minad:
I added a PR which reuses the `bug-reference-mode`. If there is an overlay at point with `bug-reference-url`, this url is returned.

The user should then activate `bug-reference-(prog-)mode` and set the url format locally in the buffer.

~~~ elisp
(setq-local bug-reference-url-format "https://debbugs.gnu.org/cgi/bugreport.cgi?bug=%s")
~~~

This is very nice since it potentially works everywhere :)

hmelman:
I was unaware of `bug-reference-mode`, I agree using it here makes sense.

oantolin:
I like the idea of using `bug-reference-mode` for this. It seems pretty easy to configure and seems to have lots of sensible defaults already. It has functions to setup the correct URL format from context if you are in a VC repo with upstream on Savannah, GitHub or Gitlab, in Gnus reading a mailing list, or chatting on rcirc or ERC.

oantolin:
One thing I wonder though is if it might be nicer to use `bug-reference-bug-regexp` and `bug-reference-url-format` directly and not require the mode to be on.

minad:
Yes, I also thought that turning the mode on is nicer since you also get highlighting. Downside is that this is a little less flexible. On the upside, we don't have to require bug-reference.

PR #238, opened 2021-06-06
Bind apropos to symbol/identifier maps, deprecate embark-save

minad:


oantolin:
When I added non-command actions I forgot to check if any of the existing actions were there solely to wrap an existing function. I should have removed `embark-save` then.

Issue #237, opened 2021-06-05
Document better keybindings in the README


minad:
Just some idea, but I kind of love this now. I have bound `embark-default-action` to `M-.` and `embark-act` to `C-.`. This makes a lot of sense since `xref-find-definitions` is bound to `M-.` and Embark generalizes this. It may also make sense to rename `embark-default-action` to `embark-act-default` for the symmetry.

hmelman:
Very interesting.  I think the command `embark-default-action` and how the default action is defined (is it by binding it to RET?) should be mentioned in the README 😄

oantolin:
@minad  These do seem like sensible key bindings. Well, for graphical Emacs: I don't think `C-.` works on many terminals (`M-.` should be fine in terminals). I guess I already have an example of some keybinding, I might as well have it be a nicer one. (Maybe what I'd really like is to not have any key binding, but people seem to like the ability to copy and paste a configuration.) 

On a personal note, I'm wondering if I should bind `embark-default-action` now, I do seem to use the default action a lot more these days...

@hmelman Good point! I will do that. I tend to regard the docstrings as the primary documentation and forget the manual a bit. 🙄

oantolin:
Oh, and to answer your question @hmelman, as the `embark-default-action` docstring says, the default action is determined as follows:

First the `embark-default-action-overrides` alist associating target types to default actions is consulted. Failing that, if you are in the minibuffer, the default action is the command you ran that opened the minibuffer; if you aren't in the minibuffer, the RET binding in the relevant action keymap is used (I call the RET binding the "default default action" in my head).

The idea of the overrides alist is that the current system says that, for example, if you run `M-x delete-file` the default action is `delete-file`, which makes sense to me, but someone might prefer that `find-file` always be the default action for files, no matter what command is prompting for a file.

The default configuration specifies no overrides and embark-consult uses one for an interesting reason: `consult-grep` candidates have the funny property that they are not valid input for `consult-grep` usually, and if they are they might not match themselves! I guess you could say `consult-grep` is really presenting a menu of options, not _completing_ candidates. So embark-consult sets an overriding default action for the `consult-grep` type that goes to the indicated location. Notice that this simulates what the minibuffer rule usually accomplishes: that using the default action is like picking that completion candidate and pressing RET.

minad:
You are right about the terminal. But you could add a comment that on terminal one might want to use sth different.

[...]

PR #241, opened 2021-06-13
Rename embark-default-action to embark-dwim

minad:
As discussed in #237

### DIFF
--- a/README.org
+++ b/README.org
@@ -152,8 +152,8 @@ make the =switch-to-buffer= command become =find-file= for this run.
 You can bind =embark-become= to a key in =minibuffer-local-map=, but it is
 also available as an action under the letter =B= (uppercase), so you
 don't need a binding if you already have one for =embark-act=. So,
-assuming I have =embark-act= bound to, say, =C-S-a=, once I realize I
-haven't open the file I can type =C-S-a B C-x C-f= to have
+assuming I have =embark-act= bound to, say, =C-.=, once I realize I
+haven't open the file I can type =C-. B C-x C-f= to have
 =switch-to-buffer= become =find-file= without losing what I have already
 typed in the minibuffer.
 
@@ -169,7 +169,7 @@ I accidentally try to switch to a buffer for a file I haven't opened
 yet, =embark-become= finds that the command I ran, =switch-to-buffer=, is
 in the keymap =embark-become-file+buffer-map=, so it activates that
 keymap (and any others that also contain a binding for
-=switch-to-buffer=). The end result is that I can type =C-S-a B f= to
+=switch-to-buffer=). The end result is that I can type =C-. B f= to
 switch to =find-file=.
 
 * Configuration
@@ -191,7 +191,8 @@ reasonable starting configuration:
     :ensure t
 
     :bind
-    (("C-S-a" . embark-act)       ;; pick some comfortable binding
+    (("C-." . embark-act)         ;; pick some comfortable binding
+     ("C-;" . embark-dwim)        ;; good alternative: M-.
      ("C-h B" . embark-bindings)) ;; alternative for `describe-bindings'
 
     :init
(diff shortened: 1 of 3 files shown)


oantolin:
I hadn't remembered until just now, but `C-.` is a suboptimal binding for `embark-act` because icomplete uses it to cycle among completions! (It's bound to `icomplete-forward-completions` in `icomplete-minibuffer-map`. If we really want to recommend that binding, we'll need a note for icomplete users.

oantolin:
I think we should probably just recommend a different binding. I don't want to recommend using what is probably an important default key bindings for icomplete users. (Shadowing `M-.` is very different, since at least `embark-dwim` is a generalization of the default binding.)

minad:
Oh okay, I wasn't aware of that icomplete binding. I always configure Icomplete to use the movement keys for rotation, similar to vertico.

oantolin:
Me too, which is why I forgot that keybinding until now.

minad:
One more nitpick - M-. is not an exact generalization, since the xref command works also if there is no symbol at point. Therefore one may still keep the xref commandbound somewhere. For me C/M-. work well at least.

minad:
@oantolin I added another simplification commit, is there a reason why embark--target returns `(nil nil nil)` when no target is found. This is not exactly intuitive ;)

oantolin:
You are mixing a lot of unrelated changes here. So it'll take me a while to evaluate the PR as a whole. I think the embark-dwim changes are ready to merge as is. The key binding docs I'll need to change to something that doesn't clash with icomplete (since I don't want bug reports claiming that `embark-act` does nothing but cycle the completion candidates :P). And you last change I'll have to think about. Most places where `embark--target` is called destructure the value. I don't know if `nil` will raise an error or not, maybe we need `(nil nil)` --we certainly don't need `(nil nil nil)`.

minad:
No, the destructuring does not raise an error.

~~~ elisp
(pcase-let ((`(,x . ,y) nil))
  nil)
~~~

But I am bailing out early with `(or ... (user-error ...))`. Of course one can also rewrite this as a simple `pcase`. But I would really like to see `embark--target` returning plain nil if no target is found.

oantolin:
I'm slow. I'll have to think a bit about whether it is OK for `embark--target` to return `nil`. It isn't designed too, but maybe it can return `nil` with no complications.

minad:
> I'll have to think a bit about whether it is OK for embark--target to return nil.

I looked at all the call sites, there are only three or. They are all okay with `embark--target` returning nil. But please give it another cross check.

Sorry for mixing the unrelated changes. I just stumbled upon this when I wrote this `dm-embark-dwim` command.

~~~ elisp
  (defun dm-embark-dwim (&optional arg)
    (interactive "P")
    (if (or arg (not (embark--target))) ;; NOTE: Check here if there is an Embark target!
        (call-interactively #'xref-find-definitions)
      (embark-default-action)))
~~~

oantolin:
Sorry, I think I just panicked! 😄 It is indeed easy to check that `embark--target` is allowed to return `nil`.

oantolin:
I have to say though, that `(pcase-let ((`(,a ,b ,c) nil)) (list a b c))` returning `(nil nil nil)` feels like a bug in `pcase-let` and I'm worried that Emacs might fix it at some point. 😛

minad:
> Sorry, I think I just panicked!

Usually I try to avoid such unrelated changes. But in some case when looking at something it occurs that one finds some small inconsistency or weirdness as in this case which can be improved quickly - like some on the fly code review. I have no intention of doing a huge overhaul here :)

> I have to say though, that (pcase-let (((,a ,b ,c) nil)) (list a b c))returning(nil nil nil)feels like a bug inpcase-let` and I'm worried that Emacs might fix it at some point.

Agree. But note that I am not using this here, since I am bailing out early with the user-error. Should be all good in any case. I was also wondering about the pcase desugaring and checked what it does.

oantolin:
> Agree. But note that I am not using the nil pcase-let here, since I am bailing out early with the user-error. Should be all good in any case.

Yes, of course. I wrote that before realizing you also changed `embark-act`. I like the `(or (thing) (user-error "boom"))` idiom better that what was there before. It would be even better with `(defalias 'die 'user-error)`, like Perl.

PR #243, opened 2021-06-14
embark-region-map: Add align action

minad:


oantolin:
I usually find `align-regexp` more useful than `align`. Maybe we should add both?

oantolin:
Although, it could just be ignorance, I've never bothered really configuring align. Maybe if I did I'd be able to use it more often.

minad:
`align` comes already preconfigured with a set of regexps. I think it is good to have here, also for discoverability. Maybe people will use it more often then ;)

oantolin:
Well, let's definitely add `align`. But maybe also `align-regexp` is worth adding.

oantolin:
Align comes with a ton of rules! I wonder if any of them does something I'd be interested in...

oantolin:
Oh, looking at the value in a customize buffer makes it much easier to understand.

Issue #244, opened 2021-06-14
Repeating actions


minad:
This is a good idea from cmap - it allows repeating actions by pressing the action key multiple times. This is similar to Emacs 28 `repeat-mode`.

For example cmap has commands `cmap-symbol-next/previous`. Furthermore it is potentially useful for org commands at point which can be repeated, e.g., `org-timestamp-up`.

~~~ elisp
(defun embark-symbol-next (sym)
  (interactive "s")
  (forward-char)
  (re-search-forward (format "\\_<%s\\_>" (regexp-quote sym))))

(defun embark-symbol-previous (sym)
  (interactive "s")
  (backward-char)
  (re-search-backward (format "\\_<%s\\_>" (regexp-quote sym))))

(define-key embark-symbol-map "p" #'embark-symbol-previous)
(define-key embark-symbol-map "n" #'embark-symbol-next)
~~~

These commands could be repeated by pressing `C-. n n n`.

I assume that support for this will be easy to implement when Embark uses transient keymaps, since then one could just keep the action keymap alive.

oantolin:
Aren't there very few actions one would want to repeat? But yes, going back to transient maps would probably make this easy.

minad:
Yes, I am not entirely sure about this. There are probably not many actions, you can take a look at cmap for other examples. Nevertheless I wanted to put it up for discussion.

hmelman:
I'm a big fan of the package `symbol-overlay`.  It's basic feature is that it lets you highlight all occurrences of `symbol-at-point` in a unique color which makes scanning code for a variable easier.  But it's bonus feature is that it installs a transient map on the symbol giving some basic commands easy key bindings: next, previous, first , last, copy (kill-new) and even replace and query-replace (which are quite nice).

If embark could generalize these for any target it would be great.  I think you'd want different maps for different target types.  symbols or variables you'd want to treat individually like symbol-overlay (go to the next occurrence of this symbol) but for URLs or bug-references (or I guess file names in a buffer) I think you'd want to just go to the next URL.  I'm not much of an org user but I'd think various org targets would behave as URLs (eg go to the next timestamp).  URLs wouldn't want a replace command but browse-url or open in eww would be nice.  So: `C-. n n n n b` would open the 4th URL from point in the browser.

oantolin:
[...]

PR #245, opened 2021-06-14
Add repeatable actions

minad:
~Prototype for #244. The search commands are not very useful because of isearch-forward-symbol-at-point, but they are at least a good demonstrator.~

Implementation of #244.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -211,7 +211,8 @@ is the key representation accepted by `define-key'."
 (defcustom embark-keybinding-repeat
   (propertize "*" 'face 'embark-keybinding-repeat)
   "Indicator string for repeatable keybindings.
-Keybindings are formatted by the `completing-read' prompter and the verbose indicator."
+Keybindings are formatted by the `completing-read' prompter and
+the verbose indicator."
   :type 'string)
 
 (defface embark-keybinding-repeat
--- a/embark.el
+++ b/embark.el
@@ -1176,6 +1176,10 @@ after this delay shows the verbose indicator."
   :type '(choice (const :tag "No delay" 0)
                  (number :tag "Delay in seconds")))
 
+(defcustom embark-mixed-indicator-both t
+  "Show both indicators, even after the verbose indicator appeared."
+  :type 'boolean)
+
 (defun embark-mixed-indicator ()
   "Mixed indicator showing keymap and targets.
 The indicator shows the `embark-minimal-indicator' by default.
@@ -1186,6 +1190,7 @@ helpful keybinding reminder still pops up automatically without
 further user intervention."
   (let ((vindicator (embark-verbose-indicator))
         (mindicator (embark-minimal-indicator))
+        vindicator-active
         vtimer)
     (lambda (&optional keymap targets prefix)
       ;; Always cancel the timer.
@@ -1198,18 +1203,20 @@ further user intervention."
       (if (not keymap)
           (progn
             (funcall vindicator)
-            (when (functionp mindicator)
+            (when mindicator
               (funcall mindicator)))
-        (if (null mindicator)
+        (when mindicator
+          (funcall mindicator keymap targets prefix))
+        (if vindicator-active
             (funcall vindicator keymap targets prefix)
-          (funcall mindicator keymap targets prefix)
           (setq vtimer
                 (run-at-time
                  embark-mixed-indicator-delay nil
                  (lambda ()
-                   (when mindicator
+                   (when (and (not embark-mixed-indicator-both) mindicator)
                      (funcall mindicator)
                      (setq mindicator nil))
+                   (setq vindicator-active t)
                    (funcall vindicator keymap targets prefix)))))))))
 
 ;;;###autoload
--- a/embark.el
+++ b/embark.el
@@ -820,11 +820,33 @@ first line of the documentation string; otherwise use the word
     ((stringp (car-safe cmd)) (car cmd))
     ((symbolp cmd) (symbol-name cmd))
     ((keymapp cmd) "<keymap>")
-    ((when-let (doc (and (functionp cmd) (documentation cmd)))
+    ((when-let (doc (and (functionp cmd) (ignore-errors (documentation cmd))))
        (save-match-data
          (when (string-match "^\\(.*\\)$" doc)
            (match-string 1 doc)))))
     (t "<unnamed>"))))
+
+;; Taken from Marginalia, needed by the verbose indicator.
+;; We cannot use the completion annotators in this case.
+(defconst embark--advice-regexp
+  (rx bos
+      (1+ (seq (? "This function has ")
+               (or ":before" ":after" ":around" ":override"
+                   ":before-while" ":before-until" ":after-while"
+                   ":after-until" ":filter-args" ":filter-return")
+               " advice: " (0+ nonl) "\n"))
+      "\n")
+  "Regexp to match lines about advice in function documentation strings.")
+
+;; Taken from Marginalia, needed by the verbose indicator.
+;; We cannot use the completion annotators in this case.
+(defun embark--function-doc (sym)
+  "Documentation string of function SYM."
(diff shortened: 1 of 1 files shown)


minad:
Given the new indicator and target cycling this feature becomes more important. Ideally it allows to repeat actions within the same target type without reopening the delayed indicator and without requiring cycling to the target again. There are these TODOs in the code:

                 ;; TODO temporarily remove all target finders except the
                 ;; current one We have to rerun the target finder, but we want
                 ;; to stay within the same target type for repeated commands.
                 ;; TODO Ensure that the mixed indicator stays open if it is
                 ;; already open.

What do you think about this feature?

oantolin:
This would definitely be nice to have. Even if the only repeatable actions were to scroll the verbose indicator it would be worth it. 😛

minad:
Hmm, scrolling is not really an action. I am not sure if it should use the same mechanism, probably not.

minad:
But if you like this, I will implement this cleanly such that it is ready for merge.

oantolin:
You're probably right about scrolling. That should probably be directly supported in the keymap prompter.

To be honest, I'm a little conflicted about repeatable actions. I mean, it sounds nice to have them in the abstract, but I' not sure it's worth it. There seem to be very few actions one would want to repeat. On the other hand, maybe if we had them more uses would crop up over time and it doesn't really seem like that much code to support them...

How stongly or weakly do you feel about repeatable actions?

oantolin:
I guess the case for repeatable actions did get significantly stronger with multiple targets. Before we could just tell people to run embark-act again and it was not too onerous, but now it entails run embark-act again _plus_ cycling to the correct target again...

minad:
Yes indeed. I would like to have them. Also since we got multiple targets at point, this basically opened the gates to add many more targets, which would otherwise be in conflict. Before that I could not see me using Embark for more things.

Of course I would only add this if we can do it with reasonable adjustments and without changing too much. But I believe this should be doable, given the simplicity of the prototype here.

oantolin:
Let's do it then, since you so graciously offered to complete the prototype; and since, as you point, the prototype shows this is not too complicated.

minad:
I implemented proper repeating. Thanks to the new indicator protocol which keeps the indicator alive this is a pretty simple addition. Please take a look!

embark.el
```diff
@@ -2851,7 +2922,9 @@ and leaves the point to the left of it."
   ("r" xref-find-references)
   ("b" where-is)
   ("e" pp-eval-expression)
-  ("a" apropos))
+  ("a" apropos)
+  ("n" search-forward)
```
oantolin, line 2926: Nice! I like it, even though it doesn't search with symbol boundaries. I wonder if that'll be a problem.

embark.el
```diff
@@ -1447,45 +1507,52 @@ done by cycling backwards) and cycling starts from the following
...
+                             target)
+                           bounds
+                           (if embark-quit-after-action (not arg) arg))
+              (when-let (new-targets (and (memq action embark-repeat-commands)
+                                          (embark--targets)))
+                ;; Terminate repeated prompter on default action, when repeating.
+                ;; Jump to the same target type.
+                (setq default-done #'embark-done
+                      targets (embark--rotate new-targets
+                                              (or (cl-position-if (lambda (x) (eq (caar x) (caaar targets)))
```
oantolin, line 1554: You don't often see caaar!

embark.el
minad: Yes, I figured that we prefer to use the builtins - also as a demonstration how Embark can power up built-in commands. In an older version I had the extra embark-symbol-next/previous commands.

embark.el
minad: Isn't it lovely ;)

embark.el
```diff
@@ -1148,6 +1196,10 @@ after this delay shows the verbose indicator."
   :type '(choice (const :tag "No delay" 0)
                  (number :tag "Delay in seconds")))
 
+(defcustom embark-mixed-indicator-both t
```
oantolin, line 1199: Oh, you prefer the repetition Prot complained about?

minad: Yes, indeed I do. I also mentioned it to @protesilaos in the discussion of #310, see https://github.com/oantolin/embark/issues/310#issuecomment-889732884.

minad: But making it the default setting is of course debatable. I usually think it is easier for users to turn something off than on, since how would they know. It seems to me that these configurations work well as default:

* reuse-window and indicator-both=t
* resizable at bottom and indicator-both=nil

embark.el
oantolin: Right, and the prototype did add the symbol boundaries to the regexp. I like this, though, let's keep it and see if the symbol boundaries get requested.

embark.el
oantolin: OK. Would you consider making the default nil? I only thought removing the minimal indicator looked "jumpy" when the minimal indicator came back while cycling. Now I think the single jump when switching from minimal to verbose is smooth.

oantolin: Even with reuse-window I prefer nil: the sudden appearance of the verbose indicator naturally takes your eyes away from the echo area, so you don't even notice the minimal indicator disappearing. Conversely, if you do keep your eyes on the echo area, the disappearance reminds you the verbose indicator has now appeared elsewhere.

oantolin: Well, this isn't important. We can discuss it once merged.

minad: > Would you consider making the default nil?

Ok, done.

> Even with reuse-window I prefer nil: the sudden appearance of the verbose indicator naturally takes your eyes away from the echo area, so you don't even notice the minimal indicator disappearing. Conversely, if you do keep your eyes on the echo area, the disappearance reminds you the verbose indicator has now appeared elsewhere.

For me it is like this - if I keep my eyes focused on the minibuffer or on the element I act on, I notice the vindicator appearing but I don't really look at it, probably in particular when repeating or cycling. The vindicator is still helpful in order to have the help readily available. But the disappearance of the Act symbol is disturbing, e.g., in the minibuffer without quit it looks as if I already acted.

embark.el
minad: We can "maintain" the version with the boundaries in the wiki ;)

We will see - I am not yet sure how I am going to use all this repeat niceness. The n/p actions on symbols almost replace symbol-overlay. If we add more specialized next/previous actions (e.g. for urls or other targets), we may reconsider adding the proper symbol actions too. For which targets at point do we want such actions? @hmelman mentioned acting on the next target in #244.

oantolin:
Merged. Thanks again, @minad! We should think about whether any other actions should be marked repeatable. I'll bring that up at #244 because I think issues can be a more comfortable place for people to chime in than PRs.

minad:
Thanks for merging! Embark is getting really powerful. I like in particular how Embark can do more now than a usual UI right click context menu with the cycling and repeating :)

oantolin:
Yes! Cycling was a really great idea, @minad. I kept putting off the multiple targets things because I knew in my heart that I wouldn't really like the UI I was planning (which involved first prompting you for which target you meant). And when you implemented cycling the ball started rolling again.

For me another big thing was this `command-loop` parameter of read-key-sequence, that was a great find too, since I really didn't want to invert control again, because that would touch code all over the place.

minad:
We could still invert control. Now that I understand how this works it wouldn't affect so many places. But now that `read-key-sequence` also uses the command loop I don't see the point in control inversion. But maybe there are edge cases.

oantolin:
I don't think we have any motivation for inverting control again.

minad:
Currently not, I fully agree. Therefore I also didn't bug you about it anymore. But then, packages like transient hydra etc all use inverted control, so maybe there is something to it. I really don't know if there are certain cases where inverted control is better.

In the wiki you describe some hack regarding key capturing, I think this should also be unnecessary now with read-key-sequence+control-loop.

PR #248, opened 2021-06-15
Add target finders for defuns and expressions

minad:
I am using the usual Embark approach for the expressions now - injecting the expression string.

EDIT: Due to the defun target finder we lose the symbol actions of `(defun`.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -120,7 +120,9 @@
     (environment-variables . embark-file-map) ; they come up in file completion
     (url . embark-url-map)
     (buffer . embark-buffer-map)
+    (expression . embark-expression-map)
     (identifier . embark-identifier-map)
+    (defun . embark-defun-map)
     (symbol . embark-symbol-map)
     (command . embark-command-map)
     (variable . embark-variable-map)
@@ -143,6 +145,8 @@ For any type not listed here, `embark-act' will use
     embark-target-bug-reference-at-point
     embark-target-url-at-point
     embark-target-file-at-point
+    embark-target-defun-at-point
+    embark-target-expression-at-point
     embark-target-custom-variable-at-point
     embark-target-identifier-at-point)
   "List of functions to determine the target in current context.
@@ -234,7 +238,7 @@ It can be overriden by the `embark-setup-overrides' alist."
 (defcustom embark-setup-overrides
   '((async-shell-command embark--shell-prep)
     (shell-command embark--shell-prep)
-    (eval-expression embark--eval-prep)
+    (pp-eval-expression embark--eval-prep)
     (package-delete minibuffer-force-complete))
   "Alist associating commands with post-injection setup hooks.
 For commands appearing as keys in this alist, run the
@@ -284,7 +288,7 @@ When this variable is nil, it is overridden by
     shell-command
     async-shell-command
     embark-kill-buffer-and-window
-    eval-expression)
+    pp-eval-expression)
   "Allowing editing of target prior to acting for these commands.
 This list is used only when `embark-allow-edit-default' is nil."
   :type 'hook)
@@ -494,6 +498,26 @@ This function mostly relies on `ffap-file-at-point', with two exceptions:
     (when-let ((symbol (get-text-property (point) 'custom-data)))
       (cons 'variable (symbol-name symbol)))))
 
+(defun embark-target-expression-at-point ()
+  "Target expression at point."
+  (when-let (exp (and
+                  (or (derived-mode-p 'emacs-lisp-mode)
+                      (not (derived-mode-p 'prog-mode)))
+                  (or (memq (char-syntax (char-after)) '(?\( ?\"))
+                      (memq (char-syntax (char-before)) '(?\) ?\")))
+                  (thing-at-point 'sexp)))
+    (cons 'expression exp)))
+
+(defun embark-target-defun-at-point ()
+  "Target defun at point."
+  (when-let (bounds (bounds-of-thing-at-point 'defun))
+    (let ((str (buffer-substring (car bounds) (cdr bounds))))
+      (when (and
+             (string-match "\\`(\\(?:\\w\\|\\s_\\)+" str)
+             (or (>= (point) (1- (cdr bounds)))
+                 (<= (point) (+ (car bounds) (match-end 0)))))
+        (cons 'defun str)))))
+
 (defun embark-target-identifier-at-point ()
   "Target identifier at point.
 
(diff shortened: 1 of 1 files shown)


minad:
@oantolin Which concrete sexp-at-point actions would you like to see here? We could normalize the commands by calling `(bounds-of-thing-at-point 'sexp)` and then moving the point either to the beginning or the end depending on what the command requires.

minad:
@oantolin Did you have another chance to look at this PR?

minad:
@oantolin Is there interest in this?

oantolin:
I am interested in the functionality. I haven't had a chance to look at it, but I remember thinking the wrappers you mentioned sounded more complicated than necessary and that the sexp functions should probably use the region trick, not return the sexp as a string (which I seem to recall is what you were using). I might be able to look at it soon, but I'm still swamped with other stuff.

minad:
> I remember thinking the wrappers you mentioned sounded more complicated than necessary and that the sexp functions should probably use the region trick, not return the sexp as a string (which I seem to recall is what you were using).

I looked into this, and no, it won't be simpler. There is some value in having the objects available as strings (to make use of the general map and other functions which consume strings). Furthermore there are not many more sexp at point functions from which we would benefit if we go with the region trick. 

The only thing which I don't like about this PR here is that we have to write these wrappers. But even with the region trick we would need wrappers to normalize the point position (to before or after the sexp), depending on what we decide. Given that we need wrappers in any case, the string consuming wrappers are the better solution.

minad:
Rebased on master

oantolin:
So out of 13 actions for sexp and defun targets only 2 seem to want a string injected: pp-eval-expression and pp-macroexpand-expression. And both of those have variants that use the last sexp. It really seems like strings are the wrong way to go.

oantolin:
Or maybe I should embrace useless strings and also have them for regions? There they would actually cause problems for a couple of actions, if I recall correctly, get injected at the wrong prompt.

minad:
> So out of 13 actions for sexp and defun targets only 2 seem to want a string injected: pp-eval-expression and pp-macroexpand-expression. And both of those have variants that use the last sexp. It really seems like strings are the wrong way to go.

Well, as I argued you need a wrapper in any case since some commands work at the beginning and some work at the end. And by reusing strings we also get the string actions for free. Why wouldn't you want that?

But I am not fixated on this approach, if the other approach works better, then we should of course use the other approach. Maybe implement the other variant with the region approach and then we can compare side by side. Which approach leads to less code/less complexity, which approach offers more actions and so on. You know, I am always happy if Embark reuses commands without wrappers just as is. But it seems this won't be the case here :(

oantolin:
Getting the general actions for free does sound good, which is why I wondered if maybe the region target should be change to return a string (probably more trouble than its worth in that case since some region commands do prompt for a file or buffer name).

But then again, which general commands would one really use on a sexp or defun? I think maybe only saving to the kill ring.

minad:
I always found it a bit odd that regions are treated specially. But if it leads to trouble when this is changed to strings, I wouldn't do it.

oantolin:
I'm sorry for taking so long. I took a terrible approach! I should have merged first and thought about possible implementation changes later.

oantolin:
If I can get rid of the few wrappers by using a setup overide hook, I will, but that can always be done later.

minad:
> I'm sorry for taking so long. I took a terrible approach! I should have merged first and thought about possible implementation changes later.

No worries, actually I prefer if you only merge functionality with which you fully agree, or at least 90% ;)

> If I can get rid of the few wrappers by using a setup overide hook, I will, but that can always be done later.

I am not sure it can be done that way. The setup hooks run only after injection in the minibuffer, but for these sexp commands the minibuffer is not used. One could fiddle with the embark-pre/post-command-hook but this does not seem like a cleaner approach.

oantolin:
> I am not sure it can be done that way. The setup hooks run only after injection in the minibuffer, but for these sexp commands the minibuffer is not used. One could fiddle with the embark-pre/post-command-hook but this does not seem like a cleaner approach.

Right! I got confused, I meant the pre-action hook. Which unfortunately is "global", not an alist keyed by action. Maybe it should be an alist with a `t` key for global behavior...

oantolin:
Mmh. Should I maybe also bind `eval-defun` and `pp-eval-expression` to `e`? They are currently only bound to `RET`. In all other Embark keymaps, the command bound to `RET` is also available in another way, in case the default action is different. Now, these new keymaps are for targets that never come from the minibuffer, so the `RET` binding would never be overridden, but even so it feels more consistent to have all actions accessible from non-RET bindings, and have RET duplicate the one that we want `embark-dwim` to carry out.

minad:
> Right! I got confused, I meant the pre-action hook. Which unfortunately is "global", not an alist keyed by action. Maybe it should be an alist with a t key for global behavior...

Yes, something like this would be possible. However you actually need a wrapper hook because we want to save the excursion. I am not sure if I prefer such hooks or if I prefer to write wrapper commands manually. The wrapper commands will at least make the code flow more clear, one could also think about getting rid of the setup hooks and doing everything via wrappers.

I am not sure if you like such patterns, but we could use macros for the generic wrappers like this:

~~~ elisp
(defun embark--beginning-of-sexp ()
  "Go to the beginning of the Sexp at point."
  (goto-char (car (bounds-of-thing-at-point 'sexp))))

(defmacro embark--sexp-command (cmd)
  "Derive sexp command acting on sexp at point from CMD."
  `(defun ,(intern (format "embark-%s" cmd)) ()
     ,(format "Run `%s' on the sexp at point." cmd)
     (interactive)
     (save-excursion
       (embark--beginning-of-sexp)
       (,cmd))))

(embark--sexp-command indent-sexp)
(embark--sexp-command kill-sexp)
(embark--sexp-command raise-sexp)
~~~

> Mmh. Should I maybe also bind eval-defun and pp-eval-expression to e? They are currently only bound to RET. In all other Embark keymaps, the command bound to RET is also available in another way, in case the default action is different. 

You are right, I missed this!

> Now, these new keymaps are for targets that never come from the minibuffer, so the RET binding would never be overridden, but even so it feels more consistent to have all actions accessible from non-RET bindings, and have RET duplicate the one that we want embark-dwim to carry out.

Indeed.

oantolin:
> However you actually need a wrapper hook because we want to save the excursion.

I had missed this. I think you're probably right that a wrapper is the best way to go.

> I am not sure if you like such patterns, but we can use macros for the generic wrappers

I do like using macros for precisely this type of thing. I'll probably make the change.

> Indeed, the e should be bound for consistency.

Will do.

oantolin:
Argh, pp-eval-expression has a kind of a low threshold for popping up a buffer. 😆 Maybe that's configurable.

oantolin:
I quite like the ellipsis for multiline targets, thanks for that too!

minad:
> I quite like the ellipsis for multiline targets, thanks for that too!

Thanks :) When I added this, I wondered - didn't Embark have multiline targets before or they were uncommon? For the defun/expression targets there is no way around the truncation, at least as long as we use string targets.

oantolin:
The main (only?) source of multiline targets before were region contents.

oantolin:
Now I'm tempted to rework the region target so it is the region contents as a string. I think it cause trouble with very few of the actions, and maybe it's better to just have a wrapper for those.

minad:
I think it is generally a nice paradigm if most embark targets are strings. Most objects have a string representation and this gives us maximal reusability of commands acting on strings. For the few scenarios were non-serializable objects are acted on we may keep the non-string targets, but are there any besides region?

Issue #258, opened 2021-07-12
Add a possibility to specify the parent keymap of an embark-keymap (general/meta)


minad:
The parent keymap is currently hardcoded to either the meta or the general map, depending on if the Embark target is region. Wouldn't it make sense to offer the possibility to specify the parent keymap per Embark keymap or offer some override?

~~~ elisp
(defvar embark-parent-map
   '((region . embark-meta-map)
     (something . nil) ;; disable the parent
     (otherthing . other-map) ;; use another map
     (t . embark-general-map))) ;; the default
~~~

(Why are you not simply using `:parent`? I think you told me before, or I asked this before, but I forgot the reason for the current approach.)

bdarcus:
Since I have a screenshot that suggests the impulse that prompted this.

I am experimenting here with menu descriptions, to bundle contextual actions from two packages together with common menu entries.. 

![Screenshot from 2021-07-12 15-10-12](https://user-images.githubusercontent.com/1134/125342989-8b939e80-e323-11eb-987d-30a374d92aec.png)

I'd actually like to also remove the embark map here (this is for actions at point, so they're not useful here),  and the other menu items, so it's just a very clean menu, focused only on the contextual commands.

oantolin:
I think I'd be happy to go back to the original approach: explicitly setting the parent of the keymap (which now would mean using `:parent`). I think it was @clemera that suggested to always automatically inherit from the general map (or the meta map, as the case may be), as a convenience for users who then wouldn't need to remember to inherit from the general map.

Requiring explicit inheritance feels less magic, I think I'd be happy with that even if it is a slightly breaking change.

minad:
You could still default to general map in the embark-define-keymap macro. This won't be less convenient.

oantolin:
Yes, that would leave the common case the same. Just to be through let me see if I can find where this automatic inheritance was discussed.

oantolin:
Mmh, I couldn't quickly find the discussion but I'm pretty sure it was just meant as a convenience so you don't have to remember to inherit from the general map. The commit message for the change just says:

> Make embark-general-map actions always available
    
> Previously this happened by convention only: the action keymaps that
> come with embark all had embark-general-map as their parent. Now this
> is enforced by embark--setup-action. This also means the key t in
> embark-keymap-alist is no longer necessary.

oantolin:
See also 7d11ba626d5edeb0e3427086f567e6cf18e63508 which fixes an embarrassing bug in the previous commit.

[...]

PR #260, opened 2021-07-17
Further simplification of embark--action-keymap

minad:
* Get rid of the overriding-keymap, why is it needed? In order to
  override one can also advise embark--action-keymap.
* Do not use embark-general-map by default, generalize.

(Further simplifications regarding to #258)

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -128,15 +128,12 @@
     (unicode-name . embark-unicode-name-map)
     (package . embark-package-map)
     (bookmark . embark-bookmark-map)
-    (region . embark-region-map))
+    (region . embark-region-map)
+    (t . embark-general-map))
   "Alist of action types and corresponding keymaps.
 For any type not listed here, `embark-act' will use
 `embark-general-map'."
   :type '(alist :key-type symbol :value-type variable))
-
-(defvar embark-overriding-keymap nil
-  "Can be bound to short circuit `embark-keymap-alist'.
-Embark will set the parent of this map to `embark-general-map'.")
 
 (defcustom embark-target-finders
   '(embark-target-top-minibuffer-completion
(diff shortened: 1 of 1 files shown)


minad:
Note that the export function override is redundant. Is there a point in keeping this variable? Embark already has many customization variables and here we can cut it a bit down without downsides.

minad:
Hmm, it seems @clemera introduced both these variables in https://github.com/oantolin/embark/pull/37. I am essentially reverting this PR here. @clemera argued that the overrides simplify the overriding and the amount of work the user of this functionality has to do. Maybe this should be reconsidered? From my point of view it is better to have an easier lookup logic with fewer special casing and fewer customization variables the user has to lookup and understand. What do you think?

oantolin:
> Also in #37 the parent keymap logic has been introduced which turned out to be not flexible enough.

Oh, I didn't realize this. I thought the other issue was just meant to simplify the code.

I think I agree with all your  changes here (of course, _I_ would agree, you're basically going back to how I did things originally), but I worry about breaking people's custom code, at least @clemera's.

minad:
I don't think these variables are used much - I would remove this without a replacement and then fix upcoming issues on a case by case basis. The point is that these variables are entirely redundant.

I found a single use case https://github.com/unhammer/gnus-recent/blob/cc37ab30da76ec8c9e0357006928e21488099945/gnus-recent.el, but this could easily be replaced by modifying the embark-keymap-alist instead, setting only the t key.

minad:
I think https://github.com/oantolin/embark/pull/260/commits/d687aa100218509c20f95e1e1f99b1923d431619 is a satisfactory solution, which gets rid of the override variables, but still allows overriding.

oantolin:
Let's do it, I like simplification. If it causes any trouble at all it would only be for sophisticated users who can easily update their code.

PR #261, opened 2021-07-17
embark-consult: Make setup overrides private

minad:
The setup overrides are also private in embark.el. Furthermore these functions
seem internal and about implementation details.


oantolin:
Good point, thanks.

PR #262, opened 2021-07-17
Add embark-function-map (Fix #259)

### LINKED ISSUE #259: Add embark-function-map for function category
* Add function to embark-refine-symbol-type
* Add embark-function-map, derived from embark-symbol-map, add function specific commands

See https://github.com/minad/marginalia/commit/581d161445d1ad0b32b4050c1b8780fe2980d72f. I don't recall why we didn't add the function annotator earlier given that we also have a variable annotator.

minad:
This change is necessary since Marginalia reports a a special function completion category since I've added a function classifier and a function annotator, which shows the function arguments.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -124,6 +124,7 @@
     (symbol . embark-symbol-map)
     (command . embark-command-map)
     (variable . embark-variable-map)
+    (function . embark-function-map)
     (minor-mode . embark-command-map)
     (unicode-name . embark-unicode-name-map)
     (package . embark-package-map)
@@ -859,7 +860,11 @@ minibuffer before executing the action."
   (cons (or (when-let ((symbol (intern-soft target)))
               (cond
                ((commandp symbol) 'command)
-               ((boundp symbol) 'variable)))
+               ((boundp symbol) 'variable)
+               ;; Prefer variables over functions for backward compatibility.
+               ;; Command > variable > function > symbol seems like a
+               ;; reasonable order with decreasing usefulness of the actions.
+               ((fboundp symbol) 'function)))
             'symbol)
         target))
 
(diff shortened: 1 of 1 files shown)


oantolin:
I haven't been following what you've been doing with Marginalia, but does showing function arguments mean you now require Emacs 27. 1 or something like that?

minad:
Why does that require Emacs 27.1? I don't think you've missed anything significant. I am referring to this commit: https://github.com/minad/marginalia/commit/581d161445d1ad0b32b4050c1b8780fe2980d72f

Marginalia classifies functions now, this means we won't get Embark actions (beyond the general map) for functions anymore, except if we add the embark-function-map.

oantolin:
I said that thing about requiring Emacs 27.1, because I thought the introspection functions used to get arity and argument names, etc., are relatively new. I remember wanting to get function arity and giving up on it because the version of Emacs I used at the office didn't have that function yet, but I don't remember precise version numbers.

minad:
Okay, I understand. I use code derived from `elisp-get-fnsym-args-string`. I assume it is robust.

~~~ elisp
(defun marginalia--function-args (sym)
  "Return function arguments for SYM."
  (let ((tmp))
    (elisp-function-argstring
      (cond
       ((listp (setq tmp (gethash (indirect-function sym)
                                  advertised-signature-table t)))
        tmp)
       ((setq tmp (help-split-fundoc
		   (ignore-errors (documentation sym t))
		   sym))
	(substitute-command-keys (car tmp)))
       (t (help-function-arglist sym))))))
~~~

But honestly, I am not testing anymore with Emacs 26. I have Emacs 27.1 around as my default install and from time to time play around with Emacs 28 to create patches. I will not break 26 support deliberately, but it may happen - then it can be fixed on a case by case basis.

PR #264, opened 2021-07-19
Fix issue with displaying unescaped messages.

ahyatt:
If the point under embark has %s in it, it will be interpreted as a format arg by message.  To fix this, we escape all "%", with "%%".


minad:
Why not use "%s"?

ahyatt:
I'm not sure that I understand the solution you are thinking of.

oantolin:
I decided to go with @minad's suggestion insetad, @ahyatt.

ahyatt:
Thanks, now that I see @minad's suggestion, that also makes sense.

PR #270, opened 2021-07-22
Prefer pp-eval-expression

minad:
pp-eval-expression is a bit more flexible since it shows large results in an extra buffer. Do we want to use this instead of eval-expression?

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -236,6 +236,7 @@ It can be overriden by the `embark-setup-overrides' alist."
   '((async-shell-command embark--shell-prep)
     (shell-command embark--shell-prep)
     (eval-expression embark--eval-prep)
+    (pp-eval-expression embark--eval-prep)
     (package-delete minibuffer-force-complete))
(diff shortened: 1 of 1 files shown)


oantolin:
This was also part of the s-expression PR, right? I think I gave up on figuring out a quick improvement to that one (and at any rate if I do come up with something I can always do it later), so I want to merge that one.

oantolin:
I just checked and this is included in your other PR as I remebered, so I'm closing this one.

minad:
Yes, I thought you weren't unsure about the s-exp PR so I extracted that part here. But now that you merged the sexp PR we are good. And yes, if we figure out a different approach we can still do it.

oantolin:
Indeed I wasn't sure the s-exp PR was the best way to go (and still am not!), but it's a perfectly valid way to go and any alternative implementation could be easily done at a later time, so I really should have merged it long ago. I do apologize.

PR #272, opened 2021-07-22
Add embark-export-customize

minad:


oantolin:
Great idea!

minad:
Sparked by the customize discussion on reddit, I am astonished that this is so easy :smile:

oantolin:
Looks ready to merge to me. Was there anything else you wanted to do first?

minad:
Yes, it works well like this. From my side you can merge this as is. If we come up with further improvements we can still apply them later.

 One thing we could do - group the variables by `custom-variable-p` such that the customizable variables come first and the non-customizables come later. The customize buffer complains "NO CUSTOMIZATION DATA" for the non-custom-variables. But then we change the export order, you won't like this? Alternatively we could tone the complaint down a bit, by adjusting the field `rogue` in the `custom-magic-alist`.

minad:
I think the customize interface suddenly became much more useful :tada:

oantolin:
I don't really mind the NO CUSTOMIZATION DATA messages. I also wouldn't mind reordering the candidates if you think it looks better to have all of those together at the end. How about we keep it simple for now and see if it is really bothersome in practice?

minad:
Agree, I played a bit around with it and it is good as is. We are not filtering anything out and the UI will show the full truth.

oantolin:
This is really great. Thanks again!

PR #273, opened 2021-07-23
embark-export-customize: Cleanup UI

minad:
For our purposes we don't need the search field. What do you think?

EDIT: Okay, the search field runs `customize-apropos`, so I will just disable the search field globally.


oantolin:
And people who do use the search field, will probably want it in these exported buffers too.

PR #275, opened 2021-07-23
Ensure that expression actions work for defuns

minad:
Do you have some odd whitespace setting btw?

Please check if you think it is correct to do it like this.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -2167,8 +2167,10 @@ before or after the sexp (those are the two locations at which
 `embark-target-expression-at-point' detects a sexp)."
   `(defun ,(intern (format "embark-%s" cmd)) ()
      ,(format "Run `%s' on the sexp at or before point." cmd)
-     (interactive)     
+     (interactive)
      (goto-char (car (bounds-of-thing-at-point 'sexp)))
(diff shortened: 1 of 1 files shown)


minad:
Oh, this should also be adjusted to your changes in https://github.com/oantolin/embark/commit/37aa15d201cfd9b90a869da1d2b65a5322a203c2.

oantolin:
I don't think it's correct. If point is on the `f` of `(defun ...)`, wouldn't you just move it one character to `e`?

minad:
The point will be on the first character of `defun`.

oantolin:
I'm very confused:

1. `backward-char` would move you to `e`, no?
2. Even if you do get moved to `d` as you say, that's still not enough: you want to move to the `(`.

oantolin:
Also, I have no idea how I wind up with that extra whitespace. :(

minad:
You missed the `(goto-char (car (bounds-of-thing-at-point 'sexp)))` right before the parenthesis check.

oantolin:
> You missed the `(goto-char (car (bounds-of-thing-at-point 'sexp)))` right before the parenthesis check.

I could have sworn that line was in red. 😛 Sorry about the confusion.

PR #276, opened 2021-07-24
embark--show-indicator: Show target type when acting at point

minad:
It is good to see the type of the target when acting at point. This becomes more important when we add more types. It also makes the magic of Embark a little bit easier to understand and helps debugging target finders ;)

Also simplify `embark--show-indicator` - do not return overlays anymore. I think we talked about this a long time ago, but this has been forgotten since.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -202,12 +202,13 @@ Used by `embark-completing-read-prompter' and `embark-keymap-help'.")
 
 (defcustom embark-action-indicator
   (let ((act (propertize "Act" 'face 'highlight)))
-    (cons act (concat act " on '%s'")))
+    (cons act (concat act " on %2$s '%1$s'")))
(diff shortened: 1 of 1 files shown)


oantolin:
Sure, let's do this: I've almost added the type myself at some point.

Did you forget to update the docstring of the variable `embark-action-indicator` to mention the type?

oantolin:
Oh and the docstring for `embark--show-indicator ` looks like it mentions TYPE in the wrong place: it doesn't say that it's used in the format string case. And it does mention it in the function case and while it does say the TYPE is passed as an argument to the indicator function, it's a little misleading to put it there and some people might think that's what you're saying.

minad:
Okay, I fixed the docstrings and also passed TYPE to the indicator function.

oantolin:
Looks good, thanks!

Issue #92, opened 2021-01-08
Inability to consider current context as matching multiple target types


jyp:
I would be able to use Embark as a general context-dependent way to execute actions. This may mean that several things may be relevant in the current context (and thus several keymaps may apply). Currently, only the first matching target is taken into account.

Why do I want to do this? For example, there may be simultaneously 1. a flymake error at point to deal with and 2. a symbol whose definition can be looked up.  Another use-case would be to handle all xref-identifiers uniformly (do `xref-find-definition` on the identifier at point), which currently with the elisp symbol definition (which is only valid in elisp contexts, but offers a wider set of actions).

Do you think that the scope of Embark can be widened to support this sort of use-case? Perhaps there is another package which is more suitable for this?

oantolin:
I've wanted this too for a while! The case that bug me the most is functions, variables or macros with the same name as a package. For example, I've tried several times to run describe-symbol on `pcase`, only to be reminded that ffap got to it first and I can only do file actions on `pcase.el`.

The fairly recent change from separate target finders and classifiers to functions that return a pair of a calssification symbol and a string target was made with an eye toward this.

There are some choices about how to present this functionality. I think that to start somewhere I might just gather all the targets found in the current context, and enable all their keymaps simultaneously. If this proves too frustrating with actions shadowing other actions, I guess I could prompt for which target you mean before enabling a keymaps. Something like that, probably some experimentation is needed to find a good balance.

(On a related note, for embark-become I recently switched from enabling the first key map containing the current command to enabling all such key maps.

minad:
What about nesting the keymaps somehow using a reserved key.

~~~
keymap of first target
x -> keymap of second target
x x -> keymap of third target
x x x -> keymap of first target
~~~

I don't know if this works but one could even cycle through the keymaps using long prefix key sequences :)

Maybe Emacs refuses to handle cyclic prefix keymaps?

jyp:
@oantolin Cool! For my fingers memory I think that I need not clash between keymaps, so enabling them all seems the right option. I might try to code it up, if so I'll submit a PR.

oantolin:
I wrote a quick sketch in the multiple-targets branch 73a4c6248c30a86f221adbc440e4871897379c57. Please test, @jyp.

It at least solves my problem of wanting to run `h` on `pcase`. But there is some shadowing, for example, `d` on `pcase` will delete the file `pcase.el.gz`, and not go to definition.

[...]

PR #277, opened 2021-07-24
Proposal: Adjust defun finder to match less

minad:
* Recognize keywords as symbols and not as variables
* The defun finder requires a def in the name, otherwise
  it will recognize toplevel expressions too aggressively, e.g.,
  (mes|sage "..."), where | is the point.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -521,7 +521,7 @@ This function mostly relies on `ffap-file-at-point', with two exceptions:
   (when-let (bounds (bounds-of-thing-at-point 'defun))
     (let ((str (buffer-substring (car bounds) (cdr bounds))))
       (when (and
-             (string-match "\\`(\\(?:\\w\\|\\s_\\)+" str)
+             (string-match "\\`(\\(?:\\w\\|\\s_\\)*def\\(?:\\w\\|\\s_\\)*" str)
              (or (>= (point) (1- (cdr bounds)))
                  (<= (point) (+ (car bounds) (match-end 0)))))
         (cons 'defun str)))))
@@ -895,6 +895,8 @@ minibuffer before executing the action."
   "Refine symbol TARGET to command or variable if possible."
   (cons (or (when-let ((symbol (intern-soft target)))
               (cond
+               ;; keywords are boundp?
+               ((keywordp symbol) 'symbol)
                ((commandp symbol) 'command)
                ((boundp symbol) 'variable)
                ;; Prefer variables over functions for backward compatibility.


oantolin:
I agree with the keyword change.

But I'm not so sure about the change to the defun target finders. I mean, in Emacs defun _does_ just mean toplevel expression. Some users of `beginning-of-defun`, `mark-defun`, `edebug-defun` and `eval-defun` might be used to this meaning. I personally do use `mark-defun` and `eval-defun` fairly often on toplevel expression with no "def" in the symbol name (not through embark, directly using the default bindings, which have the advantage of working anywhere in the expression). This change you are proposing does maybe remove some surprising behavior, but at the same time that surprising behavior is standard in Emacs and removing it might be construed as a bug...

I guess the real problem is Emacs's odd choice of the name "defun". How sure are you we should make this change?

oantolin:
I went ahead and did the keyword change while we discuss the desirability of chaging the defun target finder.

minad:
> I went ahead and did the keyword change while we discuss the desirability of chaging the defun target finder.

Good!

> This change you are proposing does maybe remove some surprising behavior, but at the same time that surprising behavior is standard in Emacs and removing it might be construed as a bug...

You are right about that, but given that Embark does not support multiple targets at point yet I think it is better to go with this patch. But we can also keep the PR open and try to resolve first the multiple targets at point problem (#92).

> How sure are you we should make this change?

Not much, I suggest we keep the PR open as reminder and not merge it for now. Let's revisit this later.

hmelman:
> I mean, in Emacs defun _does_ just mean toplevel expression.

I'm not sure if this applying to just elisp or other languages too.  defun does means a function definition and the idea that an open paren or brace in the left margin is the start of a defun is just a performance optimization which can be overridden with `open-paren-in-column-0-is-defun-start`. See the section "Left Margin Convention" in the Emacs Manual.

minad:
@hmelman Okay what kind of craziness is this? If you set this to nil, will the thing-at-point defun not match anymore? If that is the case we should recommend disabling this optimization or disable it locally. Do you know what happens if one disables it globally? Is this optimization important, does it affect hot loops?

oantolin:
I'm not sure I agree with your reading of the manual, @hmelman. I think the situation is as follows:

- "defun" in Emacs does not just mean "function definition", it really is more general than that and means "toplevel expression".
- The optimization mentioned in the manual is that it assumes that any opening delimiter in the first column really does start a toplevel expression.

Except possibly for multiline strings and multiline comments (in languages that have them, Emacs Lisp doesn't), this optimization is harmless in a properly indented buffer. The multiline string problem does arise in Emacs Lisp, for example which is why the linter will recommend you escape opening parenthesis in the first column of a docstring.

With the optimiztion on `mark-defun` selects the inner `progn` in the following example (which is valid Emacs Lisp but is not nicely indented). But even with the optimization off, the outer `progn` counts as a defun:

```elisp
(progn
(progn
(message "Hello)))
```

minad:
@oantolin Okay, thanks for clarifying.

oantolin:
So, to focus the discussion a bit: which cases are you worried about, @minad? What are some examples where you'd prefer the symbol be recognized instead of the whole toplevel expression? Note there is a roundabout of making Embark act on the symbol: mark it, act on the region, press RET to act on the region contents. (I'm am emphatically _not_ saying this is a _comfortable_ workaround. 😛)

Also, how do you feel about a compromise: restricting the defun target finder but not as much as you propose. How about in addition to any symbol with `def` in the name, also including `use-package` and other toplevel forms from package managers? Maybe also other toplevel forms common in configuration such as `custom-set-variables`, `with-eval-after-load`, `add-to-list`, etc. (I am very used to Emacs's defun = toplevel-expression, so I don't really know what your perspective is here, when you want the defun to be recognized and when you wouldn't.)

minad:
@oantolin I would be okay with doing anything and keeping the target finder unrestricted. I am currently playing around with a prototype for #92, so this may resolve the issue in a better way. But of course it depends on if you like my solution for #92. I will send a PR in a while.

oantolin:
> I would be okay with doing anything and keeping the target finder unrestricted.

OK, let's do that for now.

> Your perspective of toplevel=defun makes sense.

Not mine! Emacs's. I just meekly got used to it.

> I am currently playing around with a prototype for #92, so this may resolve the issue in a better way. But of course it depends on if you like my solution for #92. I will send a PR in a while.

That would be amazing! I've been meaning to look at that again, but definitely am way slower and less energetic than you. Thank you!

hmelman:
> Okay what kind of craziness is this? 

@minad I didn't write it, I just wanted to make sure you were aware of it in this context.  😉 

I'm not sure what the current state of this option is.  There was [discussion a year ago](https://lists.gnu.org/archive/html/emacs-devel/2020-04/msg00272.html) and it comes up every now and then.

As best as I understand the heuristic is from very early emacs and does almost always work in lisp where defuns are at top level and the idea that a top level `(require 'foo)` or similar also worked was a bonus and was codified.  It's possible but unusual in lisp to have a defun not at the top level, e.g., within a `with-eval-after-load` or something.  In such cases for me with Emacs 27 I see the defun commands work on the top level expression and not just the second-level defun (with the option set to t or nil), but I think this is viewed as a hard to fix bug that can be lived with.  

But the reason I mentioned that I don't know if this PR is lisp only or not. In python, the defun commands work on functions even if they are within classes (which would be the top level object) and that's very much desired.  I believe it's the same for Java (it's been a long time) and I don't know about C++.

minad:
@hmelman Thanks for the context. I close this in favor of #278, since with that it is much less problematic if multiple finders match something at point.

PR #278, opened 2021-07-24
Implement multiple targets at point (Fix #92)

### LINKED ISSUE #92: Inability to consider current context as matching multiple target types

minad:
When not acting in the minibuffer all target finders are executed. The action
indicator will then indicate that multiple targets exist. By pressing the
`embark-act` key again, the user can cycle to the next target.

The alternative approach discussed in #92 was to merge the keymaps. This
approach has disadvantages: Multiple targets are active at the same time and
depending on the selected action, the target is selected. This complicates the
current `embark--act` implementation, which is untouched by this PR. Furthermore
keybindings are shadowed, which makes the individual keymaps a lot less useful.
This shadowing will lead to confusion and it will not be obvious to the user
which target is actually being used.

The current approach is also easy to implement, it fits well within the existing
codebase. This is a good indication to go this route.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -613,6 +613,10 @@ relative path."
                     (abbreviate-file-name (expand-file-name raw))
                   raw)))))))
 
+(defun embark--cycle-key ()
+  "Return the key to use for `embark-cycle'."
+  (or embark-cycle-key (car (where-is-internal #'embark-act))))
+
 (defun embark--action-keymap (type cycle)
   "Return action keymap for targets of given TYPE.
 If CYCLE is non-nil bind `embark-cycle'."
@@ -620,10 +624,7 @@ If CYCLE is non-nil bind `embark-cycle'."
    (let ((map (make-sparse-keymap)))
      (define-key map [13] (embark--default-action type))
      (when cycle
-       (define-key map
-         (or embark-cycle-key
-             (car (where-is-internal #'embark-act)))
-         #'embark-cycle))
+       (define-key map (embark--cycle-key) #'embark-cycle))
      map)
    (symbol-value (or (alist-get type embark-keymap-alist)
                      (alist-get t embark-keymap-alist)))))
@@ -753,16 +754,19 @@ If NO-DEFAULT is t, no default value is passed to `completing-read'."
                 (when embark-keymap-prompter-key
                   (use-local-map
                    (make-composed-keymap
-                    (let ((map (make-sparse-keymap)))
+                    (let ((map (make-sparse-keymap))
+                          (cycle (embark--cycle-key)))
                       ;; Rebind `embark-cycle' in order allow cycling
                       ;; from the `completing-read' prompter. Additionally
                       ;; `embark-cycle' can be selected via
                       ;; `completing-read'. The downside is that this breaks
                       ;; recursively acting on the candidates of type
                       ;; embark-keybinding in the `completing-read' prompter.
-                      (when-let (key (where-is-internal
-                                      #'embark-cycle keymap))
-                        (define-key map (car key) #'embark-cycle))
+                      (define-key map cycle
+                        (if (lookup-key keymap cycle) #'embark-cycle
+                          (lambda ()
+                            (interactive)
+                            (minibuffer-message "Only a single target"))))
                       (define-key map embark-keymap-prompter-key
                         (lambda ()
                           (interactive)
--- a/embark.el
+++ b/embark.el
@@ -829,8 +829,8 @@ be restricted by passing a PREFIX key."
     (when-let (command (embark-completing-read-prompter keymap 'no-default))
       (call-interactively command))))
 
-(defun embark--with-indicator (indicator prompter keymap targets)
-  "Display INDICATOR while calling PROMPTER with KEYMAP.
+(defun embark--prompt (indicator keymap targets)
+  "Display INDICATOR while calling the prompter with KEYMAP.
 The TARGETS are displayed for actions outside the minibuffer."
   (let ((remove-indicator (embark--show-indicator indicator keymap targets)))
     (unwind-protect
@@ -849,7 +849,7 @@ The TARGETS are displayed for actions outside the minibuffer."
                                             indicator)
                                           keymap targets))
               (let ((enable-recursive-minibuffers t))
-                (funcall prompter keymap)))
+                (funcall embark-prompter keymap)))
           (quit nil))
       (when (functionp remove-indicator)
         (funcall remove-indicator)))))
@@ -1060,11 +1060,10 @@ ARG is the prefix argument."
         (and
          (catch 'embark--cycle
            (pcase-let* ((`((,type . ,target) . (,_ . ,original)) (car targets))
-                        (action (or (embark--with-indicator embark-action-indicator
-                                                            embark-prompter
-                                                            (embark--action-keymap
-                                                             type (cdr targets))
-                                                            (mapcar #'car targets))
+                        (action (or (embark--prompt embark-action-indicator
+                                                    (embark--action-keymap
+                                                     type (cdr targets))
+                                                    (mapcar #'car targets))
                                     (user-error "Canceled")))
                         (default-action (embark--default-action type)))
              (embark--act action
@@ -1146,11 +1145,10 @@ point."
                      (pcase-let ((`(,beg . ,end) (embark--boundaries)))
                        (substring (minibuffer-contents) beg
                                   (+ end (embark--minibuffer-point))))))
-           (become (embark--with-indicator embark-become-indicator
-                                           embark-prompter
-                                           (embark--become-keymap)
-                                           ;; Pass a fake target list here
-                                           `((nil . ,target)))))
+           (become (embark--prompt embark-become-indicator
+                                   (embark--become-keymap)
+                                   ;; Pass a fake target list here
+                                   `((nil . ,target)))))
       (if (null become)
           (user-error "Canceled")
         (embark--quit-and-run
--- a/embark.el
+++ b/embark.el
@@ -196,13 +196,23 @@ The key must be either a string or a vector.
 This is the key representation accepted by `define-key'."
   :type '(choice key-sequence (const nil)))
 
+(defcustom embark-cycle-key nil
+  "Key used for `embark-cycle'.
+
+If the key is set to nil it defaults to the global binding of
+`embark-act'. The key must be either a string or a vector. This
+is the key representation accepted by `define-key'."
+  :type '(choice key-sequence (const nil)))
+
 (defface embark-keybinding '((t :inherit success))
   "Face used to display key bindings.
 Used by `embark-completing-read-prompter' and `embark-keymap-help'.")
 
 (defcustom embark-action-indicator
   (let ((act (propertize "Act" 'face 'highlight)))
-    (cons act (concat act " on %2$s '%1$s'")))
+    (cons act (concat act " on %2$s"
+                      (propertize "%3$s" 'face 'shadow)
+                      " '%1$s'")))
   "Indicator to use when embarking upon an action.
 If set to a string it is used as a format string where %1$s will
 be replaced by the target of `embark-act' and %2$s will be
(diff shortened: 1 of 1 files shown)


minad:
@oantolin I just saw that you had an initial implementation with the keymaps merged (https://github.com/oantolin/embark/commit/73a4c6248c30a86f221adbc440e4871897379c57). There is another disadvantage of that approach I forgot to mention above - since multiple targets are active at the same time, you cannot execute M-x on the mixtarget.

jaor:
On Sat, Jul 24 2021, Daniel Mendler wrote:

> When not acting in the minibuffer all target finders are executed. The action
> indicator will then indicate that multiple targets exist. By pressing the
> embark-act key again, the user can cycle to the next target.

apologies if this has already been discussed, but how do user-defined
action indicators fit in this picture? do they receive a list of keymaps
and a list of targets?

minad:
@jaor User defined indicators receive the keymap of the current target and a list of all `(type.target)` pairs. The first element of the list will be the current target. When the user cycles between the targets a new action indicator will be created and destroyed every time.

jaor:
On Sat, Jul 24 2021, Daniel Mendler wrote:

> @jaor User defined indicators receive the keymap of the current target
> and a list of all (type.target) pairs.  The first element of the list
> will be the current target. When the user cycles between the targets a
> new action indicator will be created and destroyed every time.

I see, that works nicely.  Just curious here: why calling with a list to
type/target if only a keymap is sent? Wouldn't it be simpler to call
with just keymap type and target?

minad:
> I see, that works nicely.  Just curious here: why calling with a list to
type/target if only a keymap is sent? Wouldn't it be simpler to call
with just keymap type and target?

The idea is that the action indicator should have the possibility to indicate that other shadowed targets exist. The indicator can show the current target, its keymap and a list of alternative targets, which can be reached by cycling.

(I've seen that your action indicator has been mentioned in a discussion on reddit. Maybe it would be a good addition to Embark?)

oantolin:
This is a brilliant, @minad! I really wanted common cases to require no further keypresses, and this cycling idea achieves this if we can order the target finders carefully, prioritizing the type whose actions are most commonly needed.

It was important to me that whatever UI was chosen was composable in the sense you describe: that you can add new targets and keymaps without worrying about shadowing. I thought that maybe the only practical way to achieve that was to make embark-act a two-stage processes: first show a menu of targets and once the user picks one go through the standard embark-act process. But the menu of targets works against the idea of common cases taking no more keystrokes than currently, and also it seemed like presenting the menu would need a pop-up or enlarging the minibuffer, which I prefer to avoid (I know I'm weird about that and most people like huge minibuffers like you get with Vertico, Selectrum or Which-key).

But, as long as not too many targets are found at the same location, I think this cycling idea is pretty good. I hate cycling among 5 things, I love cycling between 2 things. I think 3 is OK and 4 is tolerable. (I know because I've tried all those values for `completion-cycle-threshold` :P.) I guess we can refine the target finders to keep the average number of targets small (and they're probably already pretty much there). I guess we can always add later a command that prompts for a target and cycles immediately to it.

Also, as you say we can have the sexp and defun target finders trigger anywhere in the corresponding region to get an expand-region like effect.

minad:
> This is brilliant, @minad! I really wanted common cases to require no further keypresses, and this cycling idea achieves this if we can order the target finders carefully, prioritizing the type whose actions are most commonly needed.

Thanks, I am glad you like it!

> (I know I'm weird about that and most people like huge minibuffers like you get with Vertico, Selectrum or Which-key).

Maybe I should try setting vertico-count to 3 ;)

> But, as long as not too many targets are found at the same location, I think this cycling idea is pretty good. I hate cycling among 5 things, I love cycling between 2 things. I think 3 is OK and 4 is tolerable.
> I guess we can always add later a command that prompts for a target and cycles immediately to it.

I don't think we won't get significantly more than 3 targets at point. But if that would happen there is also a good solution, we could add a prefix argument to `embark-cycle` which lets you jump to the right candidate right away or as you say add a separate command bound to some special jumping keys `# 1`, `# 2`, ...

At least the solution here can potentially scale. Good luck with merging keymaps if you have 5 possible targets at point.

Please let me know if there are things that should be addressed. For stylistic changes I guess you are faster to do it yourself, like changing mapcars and friends to cl-loop :)

minad:
> ...we can order the target finders carefully, prioritizing the type whose actions are most commonly needed.

One possibility I want to point out - we can even add multiple target finders with different priorities returning the same type and target. This allows for very fine-grained tuning. See also my comment in `embark--targets`.

oantolin:
I'm trying it out and loving it!

Here are a few minor quick impressions:

1. I think the default indicator should probably have the shadowed types come after the target, it feels weird to have the type and target separated by anything other than whitespace. Maybe `type 'target' (… shadowed)`?

2. The default target finders will need a little tweaking and reordering, but are already pretty good as is!

3. When you use the default `nil` setting of `embark-cycle-key` with the completing-read prompter, the single target case can be jarring: if you repeat the `embark-act` keybinding you get multiple nested `embark-act`s.

minad:
> I think the default indicator should probably have the shadowed types come after the target, it feels weird to have the type and target separated by anything other than whitespace. Maybe type 'target' (… shadowed)?

I did it like this since then the target position does not jump when cycling, but maybe your formatting is easier to understand?

~~~
symbol(defun) targetsym
defun(symbol) targetdefun
~~~

I am not sure which one I prefer.

~~~
Act on "target" type (shadowed)
Act on type(shadowed) "target"
Act on type "target" (shadowed)
~~~

But this tweaking can also come later. Since we are modifying the action indicator right now, I thought that the current code could also be a bit cleaned up in a separate PR. Make `embark-action-indicator` obsolete and use functions for everything.

~~~ elisp
(defvar embark-action-indicator-function #'embark-default-action-indicator)
(defvar embark-default-action-indicator-format ("%s...))
~~~

> When you use the default nil setting of embark-cycle-key with the completing-read prompter, the single target case can be jarring: if you repeat the embark-act keybinding you get multiple nested embark-acts.

Yes, but this has already been the case? This is when you act on the binding candidate itself. We could decide to always bind `embark-cycle` in the completing read prompter and if there is only a single candidate, print a message.

jaor:
On Sat, Jul 24 2021, Daniel Mendler wrote:

> The idea is that the action indicator should have the possibility to
> indicate that other shadowed targets exist. The indicator can show the
> current target, its keymap and a list of alternative targets, which
> can be reached by cycling.

Aha, that makes sense.  I don't have a very concrete use case here, but
i am thinking that receiving the full list of keymaps (together with
types and targets) could be considered: imagine i want to hack an
indicator that uses transient or hydras and that itself provides access
to the other keymaps (not necessarily in cycling order), or something of
the sort... thinking out loud here...

> (I've seen that your action indicator has been mentioned in a
> discussion on reddit. Maybe it would be a good addition to Embark?)

i'd be happy to send a PR if you guys think it's worth it... it might
have some rough edge, but i guess we could refine it as needed.  Or add
it to some wiki or the other.

oantolin:
> Make `embark-action-indicator` obsolete and use functions for everything.

That probably would be simpler. Good idea.

> But this tweaking can also come later.

Agreed.

> Yes, but this has already been the case? This is when you act on the binding candidate itself.

Yes, of course. It just never bothered me before because I never had a reason to press the embark-act key twice in a row. :) I'm not saying anything needs to be done about this, just pointing it out.

minad:
> Yes, of course. It just never bothered me before because I never had a reason to press the embark-act key twice in a row. :) I'm not saying anything needs to be done about this, just pointing it out.

Fixed in https://github.com/oantolin/embark/pull/278/commits/46e68abfdc1f61e7a2d2a0db3b2dca1d1c540b46

oantolin:
@jaor I would love a pull request with an indicator like that, I'd probably make it the default too (but most likely wouldn't use it myself, I don't like "always on" reminders, preferring to summon them when I need them). When I saw a screenshot I do remember thinking that it would make sense to rewrite it to create an Embark collect buffer, since you wouldn't have to do any layout by hand and also would get the possibility of toggling to grid view for free (which looks a lot like which-key, in fact).

minad:
> When I saw a screenshot I do remember thinking that it would make sense to rewrite it to create an Embark collect buffer, since you wouldn't have to do any layout by hand and also would get the possibility of toggling to grid view for free (which looks a lot like which-key, in fact).

This would be kind of meta :)

oantolin:
> Fixed in [46e68ab](https://github.com/oantolin/embark/commit/46e68abfdc1f61e7a2d2a0db3b2dca1d1c540b46)

Well, now how do I act on the completion candidates? 😆  You've never used goto definition on an action from the completing-read prompter? 😉

> Regarding the embark-action-indicator, we can do this in a separate PR.

Yes, of course. I wasn't suggesting to do it here.

oantolin:
Actually `embark-act` does seem to work in the completing-read prompter. Weird, I'd never noticed that before.

oantolin:
And, just to be explicit, I was kidding earlier I think the message in 46e68ab is a better default.

minad:
@jaor 

> ...receiving the full list of keymaps (together with
types and targets) could be considered: imagine i want to hack an
indicator that uses transient or hydras and that itself provides access
to the other keymaps (not necessarily in cycling order), or something of
the sort... thinking out loud here...

This is not really possible since only one target is active at a time. In other words, the action indicator and not even the prompter should take control over the keymaps. But we can still reconsider this later. We will probably rework the action indicator later a bit.

@oantolin 

> Well, now how do I act on the completion candidates? laughing You've never used goto definition on an action from the completing-read prompter? wink

M-x :D

oantolin:
So, I'd be delighted to merge now after kicking the tires. Anything else you wanted to do before I merged, @minad?

minad:
From my side it is ready, I added these two PRs (1. adding the message, 2. rename the with-indicator to more meaningful embark--prompt). But overall it seems ready from my side!

oantolin:
Fantastic! Thanks so much, @minad.

oantolin:
I'll play around with it a couple of days, maybe refine or reorder some target finders and then bump the version number.

minad:
Sounds good! Maybe also experiment with broadening/simplifying some of the target finders, e.g., the expression and defun finder as we discussed?

oantolin:
Yes. I think I'd prefer to have those apply everywhere in their regions and move them toward the end.

As you pointed in the comment out one could also add a copy the defun matching only `.*def.*` with higher priority, but I don't think that's necessary.

I think I'll also tighten the file target so it doesn't trigger on web urls. It gives a silly tramp style `/https:/github.com/oantolin/embark` that doesn't actually seem to work as a path.

oantolin:
Oh, and the sexp target finder should not match defuns, since the defun keymap already inherits from the sexp keymap.

minad:
> Oh, and the sexp target finder should not match defuns, since the defun keymap already inherits from the sexp keymap.

Probably both should be replaced by a single finder+refinement, like for the identifiers?

oantolin:
> Probably both should be replaced by a single finder+refinement, like for the identifiers?

No, I think that if I'm somewhere in the middle of a function, I'd like to be able to act on the smallest containing list `(...)` and also on the whole defun.

oantolin:
> Generally, false positives in the low priority filters are a problem now unfortunately. But it should be possible to tighten all these finders.

I think they're not really that many. I bet almost all come out in the next couple of days.

minad:
> I'd like to be able to act on the smallest containing list (...) and also on the whole defun.

Yes this is the behavior I would like to have.

Issue #269, opened 2021-07-20
Highlight the chosen symbol/entity ala isearch-forward-symbol-at-point


jdtsmith:
The command `isearch-forward-symbol-at-point` is in many ways like a tiny baby embark.  It looks for a symbol at point, and sets you up for further action on it (search or occur for example).  One of the features it has that I'm really missing with embark: it _highlights_ the matched symbol temporarily, showing you what it will be taking action _on_. Here it is in action:

![image](https://user-images.githubusercontent.com/93749/126391318-5ab67c3f-aa34-494f-a825-ddc63fc3ab9d.png)

Would be great if embark stole this idea!

oantolin:
That's an interesting idea. If you use Embark outside of the minibuffer, it displays what you will act on in the echo area. In part this is because it was easier but also because the target finders are under no obligation to return a portion of buffer text! Even pretty standard ones do not, for example Embark uses ffap for file names, and this returns full paths, not just whatever is found in the buffer. If you use `embark-act` on `dired` in `(use-package dired ...)` the target will actually be `/usr/share/emacs/27.1/lisp/dired.el` or whatever the appropriate path is on your system.

I guess most of the target finders that come with Embark do provide targets that are at least based on a portion of the buffer text, and this portion could be highlighted. But if we do this it could be slightly confusing that what is acted on does not exactly match what is highlighted.

jdtsmith:
Interesting thanks.  I actually hadn't realized embark gives a report in the mini buffer of what it will act on! 

From the start I was using the recommended [`which-key`](https://github.com/oantolin/embark#showing-a-reminder-of-available-actions) trick to prompt with the further key options immediately, which I find preferable to always having to `C-h` and select/sort (I believe Protesilaos Stavrou recommends this in some of his videos too).  This trick apparently (and unfortunately) overwrites the mini buffer embark prompt of what it will be acting upon, which makes it feel like I'm "matching in the dark".  Maybe I'll try without which-key, or maybe the trick could be modified to not stomp on embark's message?

I actually opened a related [consult issue](https://github.com/minad/consult/issues/364) to get it to care about the symbol-at-point highlighted selection if that's set.  It's an underused but cool native Emacs feature.  Please feel free to close. Thanks for embark!

oantolin:
Ah! Which-key strikes again. I wonder if there is anywhere to put the target in which-key's popup. That would probably be the best solution.

oantolin:
Try this:

```elisp
[...]

PR #279, opened 2021-07-25
Target highlight

minad:
I hope you don't mind me throwing another PR at you. I am a big fan of the universal Embark at point functionality! Addresses #269 partially. Some of the finders must still be overhauled, but this is also due to #278/#92.

EDIT: I think all the at point target finders work properly with highlighting now. I also deprioritized and broadened the sexp and defun finders.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -162,8 +162,8 @@ a string, or nil to indicate it found no target."
     (embark-keybinding . embark-keybinding-command)
     (project-file . embark-project-file-full-path))
   "Alist associating type to functions for transforming targets.
-Each function should take a target string and return a pair of
-the form a `cons' of the new type and the new target."
+Each function should take a type and a target string and return a
+pair of the form a `cons' of the new type and the new target."
   :type '(alist :key-type symbol :value-type function))
 
 (defcustom embark-become-keymaps
@@ -928,7 +928,7 @@ minibuffer before executing the action."
           (funcall run-action)
         (embark--quit-and-run run-action)))))
 
-(defun embark-refine-symbol-type (target)
+(defun embark-refine-symbol-type (_type target)
   "Refine symbol TARGET to command or variable if possible."
   (cons (or (when-let ((symbol (intern-soft target)))
               (cond
@@ -942,12 +942,12 @@ minibuffer before executing the action."
             'symbol)
         target))
 
-(defun embark-keybinding-command (target)
+(defun embark-keybinding-command (_type target)
   "Treat an `embark-keybinding' TARGET as a command."
   (when-let ((cmd (get-text-property 0 'embark-command target)))
     (cons 'command cmd)))
 
-(defun embark-lookup-lighter-minor-mode (target)
+(defun embark-lookup-lighter-minor-mode (_type target)
   "If TARGET is a lighter, look up its minor mode.
 
 The `describe-minor-mode' command has as completion candidates
@@ -964,7 +964,7 @@ work on them."
 (declare-function project-roots "project")
 (declare-function project-root "project")
 
-(defun embark-project-file-full-path (target)
+(defun embark-project-file-full-path (_type target)
   "Get full path of project file TARGET."
   ;; TODO project-find-file can be called from outside all projects in
   ;; which case it prompts for a project first; we don't support that
@@ -992,8 +992,8 @@ buffers, each target finder function is executed.
 
 For each target, the type is then looked up as a key in the
 variable `embark-transformer-alist'. If there is a transformer
-for the type, it is called with the target, and must return a
-`cons' of the transformed type and transformed target.
+for the type, it is called with the type and target, and must
+return a `cons' of the transformed type and transformed target.
 
 The return value of `embark--targets' is a list of pairs, where
 each car is the transformed type and target and each cdr is the
@@ -1006,7 +1006,7 @@ original type and target."
          (push (if-let (transformer (alist-get
                                      (car target)
                                      embark-transformer-alist))
-                   (cons (funcall transformer (cdr target)) target)
+                   (cons (funcall transformer (car target) (cdr target)) target)
                  (cons target target))
                targets)
          (minibufferp))))
@@ -1887,7 +1887,7 @@ buffer for each type of completion."
         ;; check to see if all candidates transform to same type
         (when transformer
           (pcase-let* ((`(,new-type . ,first-cand)
-                        (funcall transformer (car candidates))))
+                        (funcall transformer type (car candidates))))
             (unless (eq type new-type)
               (when-let ((new-exporter
                           (alist-get new-type embark-exporters-alist))
@@ -1895,7 +1895,7 @@ buffer for each type of completion."
                 (when (cl-every
                        (lambda (cand)
                          (pcase-let ((`(,t-type . ,t-cand)
-                                      (funcall transformer cand)))
+                                      (funcall transformer type cand)))
                            (when (eq t-type new-type)
                              (push t-cand new-candidates)
                              t)))
--- a/embark.el
+++ b/embark.el
@@ -146,10 +146,10 @@ For any type not listed here, `embark-act' will use
     embark-target-bug-reference-at-point
     embark-target-url-at-point
     embark-target-file-at-point
-    embark-target-defun-at-point
     embark-target-expression-at-point
     embark-target-custom-variable-at-point
-    embark-target-identifier-at-point)
+    embark-target-identifier-at-point
+    embark-target-defun-at-point)
   "List of functions to determine the target in current context.
 Each function should take no arguments and return either nil to
 indicate that no target has been found, a cons (type . target)
@@ -542,12 +542,7 @@ In `dired-mode', it uses `dired-get-filename' instead."
 (defun embark-target-defun-at-point ()
   "Target defun at point."
   (when-let (bounds (bounds-of-thing-at-point 'defun))
-    (let ((str (buffer-substring (car bounds) (cdr bounds))))
-      (when (and
-             (string-match "\\`(\\(?:\\w\\|\\s_\\)+" str)
-             (or (>= (point) (1- (cdr bounds)))
-                 (<= (point) (+ (car bounds) (match-end 0)))))
-        `(defun ,str . ,bounds)))))
+    `(defun ,(buffer-substring (car bounds) (cdr bounds)) . ,bounds)))
 
 (defun embark-target-identifier-at-point ()
   "Target identifier at point.
--- a/embark.el
+++ b/embark.el
@@ -2433,7 +2433,7 @@ and leaves the point to the left of it."
   ("i" embark-indent-sexp)
   ("r" embark-raise-sexp)
   ("k" embark-kill-sexp)
-  ("@" embark-mark-sexp))
+  ("SPC" embark-mark-sexp))
 
 (embark-define-keymap embark-defun-map
   "Keymap for Embark defun actions."
@@ -2445,7 +2445,7 @@ and leaves the point to the left of it."
   ("d" edebug-defun)
   ("o" checkdoc-defun)
   ("n" narrow-to-defun)
-  ("h" mark-defun))
+  ("SPC" mark-defun))
 
 (embark-define-keymap embark-symbol-map
   "Keymap for Embark symbol actions."
--- a/embark.el
+++ b/embark.el
@@ -529,6 +529,9 @@ In `dired-mode', it uses `dired-get-filename' instead."
                (re-search-forward ":" (line-end-position) 'noerror)
                (point)))))))
 
+;; NOTE: There is also (thing-at-point 'list), however it does
+;; not work on strings and requires the point to be inside the
+;; parentheses. This version here is slightly more general.
 (defun embark-target-expression-at-point ()
   "Target expression at point."
   (save-excursion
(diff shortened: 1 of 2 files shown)


embark.el
```diff
@@ -526,18 +526,22 @@ In `dired-mode', it uses `dired-get-filename' instead."
...
-                      (eq (syntax-class (syntax-after pt)) 4)
-                      (setq begin (1+ begin))))
-         `(expression ,(buffer-substring begin end) . ,bounds))))))
+  (save-excursion
+    (let ((pt (point)))
+      (catch 'found
+        (while
+            ;; Looking at opening parenthesis or find last one
+            (or (memq (syntax-class (syntax-after pt)) '(4 6 7))
+                (re-search-backward "\\(\\s(\\|\\s/\\|\\s\"\\)" nil 'noerror))
```
oantolin, line 535: I would have thought that using (bounds-of-thing-at-point 'list) would have been easier. Was there some problem with that approach?

oantolin:
> I hope you don't mind me throwing another PR at you. I am a big fan of the universal Embark at point functionality!

Mind? Are you kidding? On the contrary, I'm very grateful! This is great stuff.

embark.el
```diff
@@ -466,55 +471,63 @@ There are three kinds:
...
+      (save-excursion
+        (end-of-line)
+        `(file ,(abbreviate-file-name file)
+               ,(save-excursion
+                  (re-search-backward " " (line-beginning-position) 'noerror)
+                  (1+ (point)))
+               . ,(point)))
+    (when-let (file (ffap-file-at-point))
+      `(file ,(abbreviate-file-name file)
+             . ,(bounds-of-thing-at-point 'filename)))))
```
oantolin, line 494: Does use the bounds from thingatpt really work? I used ffap because it's much smarter and detects file names in many more cases. For example in `(require 'lib)` ffap knows `lib` is a file. Does thingatpt return the correct bounds in that case?

embark.el
```diff
@@ -466,55 +471,63 @@ There are three kinds:
...
+  (when-let ((ov (seq-find (lambda (ov) (overlay-get ov 'bug-reference-url))
+                           (overlays-at (point)))))
+    `(url ,(overlay-get ov 'bug-reference-url)
+          ,(overlay-start ov) . ,(overlay-end ov))))
 
 (defun embark-target-url-at-point ()
   "Target the URL at point."
   (when-let ((url (ffap-url-at-point)))
-    (cons 'url url)))
+    `(url ,url . ,(bounds-of-thing-at-point 'url))))
```
oantolin, line 506: Same question about thingatpt vs ffap here.

embark.el
```diff
@@ -466,55 +471,63 @@ There are three kinds:
...
 
+(declare-function widget-at "wid-edit")
 (defun embark-target-custom-variable-at-point ()
   "Target the variable corresponding to the customize widget at point."
   (when (derived-mode-p 'Custom-mode)
-    (when-let ((symbol (get-text-property (point) 'custom-data)))
-      (cons 'variable (symbol-name symbol)))))
+    (save-excursion
+      (beginning-of-line)
+      ;; TODO Please check, the old custom variable finder did not work for me.
```
oantolin, line 514: I think the old one may have needed Emacs 28.

embark.el
```diff
@@ -666,7 +683,11 @@ when the indicator is no longer needed."
                           indicator)
                         target type
                         (if (cdr targets)
-                            (format "%S" (mapcar #'car (cdr targets)))
+                            ;; This is a weird feature of format/prin1-to-string.
+                            ;; A symbol list (function weird arbitrary symbols)
+                            ;; is printed as #'weird!
```
oantolin, line 687: I noticed!

embark.el
```diff
@@ -879,6 +900,8 @@ minibuffer before executing the action."
                            (not (memq action embark-skip-edit-commands))
                          (memq action embark-allow-edit-commands)))
            (inject
+            ;; TODO consider using strings for regions,
```
oantolin, line 902: I'd like to use strings. The only issue are some actions like write-region that prompt for a file name or something and would have the region contents incorrectly injected. There aren't that many of those, I think, so we just need to wrap them. An `embark--ignore-target-command` macro would be good enough.

minad:
> Mind? Are you kidding? On the contrary, I'm very grateful! This is great stuff.

Thanks :heart:

embark.el
minad: Oh, I wasn't aware of 'list, does it do the same expansion as I am doing here?

embark.el
minad: Not really, there should probably be a TODO here to make this more general. It is a best effort thing, we may get wrong bounds but the only thing which will break is the hl. Unfortunately ffap does not make the bounds available.

embark.el
minad: Same answer.

embark.el
minad: Ah okay. I didn't test on 28. Did you add this recently?

embark.el
minad: Great! I was really confused what is going on here. The usage of lists as function seems to be a weird design in elisp. Compiled functions should be opaque :laughing:

embark.el
minad: Okay, let's just keep the TODO for now?

embark.el
minad: Okay, this seems to work as expected. Minor difference is that the point must be really inside the sexp, while my finder also works if you are on the parentheses. But this is probably not worth the complication!

~~~ elisp
(defun embark-target-expression-at-point ()
  "Target expression at point."
  (when-let (bounds (bounds-of-thing-at-point 'list))
    `(expression ,(buffer-substring (car bounds) (cdr bounds)) . ,bounds)))
~~~

minad: Oh, and it doesn't work with strings. Hmm, I am not sure. Maybe keep my more complicated version?

embark.el
oantolin: Yes, pretty recently. I'll test yours on Emacs 28. If this version work on 27 and 28, it's definitely better.

embark.el
oantolin: Sure.

embark.el
oantolin: Yes. Maybe in the future I'll add a separate string target, just to include the fill-paragraph action for it. When that happens we could maybe simplify this one (but keeping the matching on the parentheses).

embark.el
oantolin: Yes, it's very old-fashioned. I think many Lisps used to do that.

oantolin:
Expanding where the sexp and defun targets match broke all the actions. :(

minad:
> Expanding where the sexp and defun targets match broke all the actions. :(

Uh right, the primitive movement to the beginning does not work anymore. What about using the target finder itself to find the boundaries, now that the finder returns them?

minad:
The defun map should also not inhert the expression map anymore I guess. Or alternatively the map should still inherit but the expression finder should not find a target when the defun finder matches. I have to think a bit about this.

oantolin:
If we expose the bounds, say put them in an `embark--target-bounds` variable, we could write mark, kill and indent actions that work for all at-point targets at once.

minad:
> If we expose the bounds, say put them in an embark--target-bounds variable, we could write mark, kill and indent actions that work for all at-point targets at once.

Sounds like a good and simple idea. Otherwise the action would not know which finder to call.

oantolin:
I'll add the variable. I don't like adding variables, but it's probably worth it for this.

minad:
> I'll add the variable. I don't like adding variables, but it's probably worth it for this.

Me too. But I think this sounds right. I think it may bring us a step closer to unifying the region actions too.

minad:
The embark--target-bounds variable is basically the means to inject boundaries into an action. We have been lacking this. It is not pretty, but there is probably no reasonable alternative.

If there is the "r" interactive specification, we could pass the region arguments directly. But I am not fond of complicating the Embark calling convention.

oantolin:
OK, I added the variable and fixed the sexp actions. 478ffcc4dd21def4a0a552134fb2b94abe8f5630

Now we just need to decide what the right level of generality is for those actions. I guess I could move them to the general map now. What do you think?

oantolin:
I mean, not raise, but indent, kill and mark could use the region functions and operate on any target.

minad:
> Now we just need to decide what the right level of generality is for those actions. I guess I could move them to the general map now. What do you think?

Right, now the indent/kill/mark are more general. But having it in the general map is not a good idea because it does not work in the minibuffer.

Now I would like to have some automatic keymap derivation mechanism again. Hmm. Maybe introduce an embark-bounds-map which contains these actions which require these target bounds. If the action keymap derives from the general map also merge in the embark-bounds-map automatically.

Then the question is if the embark-bounds-map could ultimately be the same as the embark-region-map.

minad:
Okay, it should be like I described above, but embark-region-map/=embark-bounds-map. But the region finder should also return the boundaries of the region, and the region-map should inherit from the new bounds-map, since the bounds-map is more general.

oantolin:
> Then the question is if the embark-bounds-map could ultimately be the same as the embark-region-map.

I was thinking that maybe if the action name ends in `-region`, `embark--act` could arrange for the target bounds to be set as as the region. I mean, it could also always set the region to the bounds, but people probably don't want Embark messing with the region all the time.

minad:
> I was thinking that maybe if the action name ends in -region, embark--act could arrange for the target bounds to be set as as the region. I mean, it could also always set the region to the bounds, but people probably don't want Embark messing with the region all the time.

I've also considered adjusting the region, but it does not seem like a good idea. I also don't like the magic of checking if the action name ends with `-region`.

We could also call the bounds map `embark-at-point-map`.

oantolin:
- `embark-at-point-map` sounds nicer.
- I don't think the region map should inherit from the at-point map: the region map contains unwrapped `-region` commands for which this new map would contain a wrapper.

oantolin:
You know, we could also just leave things as they are for a bit. I don't think I've ever really wanted to indent the region consisting of the file name at point. 😉 Maybe only kill would really be useful to generalize.

minad:
> I don't think the region map should inherit from the at-point map: the region map contains unwrapped -region commands for which this new map would contain a wrapper.

I agree. But it would work in principle and it would unify things. At the same time it wouldn't make much sense, since marking the region again is nonsense :laughing: 

> You know, we could also just leave things as they are for a bit. I don't think I've ever really wanted to indent the region consisting of the file name at point. wink Maybe only kill would really be useful to generalize.

Okay, I agree. Better than reintroducing the auto inheritance mess again. We will regret it later.

oantolin:
By the way, the commands in the region map that also prompt at the minibuffer, and which therefore wouldn't work with a string target, seem to be only three: `shell-command-on-region`, `align-regexp`, and `write-region`.

embark.el
```diff
@@ -466,55 +471,63 @@ There are three kinds:
...
-                    ((derived-mode-p 'dired-mode)
-                     (dired-get-filename t 'no-error-if-not-filep))
-                    ((derived-mode-p 'emacs-lisp-mode)
-                     (when (or (nth 3 (syntax-ppss)) ; in a string
-                               (nth 4 (syntax-ppss)) ; or comment
-                               (save-excursion
-                                 (unless (looking-at "\\_<")
-                                   (forward-symbol -1))
-                                 (forward-symbol -1)
-                                 (looking-at "use-package\\|require")))
```
oantolin, line 492: I'm a little conflicted about removing this guard in Emacs Lisp files, requiring `use-package` or `require`. On the one hand, now with multiple targets we can loosen the criteria to allow more matches, on the other hand it causes the following problem:

Consider a `(pcase ...)` expression or some occurrence of the `align` command. Those are both functions (well, a macro and a function) and libraries which ffap targets. I'd like the function case to have priority over the file case. But now consider a file name in a string or comment in an Elisp buffer. It's typically also a symbol, but there I want the file interpretation to have priority.

minad: My general sentiment is to give files a lower priority than symbols. I would be fine if symbol wins always maybe except in dired.

What about this? 

1. Introduce an high priority file finder which only triggers in dired an for names with a file extension or a slash
2. Introduce a low priority file finder which triggers on all files, except if the high priority finder triggered?
3. Put the symbol finder between these finders.

oantolin: That sounds reasonable. For repeated targets the `delete-dups` call you put in keeps the first occurrence, right?

minad: Yes, but I am now thinking about removing it again. The problem is that it may not be effective in this case, so it only covers up bugs. Could it happen that the high priority finder returns "embark.el" and the low priority finder returns "embark"?

oantolin: I think this works the way I want: 6682da8f497b05739e1c06e41f0e1a0cd1a406ec

oantolin: I'd be OK with removing the `delete-dups` call. The commit I made makes the high priority finder not return anything in the library-without-extension case, so it doesn't rely on duplicate removal.

oantolin: I'd vote for removing the `delete-dups` call on the grounds that it's probably better to tighten up the target finders than to rely on duplicates being removed.

oantolin: I removed `delete-dups`: 275f1eb28470f77cd4c5ac6b3128681794c62de2.

minad: I just tried the updated version. Seems to work great!

PR #281, opened 2021-07-25
Add embark-face-map

minad:
Since we are classifying faces explicitly in marginalia, we may want to have a special keymap too.


minad:
In Embark if we have a category more specific than symbol there is no way going back. So we would only get the general map for faces. Same problem as in #262.

PR #282, opened 2021-07-26
Verbose action indicator: initial implementation

jaor:
This is a mostly direct translation of the action indicator i have
been using in my personal configuration for some months now.  It's
more code than i wish it had, and it's very probable that it can be
simplified in many ways (i am hoping you'll see several of those way
almost at first glance!), but it can serve as a baseline.  (I hope
this is the one that was mentioned in reddit, i had another one, way
simpler and uglier, based on `describe-keymap`, that i ditched long
ago.)

I am very fond of meta-behaviours, but i am not sure what using an
embark collect buffer would buy us here, and just wanted to send your
way a first working implementation.  Happy to learn and modify things
accordingly, though.

Thanks!

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -915,7 +915,8 @@ Used by `embark-verbose-indicator'.")
   "Action indicator that displays a list of all available action key bindings."
   (with-current-buffer (get-buffer-create embark--vindicator-prompt-buffer)
     (read-only-mode -1)
-    (setq-local cursor-type nil)
+    (setq-local cursor-type nil
+                truncate-lines t)
     (delete-region (point-min) (point-max))
     (let* ((descs (embark--vindicator-keymap-descriptor keymap ""))
            (fmt (format "%%-%ds  %%-%ds   %%s\n" (cadr descs) (car descs))))
@@ -934,6 +935,7 @@ Used by `embark-verbose-indicator'.")
     (if target
         (insert (format "\nAction for %s '%s'" (car target) (cdr target)))
       (delete-char -1))
+    (goto-char (point-min))
     (read-only-mode 1)
     (let ((display-buffer-alist
            (append display-buffer-alist embark--vindicator-display-alist)))
--- a/embark.el
+++ b/embark.el
@@ -1,6 +1,6 @@
 ;;; embark.el --- Conveniently act on minibuffer completions   -*- lexical-binding: t; -*-
 
-;; Copyright (C) 2020  Omar Antolín Camarena
+;; Copyright (C) 2020, 2021  Omar Antolín Camarena
 
 ;; Author: Omar Antolín Camarena <omar@matem.unam.mx>
 ;; Keywords: convenience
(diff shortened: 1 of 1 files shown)


embark.el
```diff
@@ -847,6 +847,103 @@ If NO-DEFAULT is t, no default value is passed to `completing-read'."
...
+  "Parameters added to display-buffer-alist for displaying the actions buffer.")
+
+(defvar embark--vindicator-excluded-commands
+  '(embark-collect-snapshot embark-collect-live embark-export
+    embark-keymap-help embark-become embark-isearch nil)
+  "Commands not displayed by embark-verbose-indicator.")
+
+(defun embark--vindicator-key-str (k)
+  "A string describing the given key or key code."
+  (if (numberp k) (single-key-description k) (key-description k)))
```
minad, line 869: Why not use key-description?

minad:
@jaor Thank you! I am all for adding such an indicator to Embark. It will make Embark significantly easier to grasp at the beginning.

jaor:
On Mon, Jul 26 2021, Daniel Mendler wrote:

> Why not use key-description only?

Because key-description won't accept a numeric value as its argument,
and i guess i must have found cases where keymaps passed to the prompter
included them (i understood that's why single-key-description exists in
the first place... but my knowledge of how keymaps work is very limited.)

jaor:
On Mon, Jul 26 2021, Daniel Mendler wrote:

> +(defun embark--vindicator-keymap-descriptor (k prefix)
> +  "Generate a description of keymap K, recursively.
> +PREFIX is used in bindings descriptions strings, and, besides
> +that string for all seen bindings, DESCS includes the maximum
> +width of that string seen so far, for alignment purporses."
> +  (seq-reduce `(lambda (descs x) (embark--vindicator-bind-desc descs x ,prefix))
> +              (cdr (keymap-canonicalize k)) '(0 0)))
>
> Here one could also use map-keymap. Is there an advantage in using
> keymap-canonicalize and mapping?

I don't know... IIRC, i saw keymap-canonicalize being used in
describe-keymap and took it as a "recommended" procedure (i also liked
what its docstring says about making the returned value "simpler", and
how it resolves redefinitions).  I am also guessing that since i was
reducing rather than mapping, i just filtered out map-keymap.

oantolin:
Thanks a lot @jaor!

oantolin:
It could do with a couple of cosmetic tweaks: `(setq truncate-lines t)` (I'm know I'm a weirdo, what with my non-maximized Emacs), `(goto-char (point-min))` (because otherwise you only ever see the few last actions ---which thanks to the sorting are the least interesting ones, the ones that start with `embark`--- and a large blank space at the bottom).

Before those tweaks: 
image

With the tweaks: 
image

oantolin:
Also, maybe we can skip the `read-only-mode`, since you can't really move to that window and edit it anyway.

oantolin:
For users of the embark-consult package, it would be nice if commands that start with consult should also be sorted lower (but probably above the embark commands).

jaor:
On Mon, Jul 26 2021, Omar Antolín Camarena wrote:

> It could do with a couple of cosmetic tweaks

indeed: a12715e

> (setq truncate-lines t) (I'm know I'm a weirdo, what with my
> non-maximized Emacs)

you should try exwm one these days :)

oantolin:
Is alphabetical sorting really a good idea? It messes up the inheritance order of the keymaps, putting the general commands higher than they should be. I feel that pushing commands that start with "embark" back to the bottom is a workaround only necessary because of the sorting in the first place.

oantolin:
> you should try exwm one these days

I think I don't use a supported OS: I use Windows and Chrome OS. On Chrome OS I run emacs in a Linux VM so I might be able to manage at least the Linux applications with exwm, but the only GUI Linux application I use _is_ Emacs!

oantolin:
Ignore my comments about sorting, I think I managed to confuse myself.

jaor:
On Mon, Jul 26 2021, Omar Antolín Camarena wrote:

> Is alphabetical sorting really a good idea? It messes up the
> inheritance order of the keymaps, putting the general commands higher
> than they should be. I feel that pushing commands that start with
> "embark" back to the bottom is a workaround only necessary because of
> the sorting in the first place.

hmm.  when i remove sorting, i see commands in an order that i don't
fully understand.  usually, all consult commands go to the bottom, but
embark's are interspersed with others.  since command names tend to use
common prefixes, sorting by command name makes things easier to spot, at
least for me, and it feels natural (again, for me, not a golden standard
of naturalness!).

oantolin:
With those visual tweaks, this looks ready to me! Thanks again, @jaor!

oantolin:
I think a lot of the formatting can be shared between this and the completing-read prompter. I'll try to refactor later.

oantolin:
So, should this be the default indicator?

minad:
> So, should this be the default indicator?

Yes, and maybe also make the minimal indicator a normal function then. Such that Embark provides by default the embark-verbose-indicator and the embark-minimal-indicator.

minad:
Does the verbose indicator also show the other available targets at point? I miss this on the screenshot above.

oantolin:
> Does the verbose indicator also show the other available targets at point? I miss this on the screenshot above.

Oh, it doesn't (yet)!

jaor:
On Mon, Jul 26 2021, Omar Antolín Camarena wrote:

> With those visual tweaks, this looks ready to me! Thanks again, @jaor!

nah, thanks to you: i'm quite happy to have a couple lines of code in
one of my favourite projects.

jaor:
On Mon, Jul 26 2021, Omar Antolín Camarena wrote:

>     Does the verbose indicator also show the other available targets
>     at point? I miss this on the screenshot above.
>    
> Oh, it doesn't (yet)!

i initially thought it could be too crowded, but maybe not.  in fact, i
was playing with the idea of providing some customization to the way the
indicator at the bottom is displayed, in the same way that one can now
customize the indicator itself with strings, but again couldn't make my
mind up.  and, somewhat related to that, providing some kind of
customization for how each of the bindings is shown (possibly via a
special format string) also crossed my mind; but again again, i am not
sure if the added complexity is really worth it.

oantolin:
I don't think any customizaton is necessary. And I'll add the other targets hint, just as in @minad's default value of the indicator.

oantolin:
> Yes, and maybe also make the minimal indicator a normal function then.

I am tempted to allow only functions. And maybe just add the which-key indicator from the wiki (improved to show the target as [in a recent issue](https://github.com/oantolin/embark/issues/269#issuecomment-883716033)) to Embark itself.

minad:
> And maybe just add the which-key indicator from the wiki (improved to show the target as in a recent issue) to Embark itself.

Please no. Let's not integrate packages dynamically except the completion systems. Otherwise it is unclear where to stop. Add package specific target finders, add package specific actions, etc? You probably remember the old discussion with Steve Purcell.

oantolin:
> Let's not integrate packages dynamically except the completion systems. Otherwise it is unclear where to stop. 

OK, good point.

jaor:
On Mon, Jul 26 2021, Omar Antolín Camarena wrote:

> (goto-char (point-min)) (because otherwise you only ever see the few last actions
> ---which thanks to the sorting are the least interesting ones, the
> ones that start with embark--- and a large blank space at the bottom).

A comment on this: in my daily use, it's knowing the actual target is
pretty important (and it's become more so with multi targets), so it
might be better to display it at the beginning rather than at the end if
you keep `(goto-char (point-min))` (actually, if this were one of my
packages, i couldn't help making the position customizable--i see valid
reasons for both placements depending on the use case).

bdarcus:
This looks cool!

How about a section in the README, with a how to config this, and a screenshot with a nice theme?

I just noticed you don't have any screenshots here; one or two would be helpful I think.

oantolin:
Adding this to the README is still on the TODO list, @bdarcus. Also on the TODO list is to update the which-key prompter to the new convention.

Issue #172, opened 2021-02-26
Add support for group-function


minad:
I don't have a patch for Embark yet. You mentioned that collapsing would be interesting to have, is there some existing facility to be used for this, provided by tablist? Or was the idea to implement this by hand using 'invisible and creating a separate button?

oantolin:
I don't tabulated-list-mode has anything to be used for this. Maybe tablist does (I think it probably doesn't either, but am not sure), but I don't want to depend on it. My plan was to use `invisible` and a button, as you suggest.

minad:
Duplicate #389

PR #283, opened 2021-07-27
Various vindicator improvements

minad:
I just tried the new vindicator. It is a great addition and will be a good default, but we should polish it a little more. While playing around, I directly made some improvements (in my opinion):

- Truncate the target, such that it fits on a single line. This is important given that we have the defun target finder now. And more large target finders may come. (new function embark--truncate-target)
- Make the target the headline of the window
- Show shadowed targets below headline
- Support regexps in the exclude list
- Keybindings before command name for consistency with embark-bindings, reuse embark--formatted-bindings
- Reuse a window, this is a better fit for the verbose vertically arranged candidate list
- Do not hide the mode line, it is more instructive
- Remove the Selectrum-specific code
- Do not use read-only-mode, use inhibit-read-only

What do you think? cc @jaor

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -865,9 +865,10 @@ If NO-DEFAULT is t, no default value is passed to `completing-read'."
 (defvar embark--vindicator-prompt-buffer " *Embark Actions*"
   "Buffer used by `embark-verbose-indicator' to display actions and keybidings.")
 
-(defvar embark--vindicator-display-alist
-  `((,(regexp-quote embark--vindicator-prompt-buffer)
-     (display-buffer-reuse-window)))
+(defvar embark--vindicator-display-action
+  '(display-buffer-reuse-window)
+  ;;'(display-buffer-in-side-window (side . right))
+  ;;'(display-buffer-below-selected (window-height . 15))
   "Parameters added to `display-buffer-alist' for displaying the actions buffer.")
 
 (defvar embark--vindicator-excluded-commands
@@ -962,7 +963,9 @@ OTHER-TARGETS are other shadowed targets."
                                'string-greaterp (cddr descs))))
       (goto-char (point-min))
       (let ((display-buffer-alist
-             (append display-buffer-alist embark--vindicator-display-alist)))
+             (append display-buffer-alist
+                     `((,(regexp-quote embark--vindicator-prompt-buffer)
+                        ,@embark--vindicator-display-action)))))
         (pop-to-buffer (current-buffer) nil t)))
     (lambda ()
       (embark-kill-buffer-and-window embark--vindicator-prompt-buffer)
--- a/embark.el
+++ b/embark.el
@@ -215,18 +215,6 @@ Used by `embark-completing-read-prompter' and `embark-keymap-help'.")
 (defface embark-target '((t :inherit highlight))
   "Face used to highlight the target at point during `embark-act'.")
 
-(defface embark-verbose-indicator-documentation '((t :inherit completions-annotations))
-  "Face used by the verbose action indicator to display binding descriptions.
-Used by `embark-verbose-indicator'.")
-
-(defface embark-verbose-indicator-title '((t :height 1.1 :weight bold))
-  "Face used by the verbose action indicator for the title.
-Used by `embark-verbose-indicator'.")
-
-(defface embark-verbose-indicator-shadowed '((t :inherit shadow))
-  "Face used by the verbose action indicator for the shadowed targets.
-Used by `embark-verbose-indicator'.")
-
 (defcustom embark-action-indicator
   (let ((act (propertize "Act" 'face 'highlight)))
     (cons act (concat act " on %2$s"
@@ -779,6 +767,8 @@ first line of the documentation string; otherwise use the word
     (t "<unnamed>"))))
 
 (defun embark--formatted-bindings (keymap)
+  "Return the formatted keybinding of KEYMAP.
+The keybindings are returned in their order of appearance."
   (let* ((commands
           (cl-loop for (key . cmd) in (embark--all-bindings keymap)
                    for name = (embark--command-name cmd)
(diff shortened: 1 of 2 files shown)


embark.el
```diff
@@ -670,6 +678,14 @@ If CYCLE is non-nil bind `embark-cycle'."
...
                      (alist-get t embark-keymap-alist)))))
 
+(defun embark--truncate-target (target)
+  "Truncate TARGET string."
+  (unless (stringp target)
+    (setq target (format "%s" target)))
+  (if-let (pos (string-match-p "\n" target))
+      (concat (substring target 0 pos) "…")
+    target))
+
```
jaor, line 684: this is a good idea, yes.  for cases where emacs runs maximased
(for instance, i have 270 columns at my disposal) one could consider
joining lines (e.g., to show a full sexp) up until (window-width), but
i am not sure about the result.  this simpler approach is probably better.

embark.el
```diff
@@ -849,35 +862,40 @@ If NO-DEFAULT is t, no default value is passed to `completing-read'."
...
              (cons (max (or (car cds) 0) (or (car descs) 0))
                    (cons (max (or (cadr cds) 0) (or (cadr descs) 0))
                          (append (cddr descs) (cddr cds))))))
-          ((memq c embark--vindicator-excluded-commands) descs)
+          ((seq-find (lambda (x)
+                       (if (symbolp x)
+                           (eq c x)
+                         (string-match-p x (symbol-name c))))
+                     embark--vindicator-excluded-commands)
+           descs)
```
jaor: i would perhaps define an embark--vindicator-excluded-command-p predicate
to make this code obvious, but it's just a very minor thing.

jaor (changes requested):
this all looks pretty good to me.  i've added a couple of very minor nits
below.  also, i agree that showing keys first is more consistent, so i
won't complain even though i prefer my original ordering (my usual
solution here would be to make order customizable, but @oantolin said
he prefers not to).

embark.el
minad: Yes, for now I've avoided using a more complicated truncation method. Truncation is a very recent addition so we I am certain we will improve this further at some point.

embark.el
minad: Good idea, added a helper function.

jaor:
hmm, i just noticed you made by default the actions window being a
side-window.  i dislike it, and find it kind of inconsistent in a
vertico/consult environment, where i'm very used to look at the
minibuffer area for prompts (it made me think of helm :).  i would
make the default a window on top of the minibuffer.  but of course
it's trivial to configure it that way, so it's not very important.

minad:
> hmm, i just noticed you made by default the actions window a
side-window. i dislike it, and find it kind of inconsistent in a
vertico/consult environment, where i'm very used to look at the
minibuffer area for prompts (it made me think of helm :). i would
make the default a window on top of the minibuffer. but of course
it's trivial to configure it that way, so it's not very important.

Yes, it is trivial to configure. It is important that we document the different possibilities. I felt that the embark-verbose-indicator does not work particularly well in a small window at the bottom, therefore I changed it. The motivation for the change was also that the targets were not truncated before, such that the window at the bottom resized when invoking embark-cycle.

minad:
> (it made me think of helm :)

@jaor Embark+Orderless is the new Helm ;)

@oantolin The only thing missing is #172 :-P

jaor:
On Mon, Jul 26 2021, Daniel Mendler wrote:

>     hmm, i just noticed you made by default the actions window a
>     side-window. i dislike it, and find it kind of inconsistent in a
>     vertico/consult environment, where i'm very used to look at the
>     minibuffer area for prompts (it made me think of helm :). i would
>     make the default a window on top of the minibuffer. but of course
>     it's trivial to configure it that way, so it's not very important.
>    
> Yes, it is trivial to configure. It is important that we document the
> different possibilities. I felt that the embark-verbose-indicator does
> not work particularly well in a small window at the bottom, therefore
> I changed it.

conversely, a window like that doesn't work well in wide frames with
several windows in columns :) in those situations, closing the pop-up
window is prone to not recover the previous window configuration (e.g.,
i have three columns, one window each, the pop up appears in the first
one and on closing i end up with two columns, one double width of the
other; or i have a layout with 5 windows, and where the actions one will
appear is anybody's guess)... so, an advantage of popping up at the
bottom and frame-wide is that it's deterministic and always respects
complex window layouts in the frame.

> The motivation for the change was also that the targets were not
> truncated before, such that the window at the bottom resized when
> invoking embark-cycle.

you mean resized vertically, right? (a bottom window on top of the
minibuffer is always the frame's width).  i see how that can be
unconvenient.

anyway, as you say, the important thing is that we document well how to
set up things, because here clearly there's no "one size fits all"
scenario.

jaor:
On Mon, Jul 26 2021, Daniel Mendler wrote:

>     (it made me think of helm :)
>    
> @jaor Embark+Orderless is the new Helm ;)

not just new, also better! (said as someone that has never used helm
in earnest) :)

minad:
You are right, it really depends on the scenario. For me the main annoyance was that the window at the bottom resizes vertically when invoking `embark-cycle`. I agree that the bottom is very deterministic, it behaves similarly as a minibuffer and will give familiar  looks for Vertico/Selectrum users.

oantolin:
This looks pretty good @minad, except for the display function which tends to pick a skinny window so you can't see the descriptions and then annoyingly changes your window configuration every single time! I believe this will induce rage in users and cause them to destroy their computers and I really don't want to take the blame for that. 😛 Having it pop-up at the bottom is much less intrusive.

oantolin:
I really like the header line, very snazzy. People will see that and immediately know it wasn't me. 😛

oantolin:
Oh, now that I read the previous comments (sorry, I was anxious to try the code first), I realize @jaor had already made the point about the display function, although maybe not as graphically.

oantolin:
You know, with the change of order in the key binding and command name it really does look remarkably like the completing-read prompter, here are both together:
image

oantolin:
Thanks @minad! For all the changes (well... except the display function which I think was probably better in @jaor's version), but particularly for unifying the formatting code between the vindicator and the completing-read prompter.

minad:
Okay, I give in regarding the display function. Just trying to get the inner Helm out of Embark ;)

oantolin:
@jaor I wouldn't mind adding a configuration option for the order between the key bindings and the command names, now it can apply uniformly to the verbose indicator and to the completing-read prompter. I didn't think hard about the order, I just followed what restaurant menus do: first the item number (which is like the key binding because you can use it to specify your order), then comes the dish's name, finally the description.

oantolin:
> Okay, I give in regarding the display function. Just trying to get the inner Helm out of Embark ;)

To be honest, I don't mind the tall skinny windows, the only thing that really bothered me is that it kept changing my window configuration. Maybe if we try to always restore the layout this side window display would be pretty nice too.

minad:
> To be honest, I don't mind the tall skinny windows, the only thing that really bothered me is that it kept changing my window configuration. Maybe if we try to always restore the layout this side window display would be pretty nice too.

I don't mind. It is okay if we point out that it is configurable and document the different setttings.

PR #285, opened 2021-07-27
Indicator corrections and simplifications

minad:
I tried https://github.com/oantolin/embark/commit/2b9053ffe26c4441cc12b28c865e66bf09382a9c, fixes the two missing variables. Since we have a new variable embark-indicator we can simplify the indicator calling convention again.


oantolin:
Nice, thanks!

PR #287, opened 2021-07-27
Add submenu support to verbose indicator

minad:
~Now we can add a prefix argument to the action indicator or better enhance the indicator to return an update callback. The update callback could take a single argument (prefix key), if the argument is nil destroy the indicator, if the argument is non-nil update the indicator with the given prefix key.~

The indicator returns an update function.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -722,16 +722,33 @@ Display a message in the minibuffer prompt or echo area showing the TARGETS."
         (let ((indicator-overlay
                (make-overlay (point-min) (point-min) (current-buffer) t t)))
           (overlay-put indicator-overlay 'before-string (concat indicator " "))
-          (lambda () (delete-overlay indicator-overlay)))
+          (lambda (_) (delete-overlay indicator-overlay)))
       (message "%s" indicator)
       nil)))
 
-(defun embark-keymap-prompter (keymap)
+(defun embark--read-key-sequence (update)
+  "Read key sequence, call UPDATE function with prefix keys."
+  (let ((timer) (prefix))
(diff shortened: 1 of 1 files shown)


minad:
@oantolin I think I am done here

embark.el
```diff
@@ -919,27 +937,11 @@ TARGETS is the list of targets."
...
-          (format "Shadowed targets at point: %s (%s to cycle)\n"
-                  (string-join
-                   (mapcar (lambda (x)
-                             (symbol-name (car x)))
-                           other-targets)
-                   ", ")
-                  (key-description
-                   (car (where-is-internal #'embark-cycle keymap))))
-          'face 'embark-verbose-indicator-shadowed)))
+      (when (and other-targets (where-is-internal #'embark-cycle keymap))
```
oantolin, line 943: Why do you test for the presence of `embark-cycle` too?

minad: embark-cycle is not present in the sub keymap anymore. At this point you cannot switch anymore, so it would be confusing to still show this.

oantolin:
This is great! Thanks again.

embark.el
minad: We could also try to support embark-cycle in subkeymaps though, such that cycling would still be possible. But then we could weird behavior, you would end up at the top keymap again.

oantolin: Oh! I see. Let's leave it like this, you enter a prefix keymap at your own risk.

minad: Exactly, and you can always abort.

Issue #284, opened 2021-07-27
Create an embark-minimal-indicator


minad:
...and remove support for strings in embark-action-indicator. Then we should really consider the snazzy verbose indicator as default ;)

And to not forget, add a few screenshots to the README, featuring the modus themes of course.

oantolin:
I wrote the embark-minimal-indicator, and decided to unify the action and become indicators (which required a small update to the verbose indicator). 2b9053ffe26c4441cc12b28c865e66bf09382a9c

There are still several things to do:

- Rewrite the which-key indicator in the wiki to work for both acting and becoming.
- Add some documentation to the README (but a lot can simply be taken from the docstrings, I think).
- Take some screenshots!

oantolin:
I didn't explicitly mention it in that last comment, but I did remove the string case of the indicator, they need to be functions now.

oantolin:
Oh, and best of all: I made the verbose indicator the default! 🎉

minad:
Great news! Thank you!

This may be controversial: I wonder do we still need embark become? I think I've not used it a single time. Or I've not understood where I would use it.

My main confusion with the command is that it has its own keymaps. Is embark-become not just the same as embark-act such that you always keep the input and such that you don't press RET?
It would seem more embarkier to me if embark-become just offers the same keymap as the actions, meaning "become one of the actions". 

The the verbose action indicator could be more helpful too. Press B x to become x.

minad:
Oh and this issue can be closed! :tada:

oantolin:
One item on the TODO list is done: I've [rewritten the which-key indicator](https://github.com/oantolin/embark/wiki/Additional-Configuration#use-which-key-like-a-key-menu-prompt) to conform to the new indicator protocol and thanks to @minad's tireless efforts it now updates for submenus!

PR #288, opened 2021-07-27
Fix completing-read prompter

minad:
There is a bug when switching to the completing-read prompter due to #287, fixed here.

I've also observed that when you press C-. C-h on a target at point the indicator will be shown in the long form before the minibuffer prompt "Act on ...". Did I introduce this or is this a regression due to #284?

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -740,7 +740,8 @@ Display a message in the minibuffer prompt or echo area showing the TARGETS."
                                            (when (/= (length prefix) 0)
                                              (funcall update prefix))))))))
           (read-key-sequence nil 'echo nil t 'cmd-loop))
-      (cancel-timer timer))))
+      (when timer
+        (cancel-timer timer)))))
 
 (defun embark-keymap-prompter (keymap update)
   "Let the user choose an action using the bindings in KEYMAP.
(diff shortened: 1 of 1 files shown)


embark.el
```diff
@@ -751,23 +752,21 @@ UPDATE is the indicator update function."
                 (embark--read-key-sequence update)))
          (cmd (let ((overriding-terminal-local-map keymap))
                 (key-binding key))))
-    (setq cmd
```
oantolin, line 754: This was silly, wasn't it? I think there's probably a historical reason it evolved to this, but I don't remember.

minad: Embark evolved a lot

oantolin:
> I've also observed that when you press C-. C-h on a target at point the indicator will be shown in the long form before the minibuffer prompt "Act on ...".

I'm not sure I follow. Are you talking about how the minimal indicator will always use the long form "Act on type(others) 'target'" if you are in the completing-read prompter? I did that on purpose, because the completing-read prompter clobbers the message that was in the echo area and if I didn't use the long form you wouldn't see what the target was or if there are other targets.

oantolin:
Also, I seem to recall that was the behavior previously too. I seem to recall it was requested by someone who used the completing-read prompter by default.

minad:
Hmm, this makes the prompt very long. Shall we go back to the short form now that we have target highlighting? We could make the short form slightly longer, writing "Act defun(...)" and show the ellipsis only if there are more targets. Maybe we can also make this configurable? I haven't checked how the configuration works currently for the minimal indicator.

oantolin:
How about just adding a new line between the indicator string and the minibuffer prompt:
image

oantolin:
I am a little reluctant to remove the target just because, as shown in that screenshot, there is no reason for it to coincide with the highlighted portion and its informative to know what the target really is.

minad:
Somehow target highlighting is broken now with the verbose-indicator. But not with the minimal indicator. Do you observe this too?

minad:
Weird, there is some buffer switching going on.

oantolin:
I hadn't noticed. With the verbose indicator, initial highlighting is fine, but when you cycle it stop higlighting. Maybe a simple `save-window-excursion` is enough to fix this. I'll check.

minad:
See my fix here #289

oantolin:
I think that's a sensible fix for the issue you noticed about the length of the prompt: if the indicator is longer than 40% of the frame width it goes on its own line. How does that sound?

PR #289, opened 2021-07-27
Fix highlighting when cycling

minad:


oantolin:
Isn't `(save-excursion (pop-to-buffer ...))` pretty much equivalent to `display-buffer`?

oantolin:
Replacing `pop-to-buffer` with `display-buffer` also fixes the highlighting. I'm going with that instead.

minad:
> Replacing pop-to-buffer with display-buffer also fixes the highlighting. I'm going with that instead.

Good, I am not very familiar with these pop-to-buffer functions. pop-to-buffer was used by @jaor, but there is no point in switching to that buffer.

The Consult buffer preview function should also be upgrade probably.

~~~ elisp
(defvar consult--buffer-display #'switch-to-buffer
  "Buffer display function.")

...

(defun consult--buffer-preview ()
  "Buffer preview function."
  (let ((orig-buf (current-buffer)))
    (lambda (cand restore)
      (when (and (not restore)
                 ;; Only preview in current window and other window.
                 ;; Preview in frames and tabs is not possible since these don't get cleaned up.
                 (or (eq consult--buffer-display #'switch-to-buffer)
                     (eq consult--buffer-display #'switch-to-buffer-other-window)))
        (cond
         ((and cand (get-buffer cand)) (consult--buffer-action cand 'norecord))
         ((buffer-live-p orig-buf) (consult--buffer-action orig-buf 'norecord)))))))

...

(defun consult--buffer-action (buffer &optional norecord)
  "Switch to BUFFER via `consult--buffer-display' function.
If NORECORD is non-nil, do not record the buffer switch in the buffer list."
  (funcall consult--buffer-display buffer norecord))
~~~

Does it make sense to also replace this with pop-to-buffer somehow?

oantolin:
Forgot to mention this PR in the commit message: 6b7b7f2fd4de682ca7f917abb646cda732af0551

oantolin:
Oh, yes, probably the consult buffer previewing would be less disruptive with `display-buffer`! I bet the `embark-insert` action would work correctly even during previews with that change.

oantolin:
You can probably use `display-buffer` with the `reuse-window` action. (And on second thought, maybe this wouldn't help the `embark-insert` action.) I still get very confused by these buffer displaying commands.

PR #292, opened 2021-07-27
Add embark-target-mode

minad:
What do you think? Should we add something like this or is this rather a candidate for the wiki?


minad:
I say it is for the wiki, added here: https://github.com/oantolin/embark/wiki/Additional-Configuration#show-the-current-embark-target-types-in-the-modeline

PR #297, opened 2021-07-28
Customizable structure for verbose indicator buffer

jaor:
We introduce the customizable list
embark-verbose-indicator-buffer-sections, which denotes the sections
to be included in the indicator buffer, in the user's desired order.
Besides symbols denoting specific sections (target, other-targets,
etc.), we allow the symbol newline and literal strings, which are
inserted, well, literally.

For each of the elements in the list there's a corresponding insertion
function, which users can advice to provide even more control on how
the sections are printed (we could define customization variables for
each inserter, but that seems to be, perhaps, too much).

The default value of embark-verbose-indicator-buffer-sections
reproduces the current structure.

As another example, i prefer the title to be below the
bindings (because i use bottom placement), put together the current
target and the alternative ones in the same line (i've got plenty of
horizontal space) and omit the cycle key indicator (i don't need that
reminder), so i've got

   '(bindings newline target "  " other-targets)

as the value of embark-verbose-indicator-buffer-sections.

Yes, newline is redundant (one can use the literal string "\n" with
the same effect), but it looks nicer in the customization buffer as a
symbolic choice.

An alternative that might be worth considering is using formatting
functions returning strings instead of inserters which write in the
buffer, specially if one wants to make them customizable (although a
bit of flexibility is lost that way if not).

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -937,6 +937,21 @@ display actions and parameters are available."
   "Commands not displayed by `embark-verbose-indicator'."
   :type '(repeat (choice regexp symbol)))
 
+(defvar embark--verbose-indicator-shadow-str
+  (propertize "Shadowed targets at point: "
+              'face 'embark-verbose-indicator-shadowed))
+
+(defcustom embark-verbose-indicator-buffer-sections
+  `(target newline ,embark--verbose-indicator-shadow-str other-targets
+    " " cycle newline newline bindings)
+  "List of sections to display in the verbose indicator buffer, in order."
+  :type '(repeat (choice (const :tag "Current target name" target)
+                         (const :tag "List of other targets" other-targets)
+                         (const :tag "Key bindings" bindings)
+                         (const :tag "Cycle indicator" cycle)
+                         (const :tag "New line" newline)
+                         (string :tag "Literal string"))))
+
 (defvar embark--verbose-indicator-buffer " *Embark Actions*"
   "Buffer used by `embark-verbose-indicator' to display actions and keybidings.")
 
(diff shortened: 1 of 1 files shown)


jaor:
OK, i've made the thing fully general by accepting an arbitrary
function (taking the targets, bindings and cycling key) as a possible
element of embark-verbose-indicator-buffer-sections.  That way neither
further customization variables are needed, nor advising, and there's
no limitation on what a user can display in the buffer.

embark.el
```diff
@@ -942,60 +961,80 @@ display actions and parameters are available."
...
+               (case section
+                 (newline (insert "\n"))
+                 (target
+                  (embark--verbose-indicator-insert-target (car target) (cdr target)))
+                 (other-targets
+                  (embark--verbose-indicator-insert-other-targets targets))
+                 (cycle (embark--verbose-indicator-insert-cycle-key ck))
+                 (bindings (embark--verbose-indicator-insert-bindings bindings))
+                 (t (when (fboundp section)
+                      (funcall section (cons target targets) bindings ck)))))))
```
minad, line 1027: Good idea! I suggest to not use an explicit pcase here and dispatch only via funcall for conciseness:

`(funcall (intern (format "embark--verbose-indicator-insert-%s" section)) all the args...)`

embark.el
```diff
@@ -942,60 +961,80 @@ display actions and parameters are available."
                 (string-match-p x (symbol-name cmd))))
             embark-verbose-indicator-excluded-commands))
 
-(defun embark--verbose-indicator-update (keymap target other-targets)
+(defun embark--verbose-indicator-insert-target (kind target)
```
minad, line 964: Instead of inserters use pure functions retuning strings as you already suggested? Seems cleaner to me.

jaor:
On Tue, Jul 27 2021, Daniel Mendler wrote:

> Good idea! I suggest to not use an explicit pcase here and dispatch
> only via funcall for conciseness:
>
> (funcall (intern (format "embark--verbose-indicator-insert-%s" section)) all the args...)

one would still need to pcase for the accepted symbols, or a random
symbol would result in a call to a non-existent function.  one would
also need to make all functions accept arguments they don't use, which i
don't specially like either.  and, in general, i don't really like
tricks like the above, feels too much as using eval outside of a macro.

all that said, if you really prefer it, it's a breeze to change :)

jaor:
On Tue, Jul 27 2021, Daniel Mendler wrote:

> Instead of inserters use pure functions retuning strings as you
> already suggested? Seems cleaner to me.

yes, indeed.  specially with the addition of user-defined functions
which can more easily reuse existing formatters if they're pure.  i'll
refactor them later today, possibly with the "intern" trick if you're
keen on it.

oantolin:
I think the pure functions returning strings are a good idea, and since you plan on switching to them, I'll wait for that before merging. I don't think you should use the `intern` trick, because of the differing function arguments.

minad:
No the idea is to use the same arguments for all functions. This way you can use the intern trick. I prefer the intern trick since it will error on invalid symbols. This is helpful to detect misconfiguration. But of course it is up to Omar to decide on the style.

oantolin:
I find it unnatural to use the same arguments for all the functions, since the default functions will each only use one.

jaor:
On Wed, Jul 28 2021, Omar Antolín Camarena wrote:

> I think the pure functions returning strings are a good idea, and
> since you plan on switching to them, I'll wait for that before
> merging.

done, please take a look when you can. thanks!

minad:
@oantolin Why is this unnatural? I say using different arguments is unnatural since then you have the formatter somehow scattered over two places. The case which decides on what to use and the formatter which uses the argument. Why even have these separate functions then?

oantolin:
Sorry @minad! You were right, I wasn't grasping the full context! The code already accepts as a section a function of all the arguments returning a string. Your suggestion would make all these helper functions valid section-functions and then make the symbols merely abbreviations of the section-function. Know that I understand your point I do agree with it.

We could even remove the symbols, which would then merely abbreviate the functions and have the function names appear directly in the configuration. It might be nice to keep `other-targets` rather than `embark-verbose-format-other-targets`, just for legibility, though.

In the current form, with the differing arguments, they could be inlined, as you said @minad.

jaor:
On Wed, Jul 28 2021, Omar Antolín Camarena wrote:

> We could even remove the symbols, which would then merely abbreviate
> the functions and have the function names appear directly in the
> configuration. It might be nice to keep other-targets rather than
> embark-verbose-format-other-targets, just for legibility, though.

emphatic +1 to keep it (and the same for all the others, specially for
those of us that almost never use the customize interface).

minad:
Btw, @oantolin https://github.com/hlissner/doom-emacs/commit/c2e6db27d030f20d3aead4b2049695e4b2d942ba

minad:
Brace for impact...

oantolin:
> Brace for impact...

Oh, no. I feel for you! Well, and maybe for me. Does this vertico module also configure orderless and maybe even embark?

minad:
Yes, you are all in. Embark and Orderless are mainstream now.

oantolin:
> Yes, you are all in. Embark and Orderless are mainstream now.

Crap. I mean, it's great, but surely this will lead to tons of bug reports. Brace for impact, indeed!

minad:
We should add some Doom detection code which either disables, self-destructs or at least ensures that the user cannot browse https://github.com/minad and https://github.com/oantolin with eww.

oantolin:
"This has been a long time coming. Vertico et co turned out to be what I wanted ivy to be and more." This is very high praise coming from Hernik.

minad:
Yes, indeed!

oantolin:
I think I'll merge this as is, and think a bit more about switching to uniform function signatures. And don't worry, @jaor, we'll keep the short symbols.

jaor:
On Wed, Jul 28 2021, Omar Antolín Camarena wrote:

> I think I'll merge this as is, and think a bit more about switching to
> uniform function signatures. And don't worry, @jaor, we'll keep the
> short symbols.

many thanks, on both accounts.  (fwiw, i am kind of agnostic on the
function signatures issue, i just mildly dislike interning formats as a
proxy for proper generics :) but, and at any rate, it's of course your
call!)

oantolin:
No, thank you for doing this, @jaor!

PR #298, opened 2021-07-28
Add embark-mixed-indicator with delayed display of the vindicator

### LINKED ISSUE #296: Optional delay for the verbose indicator display?
As the title suggests, this is about making the verbose indicator more adaptable for those who normally know what action they will take.  An optional delay, which could be set to 0 by default, would help with a workflow where if, say, I wait for 1 second then I must have forgotten what key/action I want, so the detailed buffer comes to offer help.

At any rate, thanks for the new features!

minad:
Fixes #296. For full configurability combine this with the flexible display by @jaor in #297!

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -215,7 +215,7 @@ Used by `embark-completing-read-prompter' and `embark-keymap-help'.")
 (defface embark-target '((t :inherit highlight))
   "Face used to highlight the target at point during `embark-act'.")
 
-(defcustom embark-indicator #'embark-verbose-indicator
+(defcustom embark-indicator #'embark-mixed-indicator
   "Indicator function to use when acting or becoming.
 The indicator function is called from both `embark-act' and from
 `embark-become' and should display information about this to the
@@ -237,6 +237,10 @@ Embark comes with two such indicators:
   detailed information including key bindings and the first line
   of the docstring of the commands they run.
 
+- `embark-mixed-indicator', which combines the minimal and the
+  verbose indicator. The verbose popup is shown after
+  `embark-mixed-indicator-delay' seconds.
+
 The calling convention for indicator functions is as follows:
 
 When called from `embark-act', the indicator function will be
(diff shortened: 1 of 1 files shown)


embark.el
```diff
@@ -237,6 +237,10 @@ Embark comes with two such indicators:
   detailed information including key bindings and the first line
   of the docstring of the commands they run.
 
+- `embark-mixed-indicator', which combines the minimal and the
+  verbose indicator. The verbose popup is shown after
+  `embark-mixed-indicator-delay' seconds.
```
protesilaos, line 242: Maybe the sentence starting with "The verbose popup..." needs to be on its own line?

Also, I think the compiler complains about a single space after a period.

oantolin: I'll take care of those after merging, @protesilaos.

oantolin:
Should the mixed one really be the default?

protesilaos:
> Should the mixed one really be the default?

Seeing as this will benefit new and potentially unsuspecting users the most, I would go with the verbose.

oantolin:
That was my thinking too, @protesilaos.

oantolin:
Although, if your problem is that you are a new user and don't know what actions are available you'll probably hesitate and then the mixed indicator will pop up anyway.

protesilaos:
Yes, indeed.

protesilaos:
By the way:

```elisp
(defcustom embark-mixed-indicator-delay 0.5
  "Time in seconds after which the verbose indicator is shown."
  :type '(choice (const :tag "No delay" 0)
                 (number :tag "Delay in seconds")))
```

Is this intended only for the mixed style?  Or would it apply to the verbose as well?

oantolin:
Only for the mixed indicator. The verbose indicator has no delay. I like it that way because I find the mixed indicator strictly superior to a delayed verbose one: before the delay is over, the mixed one shows you the minimal indicator which does provide very useful information: how many other targets there are at point.

protesilaos:
I see.  Thanks!

PR #305, opened 2021-07-29
Simplify setup hooks and verbose indicator insertion

minad:


oantolin:
Funny, I very nearly made the embark-setup-hooks change myself yesterday! The fact that you want it too means I shouldn't have hesitated. 

Some people do use the Embark-setup-hook, so it probably requires an obsoletion message.

oantolin:
Did I read the verbose inserted code correctly? Did you decide to get rid of the function case? Why?

minad:
Oh, you can write custom functions with the prefix `embark-verbose-indicator-section-`. Alternatively we could check if the symbol exists, then call it, otherwise call the prefixed name. Would you prefer that?

oantolin:
> Oh, you can write custom functions with the prefix `embark-verbose-indicator-section-`

Yeah, I realized that a few minutes after posting my comment. It's probably good enough, but I see one advantage to checking if the symbol is `fboundp` first: these custom section function will most likely be a part of people's personal configuration and many people have a habit of prefixing all their personal code with some personal prefix, I wouldn't want to _force_ people to use a function name without their prefix.

By the way, I really like keyword arguments for these functions. I can't believe I didn't think of it when we were discussing whether to give all inserter functions the same signature...

minad:
> By the way, I really like keyword arguments for these functions. I can't believe I didn't think of it when we were discussing whether to give all inserter functions the same signature...

Great, I knew you are a cl fan :)

minad:
I added function support back.

oantolin:
Thanks! I'll probably expand the docstrings a bit, but the code looks good to me now. We should give @jaor a heads up: the verbose indicator functions will take all their arguments as keyword arguments; I think this is a good, readable way of doing things here, since there are so many parameters.

jaor:
On Thu, Jul 29 2021, Omar Antolín Camarena wrote:

> Thanks! I'll probably expand the docstrings a bit, but the code looks
> good to me now. We should give @jaor a heads up: the verbose indicator
> functions will take all their arguments as keyword arguments;

works for me, thanks!

PR #309, opened 2021-07-29
Some readme improvements (Fix #235)

### LINKED ISSUE #235: Document the Embark calling convention in the README
* For commands, the target is injected in the minibuffer
* The target is passed as first argument to unary functions

minad:
https://github.com/oantolin/embark/wiki/How-does-Embark-call-actions%3F also as a PR. I think you should keep the README in your style, therefore I put it in the wiki. So feel free to close or rewrite this as you like. But it would be instructive to have some very simple examples of how these actions are actually called. Also the Embark/Consult/Marginalia section should be updated. The changes happend a while ago, so it is better to describe the status quo.

### DIFF
--- a/README.org
+++ b/README.org
@@ -174,8 +174,7 @@ keymap (and any others that also contain a binding for
 =switch-to-buffer=). The end result is that I can type =C-. B f= to
 switch to =find-file=.
 
-* Configuration
-** Quick start
+* Quick start
 
 The easiest way to install Embark is from Melpa. It is highly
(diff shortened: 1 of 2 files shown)


oantolin:
I agree with the essence of everything here, I'll just reword some things a bit.

oantolin:
And I just learned I can push changes to your branch! I had no idea.

oantolin:
Thanks again!

minad:
Thanks. Good to have an updated readme :)

oantolin:
It's still lagging behind, but getting there.

Issue #301, opened 2021-07-28
Fix duplicate menus after update


salutis:
I updated Embark and now see duplicate menus on `embark-act`:

<img width="1257" alt="Screen Shot 2021-07-28 at 17 15 23" src="https://user-images.githubusercontent.com/1158274/127348238-628f8f8f-ae23-4dfa-87eb-309ef1522418.png">

My configuration:

```emacs-lisp
(package-install 'embark)
(global-set-key (kbd "C-.") 'embark-act)
(setq-default embark-prompter 'embark-completing-read-prompter)
```

As for `embark-prompter`, I followed `README.md`:

> If you find you prefer entering actions that way, you can configure embark to always prompt you for actions by setting the variable `embark-prompter` to `embark-completing-read-prompter`.

oantolin:
The duplication is because you have all three of these things in your configuration:

- You have `embark-indicator` set to `embark-verbos-indicator`. This produces the top list of actions.
- You have, as you said, `embark-prompter` set to `embark-completing-read-prompter`. This on its own doesn't produce the bottom listing, it depends on your completion UI.
- You are using Vertico as your completion UI. Vertico automatically displays the completion candidates, this is the bottom display.

You can fix the repetition changing any one of these three things. Since you are unlikely to want to stop using Vertico, and since you chose the completing-read prompter consciously, you'll probably want to change the first item: `(setq embark-indicator 'embark-minimal-indicator)`.

bdarcus:
> Since you are unlikely to want to stop using Vertico, and since you chose the completing-read prompter consciously, you'll probably want to change the first item:

Or he could remove that line to try the new "verbose" option?

oantolin:
True, as @bdarcus says, you might prefer instead of narrowing the list by typing and maybe scrolling around the action list, you might prefer to try using the key bindings directly, which can be faster. For that, as @bdarcus says, just keep the default value for `embark-indicator` and for `embark-prompter`.

salutis:
@oantolin 

Thank you so much for the detailed explanation; it gave me actual understanding of the problem. I set `embark-indicator` to  `'embark-minimal-indicator` and everything works as well.

@bdarcus @oantolin 

All right, I have tried the new UI. It is _wonky_. I am use a relatively small (13") screen and thus cannot see all options. So, the first thing I did was to move the scrollbar. That crashed with:

```
Error running timer: (error "scroll-bar-toolkit-scroll must be bound to an event with parameters")
```

[...]

PR #312, opened 2021-07-30
Fix verbose indicator

minad:
If no shadowed targets are present, do not show cycle key and other targets. This issue has been mentioned by @salutis in #301.


oantolin:
Thanks! I noticed this yesterday while I was testing scrolling and put this on my todo list.

Issue #310, opened 2021-07-30
Clear echo-area after mixed indicator shows verbose actions' buffer


protesilaos:
The `(setq embark-indicator #'embark-mixed-indicator)` first shows the minimal indicator in the echo area and then pops up the verbose buffer.  When the latter is visible, the echo area retains the information it provided, thus creating a duplication between it and the verbose buffer's header.  What do you think about clearing the echo area after the verbose buffer pops up?

![Screenshot from 2021-07-30 09-33-46](https://user-images.githubusercontent.com/12828033/127610958-e1fc6afa-c4a2-4a7d-9a68-244f02daec30.png)

minad:
Yes, this ist true. My personal preference would be to not clear the echo area in order to reduce the amount of jumping elements. Furthermore I am using the reuse window configuration where the verbose indicator does not open directly above the window. In this configuration the result looks less odd than in your case where you have the act message above and below the bindings.

We could add an option `embark-mixed-indicator-move/clear/close` (any name suggestions?) which allows to configure to destroy the mindicator early.

protesilaos:
> Yes, this ist true. My personal preference would be to not clear the
> echo area in order to reduce the amount of jumping
> elements. Furthermore I am using the reuse window configuration where
> the verbose indicator does not open directly above the window. In this
> configuration the result looks less odd than in your case where you
> have the act message above and below the bindings.

Ah yes, in that case it doesn't look odd.

> We could add an option `embark-mixed-indicator-move/clear/close` (any
> name suggestions?) which allows to configure to destroy the mindicator
> early.

Options are nice, though I guess it may be too specific and thus kind of
unnecessary or confusing?

Thinking about it, perhaps it is best we leave this for user
configurations.

-- 
Protesilaos Stavrou
https://protesilaos.com

oantolin:
I tried this change @protesilaos but didn't keep it because I too thought it looked "jumpy" when cycling. Try it for yourself with this patch:

```
index c1670d2..0045fe1 100644
--- a/embark.el
+++ b/embark.el
@@ -1102,6 +1102,7 @@ further user intervention."
               (run-at-time
                embark-mixed-indicator-delay nil
                (lambda ()
+                 (message nil)
                  (setq vindicator (embark-verbose-indicator keymap targets)))))
       (setq vindicator (embark-verbose-indicator keymap targets)))
     (setq mindicator (embark-minimal-indicator keymap targets))
```

protesilaos:
[...]

PR #315, opened 2021-07-30
Keep indicator alive when cycling

minad:
This is important for the vindicator which appears after some time. It should
not close when cycling.

WARNING: The indicator calling convention is changed again. I wouldn't guarantee that
it is the last time, so we should document that writing custom indicators is
currently an unstable feature.

As discussed in #310, https://github.com/oantolin/embark/issues/310#issuecomment-890003547

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -242,30 +242,31 @@ Embark comes with three such indicators:
   verbose popup is shown after `embark-mixed-indicator-delay'
   seconds.
 
-The calling convention for indicator functions is as follows:
+The protocol for indicator functions is as follows:
 
 When called from `embark-act', the indicator function is called
-without arguments. The indicator function should then return a
-closure, which captures the indicator state. The returned closure
-must accept up to three optional arguments, the action keymap,
-the target (in the form of a cons of the type and value) and a
-list of other shadowed targets (each of which is also a cons).
-The keymap, targets and prefix keys may be updated when cycling
-targets at point. When called from `embark-become', the indicator
-closure will be called with the keymap of commands to become,
-with a fake target of type `embark-become' and whose value is the
-minibuffer input. Note, in particular, that if an indicator
-function wishes to distinguish between `embark-act' and
-`embark-become' it should check whether the `car' of its target
-argument is `embark-become'.
+without arguments.  The indicator function should then return a
+closure, which captures the indicator state.  The returned
+closure must accept up to three optional arguments, the action
+keymap, the targets (a list of conses of the type and value of
+each target, starting with the current one) and the prefix keys
+typed by the user so far.  The keymap, targets and prefix keys
+may be updated when cycling targets at point.  When called from
+`embark-become', the indicator closure will be called with the
+keymap of commands to become, a fake target list containing a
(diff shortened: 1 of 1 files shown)


oantolin:
I merged the new master in, correctly, I hope. After testing I'm not so sure about this... It basically forbids resizing to fit the contents while cycling.

minad:
I don't use it with the fixed size, I use the reuse window configuration. But there shouldn't be a problem to resize the window to the fit the buffer after update?

I am quite sure that this is a good change since it results in a UI which flickers much less when cycling. I find this quite annoying. The resizing problem should be fixable?

oantolin:
I guess there should be a way to reapply the window parameters from the display buffer list.

minad:
Not sure if this is the correct way (probably not):

~~~
        (when indicator-window
          (bury-buffer embark--verbose-indicator-buffer))
        (setq indicator-window
              (let ((display-buffer-alist
                     `(,@display-buffer-alist
                       (,(regexp-quote embark--verbose-indicator-buffer)
                        ,@embark-verbose-indicator-display-action))))
                (display-buffer embark--verbose-indicator-buffer)))))))
~~~

But it resizes the window.

minad:
We can even simplify the function:

~~~ elisp
(defun embark-verbose-indicator ()
  "Indicator that displays a list of available key bindings."
  (lambda (&optional keymap targets prefix)
    (if (not keymap)
        (when-let ((win (get-buffer-window embark--verbose-indicator-buffer
                                           'visible)))
          (quit-window 'kill-buffer win))
      (embark--verbose-indicator-update
       (if (and prefix embark-verbose-indicator-nested)
           (lookup-key keymap prefix)
         keymap)
       targets)
      (let ((display-buffer-alist
             `(,@display-buffer-alist
               (,(regexp-quote embark--verbose-indicator-buffer)
                ,@embark-verbose-indicator-display-action))))
        (display-buffer embark--verbose-indicator-buffer)))))
~~~

minad:
I pushed the update.

oantolin:
I had inlined the `embark--verbose-indicator-buffer` variable, mostly because it was used before it was declared and its name is longer than the value. Did you reintroduce it? (Which is fine, but move the declaration to before the first use.)

minad:
Sorry, I got tripped up by this change. See my official complaint here https://github.com/oantolin/embark/commit/75b6d56b495e4d42d6f9fae86b4e537488cce692#commitcomment-54218707 ;)

minad:
But to be clear, the last force pushed commit uses string literals everywhere for the buffer name like you prefer. So it should work now.

oantolin:
I don't really prefer the string literal. A variable is fine, but I do mind the compiler warnings, so if you want to use a variable make sure it's defined before the first use.

minad:
The last commit reintroduces the variable, in case you reconsider 75b6d56.

oantolin:
Yes, let's keep the variable.

oantolin:
Thanks for doing this. It does feel smoother.

minad:
Thanks for reworking the code and checking it thoroughly!

PR #318, opened 2021-07-31
Show keymap descriptions for sub-keymaps (Fix #306)

### LINKED ISSUE #306: Verbose indicator: show keymap name/description for prefix keymaps
Afaik the name is available for keymaps set via fset.

minad:


minad:
@oantolin If you dislike this approach, we could also use mapatoms and cache the variable name and documentation. I am not entirely sure. Or one has to support all different variants of how these prefix keymaps can be bound. Someone could also use `(fset 'map (make-sparse-map))` without a separate bound variable.

embark.el
```diff
@@ -838,8 +842,10 @@ first line of the documentation string; otherwise use the word
   (concat ; fresh copy, so we can freely add text properties
    (cond
     ((stringp (car-safe cmd)) (car cmd))
+    ((keymapp cmd)
+     (propertize (if (symbolp cmd) (format "+%s" cmd) "<keymap>")
```
oantolin, line 846: What is the plus sign for? Is that a which-key convention?

embark.el
```diff
@@ -883,7 +894,7 @@ If NESTED is non-nil subkeymaps are not flattened."
                            (eq cmd #'embark-keymap-help))
                    collect (list name
                                  (cond
-                                  ((keymapp cmd) 'keymap)
+                                  ((and (not (symbolp cmd)))(keymapp cmd) 'keymap)
```
oantolin, line 897: I'm guessing you wanted the keymapp inside the and.

embark.el
minad: Yes, which-key does this. We could add a defcustom to format this?

embark.el
minad: Fixed.

embark.el
oantolin: I guess the plus is OK, which-key is probably the most recognized comencemos for this. I prefer not to add yet another defcustom unless someone really wants it.

oantolin:
OK, Emacs does seem to use this funny `fset` pattern for some keymaps, so I guess it's not just crazy which-key users that do that. 😛

minad:
It is still crazy. But at least you shouldn't feel bad now to put functions into keymaps ;)

PR #321, opened 2021-08-01
Try to use post-command-hook in favor of run-at-time

jakanakaevangeli:
As discussed in https://github.com/oantolin/embark/pull/136, I'm re-proposing
this patch.

In Emacs 28, `minibuffer-quit-recursive-edit` became available, which is like
`abort-recursive-edit`, except that it doesn't cancel keyboard macros.

This patch uses this new function if available. It also tries to use
`post-command-hook` instead of `run-at-time` to delay function evaluation. The
code for this is a bit complicated because we have to handle the recursive case
when a delayed function delays another function. This happens when using
embark-become multiple times in a row for example.

With these two adjustments, all embark commands can be used in keyboard macros.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -1313,7 +1313,14 @@ The TARGETS are displayed for actions outside the minibuffer."
   "Abnormal hook, used by `embark--run-after-command'.")
 
 (defun embark--run-after-command (fn &rest args)
-  "Call FN with ARGS after the current commands finishes."
+  "Call FN with ARGS after the current commands finishes.
+Functions will be called in the order of last in, first out.
+This means that functions added later will be called earlier."
+  ;; We don't simply add FN to `post-command-hook' because FN may recursively
+  ;; call this function.  In that case, FN would modify `post-command-hook'
+  ;; from within post-command-hook, which doesn't behave properly in our case.
+  ;; We use our own abnormal hook and run it from PCH in a way that it is OK to
+  ;; modify it from within its own functions.
   (unless embark--run-after-command-functions
     (let (pch timer has-run)
       (setq pch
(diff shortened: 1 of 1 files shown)


minad:
Thank for this and your work on Emacs! A small request - could you please add a few more comments or expand the docstrings with more explanations in particular since this is new functionality? From our recent changes I had the feeling that we should document the internals a bit better. I am also guilty of not doing this in my recent PRs. Probably I will make another PR going over embark-act and the other "critical" functions and add a few more explanations if Omar agrees. There is always the question of the right balance between code and comments.

jakanakaevangeli:
Yes no problem, this new function uses quite a few hoops in order to be
robust (for example, avoiding `dolist'), so it makes sense to add a few
comments.  Pushed.

oantolin:
If the post-commad-hook sometimes doesn't run and you use the timer as a backup, why try the post-command-hook at all? Why not just use the timer?

oantolin:
Also, I'm probably unable to test that one can really use this from keyboard macros since my copy of Emacs 28 does not have minibuffer-quit-recursive-edit (and while I could certainly copy that function, from the implementation it looks like probably something also changed in the C source code to make it work). So I'd really rely on you or others that do have an extremely recent Emacs 28 to test. I hope you don't mind.

oantolin:
Am I understanding correctly that the reason dolist would be incorrect is that you need to pop the function off the list _before_ running it? If so, that comment could be made more explicit. I needed to macroexpand a dolist to figure out what you meant. 😄

oantolin:
Also, I should have started with thanking you! I really want to be able to use embark-act from the minibuffer in macros, and thought it was impossible (well, it still seems to be impossible on Emacs <28 and my 28...).

jakanakaevangeli:
> If the post-commad-hook sometimes doesn't run and you use the timer as
> a backup, why try the post-command-hook at all? Why not just use the
> timer?

During kbd macro execution, a function delayed with run-at-time will
only be called after the whole macro finishes executing, which almost
always means that your kbd macro doesn't behave as expected. That is why
we prefer to delay our function with PCH.

The cases where PCH doesn't run are rare, usually when a minibuffer is
requested from a timer or a process filter, for example

  (run-at-time 1 nil #'read-from-minibuffer "Test: ")

PCH will actually not run after pressing C-g or RET in this minibuffer.
Because we still want embark-act to work in such minibuffers, we set up
the timer as a backup. With this, we lose the ability to use such an
embark-act command in a kbd-macro, but timer/process-based interaction
ruins our kbd-macro anyway, so we don't cause any further harm.

In normal cases, where PCH does run, it always runs before the backup
timer and cancels it, so the timer has no effect.

I decided not to document this in so much detail in the comments to
avoid making them too verbose.

> Also, I'm probably unable to test that one can really use this from
> keyboard macros since my copy of Emacs 28 does not have
> minibuffer-quit-recursive-edit (and while I could certainly copy that
> function, from the implementation it looks like probably something
> also changed in the C source code to make it work). So I'd really rely
> on you or others that do have an extremely recent Emacs 28 to test. I
> hope you don't mind.

Yes, changes to C were necessary, so a recent Emacs 28 is required to
test. I did do some testing with kbd-macros myself, but I'm the patch
author, so I would also appreciate others to try recording and executing
kbd macros containing following commands:
embark-act (with and without a prefix argument)
embark-become, embark-export and embark-collect-snapshot.

> Am I understanding correctly that the reason dolist would be incorrect
> is that you need to pop the function off the list _before_ running it?
> If so, that comment could be made more explicit. I needed to
> macroexpand a dolist to figure out what you meant.

Yes, I will push another commit to document this in more detail.

minad:
@jakanakaevangeli Thanks for putting some work in the documentation!

oantolin:
Looks good! I tested as much as I can with my non-extremely recent Emacs 28. It's nice to be able to `embark-export` in keyboard macros now. 😉 Thank you so much!

PR #322, opened 2021-08-01
Add embark-next/previous-symbol instead of reusing search-forward/backward

minad:
Unfortunately we cannot reuse `search-forward/backward` directly, since it
errors in case the symbol is not found. This will then abort the Embark command
repeat cycle. Another advantage is that the specialized commands respect the
symbol boundaries.


embark.el
```diff
@@ -2990,8 +2990,8 @@ and leaves the point to the left of it."
   ("c" customize-face)
   ("b" make-face-bold)
   ("B" make-face-unbold)
-  ("i" make-face-italic)
-  ("I" make-face-unitalic)
+  ("/" make-face-italic)
+  ("|" make-face-unitalic)
   ("!" invert-face)
   ("gf" set-face-foreground)
```
oantolin, line 2996: If we do this, how about `*` for bold? I can't think of a good binding for unbold, though. I'd like to avoid the prefix `g` for foreground and background.

embark.el
```diff
@@ -2918,7 +2936,9 @@ and leaves the point to the left of it."
   ("H" embark-toggle-highlight)
   ("d" xref-find-definitions)
   ("r" xref-find-references)
-  ("a" xref-find-apropos))
+  ("a" xref-find-apropos)
+  ("n" embark-next-symbol)
+  ("p" embark-previous-symbol))
```
oantolin, line 2941: Oh my god, I hadn't even noticed they were in the wrong map!

embark.el
minad: What about +- for bold/ubold? Do you have a proposal for foreground/background?

embark.el
```diff
@@ -2804,6 +2804,24 @@ before or after the sexp (those are the two locations at which
 (embark--sexp-command raise-sexp)
 (embark--sexp-command mark-sexp)
 
+(defun embark-next-symbol (sym)
+  "Jump to next SYM without an error in case the symbol is not found.
```
oantolin, line 2808: I think I might completely rewrite these:

1. The docstring is written with the point of view of justifying the existence of the command, not explaining what it does! It reads very defensive.
2. I'd prefer using the no-error argument than condition-case.
3. If we did use condition-case, we should probably just catch `search-failed`. Here it's probably no problem to be broad, but I like catching specific errors to avoid hiding others.

oantolin: Oh, also, the interactive prompt could be nicer.

embark.el
oantolin: Yes, `f` for foreground and `b` for background. Which is why I'm trying to get rid of `b` for bold. I like your `+`, `-` suggestion.

embark.el
minad: > The docstring is written with the point of view of justifying the existence of the command, not explaining what it does! It reads very defensive.

I am defensive. From what I've heard from you, you don't like this functionality. I don't want to push functionality in your library which you seem to dislike. But maybe I sensed this wrongly. Just in case, there are more nice use cases, string-inflection, evil-numbers-like functionality, adjusting color values at point etc. So I like this and would like to see if turns out to replace many hydra use cases, since I very much prefer Embark over Hydra since its keymap usage is more lightweight. For non-at-point commands I am playing around with repeat-mode.

> I'd prefer using the no-error argument than condition-case.

Oh you are very right about that.

> If we did use condition-case, we should probably just catch search-failed. Here it's probably no problem to be broad, but I like catching specific errors to avoid hiding others.

Also right. But we should really use noerror.

oantolin: I don't dislike the repeatable action functionality, I just don't think it's very necessary. But it's in now and I fully support it. I just don't think the debate about it should go into docstrings, which should just matter-of-factly describe the command. The evil numbers example is nice, I might even want that in my config!

I'll push a commit in a few minutes rewriting these.

minad: Pushed the update with the noerror argument.

embark.el
```diff
@@ -2805,22 +2805,20 @@ before or after the sexp (those are the two locations at which
...
-`embark-repeat-commands'. The search respects symbol boundaries."
+  "Jump to previous SYM.
+Prints a message when the symbol is not found. The search
+respects symbol boundaries."
  (interactive "s")
-  (condition-case nil
-      (re-search-backward (format "\\_<%s\\_>" (regexp-quote sym)))
-    (t (message "Symbol `%s' not found" sym))))
+  (unless (re-search-backward (format "\\_<%s\\_>" (regexp-quote sym)) nil t)
+    (search-failed (message "Symbol `%s' not found" sym))))
```
oantolin, line 2821: Is there a search-failed function? It feels more like leftovers from condition-case

minad:
Ready from my side!

embark.el
minad: +- for bold unbold and f/b for foreground/background.

oantolin (commented):
> Ready from my side!

Thanks for rewriting the functions, I would have done it, but you saved me the trouble! Could you remove the `search-failed` you accidentally left in one of them? Then I'll merge.

embark.el
minad: Oh derp, I am bit tired right now. Sorry.

minad: Fixed.

oantolin:
Perfect, thank you!

minad:
Thanks!

PR #323, opened 2021-08-02
Use alists for the hooks, pass target and bounds to the hooks

minad:
Using alist for hooks seems like a good solution. We end up with generic pre-action hooks: embark--ignore-target and embark-beginning-of-target. Furthermore we can get rid of the target bounds variable again. Or shall we keep the dynamically bound variable such that the hooks don't have to take an argument?

As discussed in https://github.com/oantolin/embark/commit/c8125e65350ec7e931d4687998fae53fbd4c95d2.

### DIFF
--- a/README.org
+++ b/README.org
@@ -334,7 +334,7 @@ it, for all commands except those listed in =embark-skip-edit-commands=.
 ** Running some setup after injecting the target
 
 You can customize what happens after the target is inserted at the minibuffer
-prompt of an action. There are =embark-setup-hooks=, that are run by default
+prompt of an action. There are =embark-setup-action-hooks=, that are run by default
 after injecting the target into the minibuffer. The hook can be specified for
 specific action commands by associating the command to the desired hook. By
 default the hooks with the key t are executed.
@@ -343,7 +343,7 @@ For example, consider using =shell-command= as an action during file
 completion. It would be useful to insert a space before the target
 file name and to leave the point at the beginning, so you can
 immediately type the shell command. That's why in =embark='s default
-configuration there is an entry in =embark-setup-hooks= associating
+configuration there is an entry in =embark-setup-action-hooks= associating
 =shell-command= to =embark--shell-prep=, a simple helper command that
 quotes all the spaces in the file name, inserts an extra space at the
 beginning of the line and leaves point to the left of it.
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -273,7 +273,7 @@ actual type."
 
 (defun embark-consult--unique-match (&rest _)
   "If there is a unique matching candidate, accept it.
-This is intended to be used in `embark-setup-hooks' for some
+This is intended to be used in `embark-setup-action-hooks' for some
 actions that are on `embark-allow-edit-commands'."
   ;; I couldn't quickly get this to work for ivy, so just skip ivy
   (unless (eq mwheel-scroll-up-function 'ivy-next-line)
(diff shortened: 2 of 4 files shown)


minad:
I tried the `backward-up-list` action, and it is quite nice. I am glad you found another repeat action ;)

Does it make sense to add `backward-up-list` to the symbol map too? On the other hand this is this problematic since it only makes sense at point. It seems to me that this is a case where Embark is conceptually problematic - one would have to introduce separate minibuffer and at-point maps to fix this, or the actions would have to become context-dependent. For example `backward-up-list` could bail out when executed on a minibuffer target. But this is probably a route we don't want to go.

oantolin:
I wouldn't add backward-up-list to the symbol map, since we can get to the expression target by cycling.

Also I've been thinking that there is no real reason to treat the minibuffer specially. We could let the rest of the target finders attempt to match in the minibuffer as well. As long as the minibuffer target finders have highest priority, this is only an improvement. For example, sometimes in `eval-expression` I want to get help for the symbol I just typed, or kill the previous sexp (to kill forward I still use `C-M-k`, but to kill backwards I find `C-. k` easier to type than `ESC <C-backspace>`).

oantolin:
Maybe we should allow regexp keys in the alist, meaning the hook applies to all commands whose name matches the regexp. The current default value you listed is missing many, _many_ -sexp and -list commands. (Even if we keep them pure alists you can pass to `alist-get`, we really should add `backward-up-list` to the alist, since otherwise if point is somewhere in the middle of the sexp the first time you use `backward-up-list` it feels like it does nothing.)

oantolin:
But probably the most serious issue is that now there is no easy way to write a pre-action hook that applies to all actions. I remember there was someone who used the pre-action hook to decide whether or not to turn off some mode which makes the minibuffer pop up in a separate frame. I guess that person could add that function to all the hooks with different keys, with a loop. Would it be really weird if we decided that the hook with the `t` key runs always?

minad:
I wouldn't add regexps to the list. All the hooks we have right now are keyed by the command and they perform very specific operations. I think these behaviors should be configured per command (Remember the alternative would be to write special action commands like we had before, embark-kill-sexp etc).

oantolin:
Agreed on the regexps. And it's reasonable for the default value of the hook to only focus on actions that are actually bound in the default keymaps. (Which sort of reinforces the incorrect idea that a command needs to be bound in an action keymap to use it, but I guess that's OK.)

minad:
> But probably the most serious issue is that now there is no easy way to write a pre-action hook that applies to all actions.

I added a solution in https://github.com/oantolin/embark/pull/323/commits/e4159c22c0f2135633b8f5031d07fbe144400b1b.

oantolin:
I like it! Except maybe we swap t and nil? Doesn't nil sound like "run this if you found nothing" and t sounds like "always"?

Also, I guess we don't need the hook variables in `embark--act` anymore. I mean they are lexically bound and capture the values of the hooks at the time embark-act is called, but I think using the dynamic variables and maybe more expected for hooks.

minad:
> I like it! Except maybe we swap t and nil? Doesn't nil sound like "run this if you found nothing" and t sounds like "always"?

Hmm, I see your point. But I wouldn't do it since t seems to be used as default by convention. We could also introduce another key `always` or `:always`.

What do you think about making the hooks more uniform, always pass three arguments as in https://github.com/oantolin/embark/pull/323/commits/5b8e15fc309ab55278f2e16114e8e87bb35d6efc?

oantolin:
> But I wouldn't do it since t seems to be used as default by convention.

OK

> We could also introduce another key `always` or `:always`.

They do read better than `nil`. Let's go with `:always` since being a keyword makes it very unlikely anyone will confuse it with a command or write a command with that name.

> What do you think about making the hooks more uniform, always pass three arguments as in [5b8e15f](https://github.com/oantolin/embark/commit/5b8e15fc309ab55278f2e16114e8e87bb35d6efc)?

Sure, uniform is better. I hate that it's a breaking change, but ignore extra arguments is easy. We can bring back the good old "Breaking News" sections of the README.

oantolin:
Oh renaming helps alert people to the change too, good idea.

minad:
I think I am done here.

oantolin:
I'm curious, why did you change command to symbol?

minad:
> I'm curious, why did you change command to symbol?

I got an error with `command`. Is this an Emacs 28 thing?

oantolin:
> I got an error with `command`. Is this an Emacs 28 thing?

Oh! Maybe. Symbol is fine.

oantolin:
Great, thank you!

oantolin:
I think I'm going to remove the special casing of the minibuffer. I do sometimes want other actions in there, mostly when using `eval-expression`, but not only then.

PR #325, opened 2021-08-03
Vertico/Selectrum: Force candidate computation

minad:
As mentioned on reddit. Do you want to have this? (In my opinion it is not strictly necessary to add this. One can always use run-at-time 0).


oantolin:
I also think it's not necessary for normal interactive use, but I like that it's a small change, clearly more correct and simplifies automation.

PR #326, opened 2021-08-03
Add embark--restart (Fix #117)

### LINKED ISSUE #117: Remove candidates from the completion system after action
Is it possible to remove candidates from the completion system after executing a delete action in particular in combination with embark-act-noexit, which I favor greatly over embark-act. This is a relatively minor thing and only applies to delete operations. In the case of static completion tables it does not suffice to refresh the completion system. Probably it is not worth to solve this inside Embark, but maybe you have given it already some thought. One would probably need another completion system `embark-delete-candidate-hook` and then register the commands somehow as deletion commands.

minad:
Now that we have the flexible `embark-post-action-hooks`, this addition is cheap. Furthermore we can reuse the code of `embark-become`.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -396,7 +396,17 @@ are executed always."
                "see the new `embark-pre-action-hooks' variable."
                "0.12")
 
-(defcustom embark-post-action-hooks nil
+(defcustom embark-post-action-hooks
+  '((bookmark-delete embark--restart)
+    (bookmark-rename embark--restart)
+    (delete-file embark--restart)
+    (rename-file embark--restart)
+    (copy-file embark--restart)
+    (delete-directory embark--restart)
+    (make-directory embark--restart)
+    (kill-buffer embark--restart)
+    (embark-rename-buffer embark--restart)
+    (package-delete embark--restart))
   "Alist associating commands with post-action hooks.
 The hooks are run after an embarked upon action concludes. The
 hooks must accept three arguments, the action, the target string
@@ -1782,17 +1792,23 @@ point."
            (become (unwind-protect
                        (embark--prompt indicator keymap targets)
                      (funcall indicator))))
-      (if (null become)
-          (user-error "Canceled")
-        (embark--quit-and-run
+      (unless become
+        (user-error "Canceled"))
+      (embark--become-command become target))))
+
+(defun embark--become-command (command input)
+  "Quit current minibuffer and start COMMAND with INPUT."
+  (embark--quit-and-run
+   (lambda ()
+     (minibuffer-with-setup-hook
          (lambda ()
-           (minibuffer-with-setup-hook
-               (lambda ()
-                 (delete-minibuffer-contents)
-                 (insert target))
-             (let ((use-dialog-box nil)
-                   (this-command become))
-               (command-execute become)))))))))
+           (delete-minibuffer-contents)
+           (insert input))
+       (let ((this-command command)
+             ;; the next two avoid mouse dialogs
+             (use-dialog-box nil)
+             (last-nonmenu-event 13)))
+         (command-execute command)))))
 
 (defmacro embark-define-keymap (name doc &rest bindings)
   "Define keymap variable NAME.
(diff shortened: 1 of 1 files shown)


embark.el
```diff
@@ -2885,6 +2901,18 @@ respects symbol boundaries."
...
 
+(defun embark--restart (&rest _)
+  "Restart current command with current input.
+Use this to refresh the list of candidates for commands that do
+not handle that themselves."
+  (when (minibufferp)
+    ;; XXX: This setter is needed in order to allow restarting
+    ;; again after the restart. I don't understand how
+    ;; embark--command is treated in Embark, sometimes it is
+    ;; let-bound, sometimes setq'ed.
```
oantolin, line 2912: Sometimes using let and sometimes setq is not the result of careful thought and is not on purpose. It's just me being sloppy. Probably they can all be let.

oantolin:
Fine, I guess this is a net win for users, so let's do it.

As you know, I'm a little reluctant to add here a workaround for this limitation of some completion tables that are too static and some completion UIs. For example, Vertico does things right! And Vertico works just fine with the built-in buffer and file completion tables that are appropriately dynamic.

This thing is not Embark's fault and Embark can't actually fix the problem! This PR only deals with the case when the problem is triggered by using Embark actions. The issue really should be fixed at the level of completion tables and completion UIs, so that the fix doesn't just apply to Embark actions, but also to recursive minibuffers, recursive editing, and just plain external changes like removing files from an external file manager.

But I think there is little or no will to correctly address the problem, and certainly having this here will help solve the part of the problem that Embark exposes. And there are many Selectum users of Embark, who would greatly benefit from this. So I guess we should do it. I do however wish people would just switch to Vertico and that we'd fix the completion tables in Consult, instead.

oantolin:
This looks good as it is, but we still need to worry about restoring the selected candidate for Vertico, Selectrum and Icomplete, right? Or do you think it's fine to not restore the selected candidate?

minad:
I agree that the situation is not ideal, but it solves a real problem I have with Embark for me. I am mostly using Vertico with embark-quit-after-action=nil. When deleting bookmarks from `bookmark-jump` for example the deleted bookmarks remain visible in the completion UI. I think having a proper fix, as you would like it is simply unrealistic. Basically it means that you cannot pass lists to `completing-read` anymore. Instead functions (programmable completion tables) must be used everywhere and the candidates must be recomputed always, which is very expensive. This also precludes this solution for `consult-buffer` for example. Furthermore since `bookmark-jump` already fails, why should I do "better" in `consult-buffer`. It is a pragmatic solution, which is actually satisfactory for me. Here we also don't restore the selected candidates, which is okay for me. But if someone really cares about this (I don't) we could upgrade `embark--restart` to a variant which is completion UI specific.

minad:
> This looks good as it is, but we still need to worry about restoring the selected candidate for Vertico, Selectrum and Icomplete, right? Or do you think it's fine to not restore the selected candidate?

Yes, it is fine for now!

oantolin:
> It is a pragmatic solution, which is actually satisfactory for me.

It's good enough for me too, which is why I already said we should do this. 😉 

If not restoring the selected candidate is good enough for you, it's good enough for me too and we can merge now.

oantolin:
Thanks for insisting on this, and I'm sorry my reluctance made it necessary to insist.

minad:
> It's good enough for me too, which is why I already said we should do this. If not restoring the selected candidate is good enough for you, it's good enough for me too and we can merge now. Thanks for insisting on this, and I'm sorry my reluctance made it necessary to insist.

Thanks! I knew that this issue wasn't on your hot list. I guess I should try to finish the job that @clemera started - converting you to Selectrum. I will of course convert you to Vertico instead by adding a vertico-for-omar.el Vertico extension, which does the right thing. That is, disable sorting, introduce a sluggish UI and show all candidates in a tablist buffer instead of through the minibuffer pin hole. Maybe I could put vertico-for-omar.el into Consult such that it doesn't involve inconvenient modifications to your configuration :stuck_out_tongue:

oantolin:
I deserve most of that, but will point out that the way I use embark-collect-completions, which is only by demand and usually after typing several characters, it is instant not at all sluggish.

I think most of all I like the peace of not having things popup or move around automatically. It's why I prefer VC to magit for simple operations, why I prefer `embark-prefix-help-command` to which-key, why I prefer the minimal indicator over the verbose or mixed indicators, and why I prefer completions to only appear when summoned.

minad:
I understand your point of view. It is not too far away from my perspective. On the other hand I am not very fond of timer based solutions (currently I have only Corfu where I enabled this for testing purposes). In contrast to the automatic prefix key help, Vertico is summoned explicitly. When I enter a command I explicitly say "show me the completion UI now", but when I enter a prefix key and hesitate it doesn't feel as I am requesting a popup now (Like you I have disabled the verbose indicator again). Somehow for me there is a difference between completion and entering keys.

PR #327, opened 2021-08-03
Correction after embark--become-command got fixed!

minad:
There is no problem with let-binding/setq-local, the problem was the parenthesis mistake! goto fail!


oantolin:
Yup, works fine!

PR #329, opened 2021-08-06
Cancel embark-act when acting again

minad:
When calling act twice in the minibuffer twice the action target jumps to the last buffer, because actions are executed in the original buffer. In order to avoid this cancel the action. Pressing the embark-act key twice in a row became common due to cycling, therefore it makes sense to catch this in the minibuffer. I don't think actual target cycling should be supported in the minibuffer.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -1664,6 +1664,11 @@ target."
                             (user-error "Canceled")))
                        (default-action (or default-done
                                            (embark--default-action type))))
+            ;; When acting again in the minibuffer, treat it as
+            ;; cancellation. Otherwise the next embark-act will find
+            ;; a target in the original buffer.
+            (when (eq action #'embark-act)
+              (user-error "Canceled"))
             (if (eq action #'embark-cycle)
                 (setq targets (embark--rotate
                                targets (prefix-numeric-value prefix-arg)))
--- a/embark.el
+++ b/embark.el
@@ -1645,54 +1645,55 @@ target."
     (when (and arg (not (minibufferp)))
       (setq targets (embark--rotate targets (prefix-numeric-value arg))))
     (unwind-protect
-      (while
-          (pcase-let* ((`((,type . ,target)
-                          (,_otype . ,otarget)
-                          . ,bounds)
-                        (car targets))
-                       (action
-                        (or (embark--highlight-target
-                             bounds
-                             #'embark--prompt
-                             indicator
-                             (let ((embark-default-action-overrides
-                                    (if default-done
-                                        `((t . ,default-done))
-                                      embark-default-action-overrides)))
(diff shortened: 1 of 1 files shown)


oantolin:
If we don't want target cycling in the minibuffer, it's probably better to go back to your version of `embark--targets`, which only found one target in the minibuffer.

When I changed that function I didn't realize that subsequent targets would come from the previous buffer. That's certainly awkward and not what I meant: I wanted target cycling in the minibuffer. I had a couple of instances in mind, both for `eval-expression`: getting help for the last symbol you typed, and saving the sexp you just closed on the kill ring.

But maybe it is not worth having target cycling in the minibuffer for just `eval-expression`. Why do you not want to have it?

minad:
> If we don't want target cycling in the minibuffer, it's probably better to go back to your version of embark--targets, which only found one target in the minibuffer.

This will fix the issue in another way, since then calling embark-act twice will act again instead of cancelling. We can implement this behavior more cleanly in `embark-act` by not aborting but staying in the while loop. I would avoid adding special casing to `embark--targets` adding a check for the minibuffer. I would go with the simplest fix which resolves the current issue.

1. Cancel as I have it in this PR.
2. Stay in the while loop like embark-cycle without rotating the targets.

> But maybe it is not worth having target cycling in the minibuffer for just eval-expression. Why do you not want to have it?

My feeling is that it will complicate the UI without significant benefits. For example one would also need a cycle indicator in the minibuffer instead of only the simple "Act" label. Then there is the complication that actions are executed in the original buffer. By adding support for at-point actions in the minibuffer we would complicate this logic. With UIs like Vertico having multiple targets in the minibuffer will feel even more odd since you have the current candidate highlighted, but the input may also act as target.

oantolin:
> (EDIT: It seems `embark--targets` only returns only the first candidate in the minibuffer, which is an optimization. But this is unrelated to the issue here.)

It does! Sorry for the confusion. I misremembered what I did. I did think of changing that, but didn't actually do it.

>  I would avoid adding special casing to embark--targets adding a check for the minibuffer.

Well, it does have a special check for the minibuffer right now, but I don't think it's too bad, we can keep it.

> My feeling is that it will complicate the UI without significant benefits.

On balance, I think you are right.

oantolin:
So it only remains to decide whether a second embark-act in the minibuffer cancels or prints a message. I take it from the new commit that you prefer to print a message?

minad:
Yes, since it's about being helpful here, the message is probably better than cancelling (https://github.com/oantolin/embark/pull/329/commits/bb5ec17fedc5173042728e0bad1656779b4ce2a7).

oantolin:
Done. Thanks!

PR #332, opened 2021-08-07
Fix the tofu problems and other cleanup (Fix #331)

### LINKED ISSUE #331: consult-line+embark-collect-snapshot has a tofu problem
1. Search using consult-line
2. embark-collect-snapshot
3. Act on some match (jump to the original window)
4. Move some lines around
5. Acting again on other matches won't work anymore since the tofus are invalidated because of the line number changes

I thought we had handled the tofus? Or did some of the recent refactorings break this?

minad:
- Replace the tofu acceptor with embark-consult-goto-location (faster, more robust)
- Rename the old embark-consult-goto-location to embark-consult-goto-grep
- Use the Consult jumper for embark-consult-goto-grep (requires newest Consult commit)
- Remove Ivy check (Consult does not support Ivy)
- Make the default action overrides functions, such that they receive the full
  target with properties. Do the same for the grep default action override for
  consistency.

Please give this a careful check, that I didn't miss some details. I think the approach I have chosen now is more robust in principle and also results in less code.

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -111,6 +111,13 @@
 (setf (alist-get 'consult-location embark-transformer-alist)
       #'embark-consult--target-strip)
 
+(defun embark-consult-goto-location (target)
+  "Jump to consult location TARGET."
+  (consult--jump (car (get-text-property 0 'consult-location target))))
+
+(setf (alist-get 'consult-location embark-default-action-overrides)
+      #'embark-consult-goto-location)
+
 (defun embark-consult-export-occur (lines)
   "Create an occur mode buffer listing LINES.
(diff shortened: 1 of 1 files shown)


oantolin:
Perfect, thanks! It's relief to finally get rid of that odd `compile-goto-error` implementation that has caused trouble in the past.

minad:
Thanks for merging! I hope you gave it a good test. I am always a bit wary with all these Embark details.

oantolin:
I think I tested the main cases: default action vs other actions for both consult-line and consult-grep, and for consult-grep I tested searching from the current directory and from a different one.

Issue #330, opened 2021-08-07
Generalize embark-indicator to a list of indicators


minad:
* Support multiple indicator hooks.
* Use an alist with target type as key? But it would probably be easier to make it a plain list. The indicator gets passed the target as argument so it can simply bail out on unsupported target types.
* Make target highlighting an `:always` indicator instead of the ad-hoc `embark--highlight-target` function we have right now
* As proposed by @jdtsmith in https://github.com/oantolin/embark/issues/291#issuecomment-894588296. Add a special symbol indicator which highlights all symbols after a delay. We could either reuse hi-lock or maybe the Isearch lazy highlighting mechanism?
* More fancy features become possible, e.g., highlighting the current target in completion UIs like Vertico with the "Act" highlight face.

I am glad we added this comment to the source:

https://github.com/oantolin/embark/blob/e4bdbd55af2e948d0bceb5d08b9d9518d91fc563/embark.el#L287-L292

oantolin:
Oops, I mentioned this issue by accident in commit 4f8b5a6, I meant #320.

minad:
Well, the issue has been resolved now. I opened #336 as a follow-up to keep the symbol highlighting idea in mind :)

minad:
See also https://github.com/oantolin/embark/wiki/Additional-Configuration#colorize-the-current-vertico-candidate-differently-when-acting

oantolin:
Nice!

oantolin:
Generalizing embark-indicator to a list of indicators was a great idea, @minad, thanks again!

minad:
Thanks! Also thanks to @jdtsmith to get the ball rolling on this :)

I guess one could even find more specialized indicators which only activate for certain target types or in certain contexts.

oantolin:
Probably time will reveal more of those. But for example, one can imagine a marginalia indicator for at-point targets that displays the marginalia annotation somewhere, the echo area, an overlay or one of those mouse hover popups.

PR #334, opened 2021-08-07
Multiple indicators

minad:
Addresses #330

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -230,8 +230,8 @@ Used by `embark-completing-read-prompter' and `embark-keymap-help'.")
 (defface embark-target '((t :inherit highlight))
   "Face used to highlight the target at point during `embark-act'.")
 
-(defcustom embark-indicator #'embark-mixed-indicator
-  "Indicator function to use when acting or becoming.
+(defcustom embark-indicators (list #'embark-mixed-indicator)
+  "Indicator functions to use when acting or becoming.
 The indicator function is called from both `embark-act' and from
 `embark-become' and should display information about this to the
 user, such as: which of those two commands is running; a
@@ -290,11 +290,7 @@ calling convention should currently be considered unstable.
 Please keep this in mind when writing a custom indicator
 function, or when using the `which-key' indicator function from
 the wiki."
-  :type '(choice
-          (const :tag "Verbose indicator" embark-verbose-indicator)
-          (const :tag "Minimal indicator" embark-minimal-indicator)
-          (const :tag "Mixed indicator" embark-mixed-indicator)
-          (function :tag "Other")))
+  :type '(repeat function))
 
 (defcustom embark-quit-after-action t
   "Should `embark-act' quit the minibuffer?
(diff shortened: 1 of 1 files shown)


embark.el
```diff
@@ -1416,20 +1416,17 @@ queued most recently to the one queued least recently."
...
 The HOOKS argument must be an alist. The keys t and :always are
 treated specially. The :always hooks are executed always and the
 t hooks are the default hooks, if there are no command-specific
 hooks."
-  (let ((embark--action-hook (or (alist-get action hooks)
-                                 (alist-get t hooks))))
-    (apply #'run-hook-with-args 'embark--action-hook action args))
-  (let ((embark--action-hook (alist-get :always hooks)))
-    (apply #'run-hook-with-args 'embark--action-hook action args)))
+  (mapc (lambda (h) (apply h action args))
```
oantolin, line 1428: Yeah, probably a saner way to do it.

embark.el
```diff
@@ -257,24 +259,26 @@ Embark comes with three such indicators:
...
-may be updated when cycling targets at point resulting in
-multiple calls to the closure.  When called from `embark-become',
-the indicator closure will be called with the keymap of commands
-to become, a fake target list containing a single target of type
-`embark-become' and whose value is the minibuffer input, and the
-prefix set to nil.  Note, in particular, that if an indicator
-function wishes to distinguish between `embark-act' and
-`embark-become' it should check whether the `car' of the first
-target is `embark-become'.
+without arguments. The indicator function should then return a
```
oantolin, line 268: Um, did you just reformat everything with a a single space between sentences? Please use `sentence-end-double-space = t`, like the built-in documentations strings do.

minad: Yes, I have disabled this insanity. I will reformat.

oantolin: I also don't like it, _except_ for docstrings, so it looks more like the built-in docs. I usually turn it on only when editing an Emacs Lisp packages (not even in my personal configuration).

minad: I don't even turn it on for Emacs packages. Call it a small rebellion against the Emacs authorities :-P

oantolin:
Mmh. It might be nice to even change the target finder API, so that target finders must now return a `:plist`. Although I guess that would break a lot of people's code.

minad:
> Mmh. It might be nice to even change the target finder API, so that target finders must now return a :plist. Although I guess that would break a lot of people's code.

I generally agree that this is a good idea. But let's not do this now but in the next iteration.

embark.el
hmelman: > I don't even turn it on for Emacs packages. Call it a small rebellion against the Emacs authorities :-P

Just curious, do you not use the sentence motion commands?  I use them all the time and hate when they don't work because of initials or something.

minad: Sentence movement works for me with `sentence-end-double-space=nil`.

embark.el
```diff
@@ -290,11 +294,7 @@ calling convention should currently be considered unstable.
 Please keep this in mind when writing a custom indicator
 function, or when using the `which-key' indicator function from
 the wiki."
-  :type '(choice
-          (const :tag "Verbose indicator" embark-verbose-indicator)
-          (const :tag "Minimal indicator" embark-minimal-indicator)
-          (const :tag "Mixed indicator" embark-mixed-indicator)
-          (function :tag "Other")))
+  :type '(repeat function))
```
oantolin, line 297: We lost some user friendliness here.

minad: Yes, let's not be too friendly for now. The indicators will probably change again ;)

oantolin: OK.

oantolin:
Everything looks good, I'll merge as soon as I test. (I'm testing the other PR right now).

oantolin:
Man, learning that I can push commits to a PR for simple fixes _before_ merging is small but noticeable workflow improvement.

embark.el
```diff
@@ -298,7 +298,13 @@ calling convention should currently be considered unstable.
 Please keep this in mind when writing a custom indicator
 function, or when using the `which-key' indicator function from
 the wiki."
-  :type '(repeat function))
+  :type '(repeat
```
oantolin, line 301: Sorry, I couldn't resist.

embark.el
```diff
@@ -230,19 +230,25 @@ Used by `embark-completing-read-prompter' and `embark-keymap-help'.")
...
   "Face used to highlight the target at point during `embark-act'.")
 
-(defcustom embark-indicator #'embark-mixed-indicator
-  "Indicator function to use when acting or becoming.
-The indicator function is called from both `embark-act' and from
-`embark-become' and should display information about this to the
-user, such as: which of those two commands is running; a
+(make-obsolete embark-indicator
+               "see the new `embark-indicators' variable."
+               "0.12")
```
minad, line 235: Does this make sense? Did you tag a release in the meantime? We went from embark-action-indicator -> embark-indicator -> embark-indicators.

oantolin: Oh, probably not! I'm supposed to tag 0.12 first, and then mention it in these messages, right? Not the other way around.

minad: I would not make a new release before this. But I usually only add deprecations based on the latest tagged release. Everything which happens in between is development and things can move around more freely.

PR #361, opened 2021-09-18
C-g is bound to #'abort-minibuffers in Emacs 28

jakanakaevangeli:
Support C-g on GNU Emacs 28.

Btw, would it be OK with you if we send such small patches to your e-mail? For
me personally, it's kind of an overkill to open a web browser and click stuff
just to propose a simple one-line change.

(Obviously, I'd only use e-mail if I'm 100% sure that the patch is trivial
enough to not require any discussion.)


oantolin:
Sure, patches by email are welcome too! (@protesilaos taught me how to apply them.)

jakanakaevangeli:
OK, thanks!

PR #367, opened 2021-10-02
Don't shadow `pp-eval-expression` binding in `embark-defun-map`

roshanshariff:
Change the binding of `eval-defun` from `e` to `d` in
`embark-defun-map` and remove its binding to `RET`. This change
unshadows the bindings of `pp-eval-expression` in the parent
`embark-expression-map`.

Top level expressions are categorized as defuns instead, and it is
useful to be able to call `pp-eval-expression` on them, which also
does the right thing for defuns. Usually, `eval-defun` is only needed
to instrument a function for edebug (using a prefix argument). This
use case is already covered by the existing `D` binding to
`edebug-defun`.

The only other reason to call `eval-defun` on top-level forms is to
reset variables defined via `defvar` or `defcustom`. The `d` binding
is provided for these cases.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -3337,8 +3337,7 @@ and leaves the point to the left of it."
 (embark-define-keymap embark-defun-map
   "Keymap for Embark defun actions."
   :parent embark-expression-map
-  ("RET" eval-defun)
-  ("e" eval-defun)
+  ("d" eval-defun)
   ("c" compile-defun)
   ("l" elint-defun)
   ("D" edebug-defun)


oantolin:
I'm a little confused. I see that you're proposal is a valid alternative. What I don't see is what advantage it has over the current bindings. It seems equally good, not better.

roshanshariff:
Currently there is no way to call `pp-eval-expression` on a top-level expression, since it is treated as a defun and only `eval-defun` is available through embark. If the expression has a complicated value, I really want it pretty-printed in a separate buffer instead of condensed and shown in the echo area as `eval-defun` does.

The proposal is to just use `pp-eval-expression` as the default action and the `e` binding, because it is almost always superior to `eval-defun`. When the return value is simple, both functions just print it in the minibuffer.

oantolin:
Got it. Good point.

minad:
> The proposal is to just use |pp-eval-expression| as the default action
> and the |e| binding, because it is almost always superior to
> |eval-defun|. When the return value is simple, both functions just print
> it in the minibuffer.

I am not a fan of this change. Using RET to redefine defvars/defcustoms
is exactly why I want eval-defun. Having to use d for this is less
convenient. Is there a way to allow redefinition and at the same time
allow pretty printing of the evaluation result? However for me pretty
printing toplevel definition is less useful and rarely needed, so I was
more happy with the old definition.

minad:
I think the problem here is that all toplevel definitions are defuns, so
pretty printing is indeed valuable. Maybe we should write our own
function which either runs eval-defun for defvar/defcustom or runs
pp-eval-expression for all other expressions? However this wouldn't work
for other definitions based on macros, which wouldn't be perfectly
satisfactory. Another alternative would be to evaluate twice or somehow
replace/redirect the printing of eval-defun to use pp instead.

minad:
I propose to use this function instead, which combines the best of both
worlds :)

(defun embark-pp-eval-defun (edebug)
  (interactive "P")
  (cl-letf (((symbol-function #'eval-expression-print-format) #'ignore))
    (pp-display-expression (eval-defun edebug) "*Pp Eval Output*")))

oantolin:
I like that idea @minad, but for you're use case doesn't it bother you @minad that if you have a defvar with a large value your `embark-pp-eval-defun` will pop up a buffer with the pretty-printed value? For example:

```elisp
(defvar foo (cl-loop for i upto 100 collect (list i (* i i))))
```

minad:
> I like that idea @minad <https://github.com/minad>, but for you're use
> case doesn't it bother you @minad <https://github.com/minad> that if you
> have a defvar with a large value your |embark-pp-eval-defun| will pop up
> a buffer with the pretty-printed value? For example:
> 
> (defvar foo (cl-loop for i upto 100 collect (list i (* i i))))

That's odd. This does not happen for me, embark-pp-eval-defun works as
expected printing only "foo". Maybe you accidentially executed the
Embark dwim action on the expression only and not on the whole defun
such that you got the large evaluation result?

minad:
> I like that idea @minad <https://github.com/minad>, but for you're use
> case doesn't it bother you @minad <https://github.com/minad> that if you
> have a defvar with a large value your |embark-pp-eval-defun| will pop up
> a buffer with the pretty-printed value? For example:

Oh, now I see the issue. When you redefine the value it indeed opens the
Pp buffer. I will try to revise the function.

minad:
Okay, this revised function seems to work properly. Please give it
another try.

(defun embark-pp-eval-defun (edebug)
  (interactive "P")
  (cl-letf (((symbol-function #'eval-expression-print-format)
             (lambda (result)
               (pp-display-expression result "*Pp Eval Output*"))))
    (eval-defun edebug)))

roshanshariff:
Thank you! I tried out this function and it works very well for defvar, defcustom, defun, etc., and at the same time also for other top-level forms like one might enter in lisp-interaction-mode. This is perfect for my use case.

Issue #369, opened 2021-10-05
Indicators give errors for keys not in action keymap


roshanshariff:
Currently, `embark-verbose-indicator` and `embark-which-key-indicator` (from the wiki) give errors when typing a key sequence that is not in the action keymap, but is bound globally. This is because they show the keymap returned by `(lookup-key keymap prefix)`, which is nil when the prefix is not bound in the keymap. Here is an updated version of `embark-which-key-indicator` that solves the problem:
```emacs-lisp
(defun embark-which-key-indicator ()
  "An embark indicator that displays keymaps using which-key.
The which-key help message will show the type and value of the
current target followed by an ellipsis if there are further
targets."
  (lambda (&optional keymap targets prefix)
    (if (null keymap)
        (which-key--hide-popup-ignore-command)
      (when-let ((prefix-keymap
                  (if prefix
                      (let ((overriding-terminal-local-map keymap))
                        (key-binding prefix 'accept-default))
                    keymap))
                 ((keymapp prefix-keymap)))
        (which-key--show-keymap
         (if (eq (caar targets) 'embark-become)
             "Become"
           (format "Act on %s '%s'%s"
                   (plist-get (car targets) :type)
                   (embark--truncate-target (plist-get (car targets) :target))
                   (if (cdr targets) "…" "")))
         prefix-keymap nil nil t)))))
```
A similar change can be made to `embark-verbose-indicator` as well, though perhaps the user should be able to decide whether the indicator remains visible once an unbound prefix is entered.

If you agree with this change, I can prepare a pull request fixing these functions (and adding the customization variable).

minad:
> A similar change can be made to |embark-verbose-indicator| as well,
> though perhaps the user should be able to decide whether the indicator
> remains visible once an unbound prefix is entered.
> 
> If you agree with this change, I can prepare a pull request fixing these
> functions (and adding the customization variable).

I don't see an error for the embark-verbose-indicator. The indicator
stays alive but is updated and shows nothing (the empty/nil keymap). It
would be better if we lookup the prefix in the currently active keymaps.
This can be achieved by using

(or (lookup-key keymap prefix) (key-binding prefix))

instead of only lookup-key.

roshanshariff:
Right, the error happens if you type a key sequence that is bound to a prefix keymap. So for example `C-x r` shows "Error running timer: (wrong-type-argument keymapp 1)" in the echo area. From looking at the documentation of `lookup-key` it appears this is happening:
[...]

PR #370, opened 2021-10-05
verbose-indicator: lookup key globally

minad:
See #369


oantolin:
Oh, I think it didn't work before either. 😄

oantolin:
And as an implementation detail, I like the `embark--export-rename` macro. It certainly makes the bookmark and apropos exporters tidier.

PR #372, opened 2021-10-07
Run and quit

minad:
This is another approach to implement embark-quit-after-action. I am trying to restore the window configuration but I am not yet completely successful. This PR is based on #370 and the only relevant commit here is the last one.


oantolin:
I also vote for leaving things as they are now. We can certainly revisit this later.

And having a single dynamic-variable-smuggling location does sound very sensible.

PR #375, opened 2021-10-08
Extend the embark url target to find shr urls

karthink:
This is useful when running embark-act on links in shr buffers: webpages, elfeed buffers, mail buffers etc. If the point is on an image and the image is not a link, the image url is found instead.

It's actually possible to find the bounds of the shr url correctly with `(next-single-property-change (point) 'mouse-face)`, but I haven't added it yet. Do you think it'll be useful?

EDIT: I added the bounds.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -674,9 +674,10 @@ In `dired-mode', it uses `dired-get-filename' instead."
     (when-let ((url (or (get-text-property (point) 'shr-url)
                         (get-text-property (point) 'image-url))))
       `(url ,url
-            . ,(cons (previous-single-property-change
-                      (min (1+ (point)) (point-max)) 'mouse-face)
-                     (next-single-property-change (point) 'mouse-face))))))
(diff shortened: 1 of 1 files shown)


oantolin:
I added bounds to the search to fix that error I mentioned and merged. Thanks @karthink!

oantolin:
This reminded me: I've sometimes wished for a download action for URLs. I'm not sure whether it should prompt for a directory or just use `eww-download-directory`. It could be just `eww-download` but prompting for a URL instead of using the shr URL at point:

```elisp
(defun embark-download (url)
  "Download URL to `eww-download-directory'."
  (interactive "sURL: ")
  (let ((dir (if (stringp eww-download-directory)
                 eww-download-directory
               (funcall eww-download-directory))))
    (access-file dir "Download failed")
    (url-retrieve url #'eww-download-callback (list url dir))))
```

PR #377, opened 2021-10-09
Emacs 28: Use current-minibuffer-command

minad:


minad:
Sorry Omar, current-minibuffer-command does not work as expected. It sometimes has the value nil and for M-x something it also has a wrong value. Please revert, I was too fast here.

hmelman:
If it's broken, please file an Emacs 28 bug, it's in pretest.

Issue #187, opened 2021-03-22
Also recognize C-h in sub-keymaps


minad:
When I press `M-x embark-act C C-h` I don't get a completing-read prompter. This is the equivalent of #139 ;)

oantolin:
> Not really, but I still like it

Most other things should be working now. 🤞🏽

oantolin:
> This is the equivalent of #139 ;)

Wow, that bug just won't let me off the hook, huh?

rswgnu:
Minibuffer completions used '?' to show completions so I would advise you stick to that as a default and then that would open up using C-h normally in embark keymaps.

oantolin:
@rswgnu In my personal configuration I unbind `C-h` and use `?` instead, precisely because several useful commands are on the `C-h` prefix. So I definitely agree with you. The reason I picked `C-h` is for consistency with normal prefix keys: if you need a reminder of the suffixes, you press `C-h`.

PR #382, opened 2021-10-15
Support typing help-char in embark-keymap-prompter

roshanshariff:
When the user types `C-h` (the default value of `help-char`) after a prefix key, emacs shows help on the prefix keymap by calling `prefix-help-command`.  This change makes the embark keymap prompter behave similarly: it switches to the completing-read prompter for the prefix keymap typed so far.

This is in addition to the existing functionality, where C-h is bound to embark-keymap-help in embark-meta-map. That is only useful for getting help on the embark action map, not for other keys.

## Example usage

1. Run `embark-act`.
2. Type `C-x` because you want to run some command not in the action map.
3. Type `C-h` to get help on the prefix keymap, because you can't remember the exact keybinding you want.

Currently, embark gives an error because `C-x C-h` is not bound to any command. With this patch, embark opens the completing read prompter with `ctl-x-map`.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -946,7 +946,7 @@ the minibuffer is open, the message is added to the prompt."
                                (setq prefix new-prefix)
                                (when (/= (length prefix) 0)
                                  (funcall update prefix))))))))
-          (read-key-sequence nil nil nil t 'cmd-loop))
+          (read-key-sequence-vector nil nil nil t 'cmd-loop))
       (when timer
         (cancel-timer timer)))))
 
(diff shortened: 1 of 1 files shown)


roshanshariff:
Ah, yes, I hadn't noticed that issue but this should indeed fix it.

embark.el
```diff
@@ -955,25 +955,32 @@ the minibuffer is open, the message is added to the prompt."
...
-  (let* ((key (let ((overriding-terminal-local-map keymap))
-                (embark--read-key-sequence update)))
+  (let* ((keys (let ((overriding-terminal-local-map keymap))
+                 (embark--read-key-sequence update)))
          (cmd (let ((overriding-terminal-local-map keymap))
-                (key-binding key 'accept-default))))
+                (key-binding keys 'accept-default))))
     (pcase cmd
-      ('embark-keymap-help
+      ((or 'embark-keymap-help
```
minad, line 963: I wonder why it is even necessary to check `embark-keymap-help` here. Should this not be subsumed automatically by the more general logic here?

We should probably also not bind `embark-keymap-help` in the `embark-meta-map`.

minad: Ideally we should probably remove all special casing of `embark-keymap-help` and replace it with `help-char` checks as you have it here.

roshanshariff: Yes, I was thinking this too, but there's still a reason to have something like `embark-keymap-help`. The emacs command loop, as well as this, shows help only if `help-char` is not otherwise currently bound. By default, `C-h` is bound to `help-map`, so without special treatment, `C-. C-h` would go into the help prefix commands instead of showing help for the embark action map.

So there needs to be some way of signaling that we want a key to show embark help, rather than behave just as it does outside embark.

minad: I see, you are right. The issue is that for `C-. C-h` the keymap prompter enters the global prefix map `help-map` and does not return immediately. We need `embark-keymap-help` to return immediately. Thanks for the explanation!

minad:
@roshanshariff I gave it a test and it works! Thanks for fixing the long standing #187 :+1:

minad:
Btw, @roshanshariff do you have the FSF assignment since this package is on ELPA? I believe as of now your contributions still fall below the critical threshold of 15 lines of significant changes. In particular in this PR there are multiple lines which only change the variable name "key" to "keys". But it is important to keep the assignment in mind for the future. Sorry for the bureaucracy!

oantolin:
I should add a "contributing" section to the README.

minad:
@oantolin Thanks for adding the section!

roshanshariff:
I don't have an FSF assignment yet @oantolin, but I will work on getting it. Good to know it's not needed for this patch, but I had some more potential pull requests in mind and I'll make sure I get the paperwork done for those.

Issue #203, opened 2021-04-13
Run annotators during export/snapshot


minad:
This way you always get annotations even if the minibuffer is closed. And I suspect it would also be less problematic regarding the annotation function requirements (default-directory, target, ...).

oantolin:
Export has no use for annotations, there is nowhere to put them. So let's forget about export.

I take it you are proposing that Embark collect snapshot buffers compute the annotations once when they are created and then just store them? I guess they could do that. But that strategy doesn't really work for live collect buffers or collect completions buffers. Since I needed to implement recomputing annotations upon revert for those buffer anyway, I just use the same code for snapshot buffers.

minad:
> I take it you are proposing that Embark collect snapshot buffers compute the annotations once when they are created and then just store them? I guess they could do that. But that strategy doesn't really work for live collect buffers or collect completions buffers. Since I needed to implement recomputing annotations upon revert for those buffer anyway, I just use the same code for snapshot buffers.

Yes. Why does it not work for live collect? When collecting for live collect you also run the annotators directly. The only downside I see is that you don't want to collections the annotations when using the grid view.

My proposal would be to store the annotations in `embark-collect-candidates`. Each candidate can either be a string or a list with two or three elements (like the affixation function specifies).

oantolin:
I'm sorry, I think i misunderstood you. The annotations are currently computed when `embark-collect-candidates` whenever is updated. I thought you were proposing to change when they were computed, but you are only proposing to change where they are stored?

minad:
>  The annotations are currently computed whenever embark-collect-candidates is updated (and you are in list view). 

Yes, I know. This means it does not work when the minibuffer is closed.

> I thought you were proposing to change when they were computed, but you are only proposing to change where they are stored (and to also compute them in grid view)?

I propose to compute annotations directly when collecting the candidates. Then the minibuffer is still open and we don't lose them. Second, I propose to store the annotations together with the candidates in the `embark-collect-candidates` list in the three variants I mentioned, `cand`, `(cand suffix)`, `(cand prefix suffix)`. The cand-prefix-suffix list elements are returned as such by the affixation function.

oantolin:
Ah, sorry. This whole conversation I was laboring under the mistaken impression that Embark stored the annotation-function while the minibuffer was still open. I see your point now. I'm sorry for confusion. I think I did use to call revert before the minibuffer was closed, which is why I thought I still did.

oantolin:
[...]

PR #384, opened 2021-10-17
Minor affixator cleanup

minad:
Found this while looking into #203.

EDIT: I guess it is fine to run the Marginalia annotation functions later and not within the minibuffer, so we are alright. The `marginalia--base-position` will be wrong, but that's all that is wrong, I guess? The assumption that the annotators should run in the minibuffer led me to propose #203.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -2401,6 +2401,9 @@ determine the width."
 (defun embark-collect--list-view ()
   "List view of candidates and annotations for Embark Collect buffer."
   (let ((candidates (if embark-collect-affixator
+                        ;; TODO shouldn't this run in the minibuffer?!
+                        ;; Marginalia assumes that the annotation
+                        ;; functions run in the minibuffer.
                         (funcall embark-collect-affixator
                                  embark-collect-candidates)
                       embark-collect-candidates)))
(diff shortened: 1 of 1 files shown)


oantolin:
I think I was confused, and that neither the current code nor the proposed change will run the affixator in the target buffer, which is necessary, for example, to get the right keybindings for commands.

oantolin:
I did do your proposed simplification at the point in which it occurs, but now I also switch windows to the target window if possible, so that the marginalia annotator for commands shows the correct key bindings. 6b9524d12fc6f8ea12241f402aa498f06936c7c3 That sounds more correct to me (of course, once you've closed the target window, the annotator is run in the embark collect window and you get the wrong key bindings again 😢). Maybe this points to your idea of just gathering all the annotations when the embark collect buffer is created and not rerunning them afterwards...

minad:
On 10/18/21 5:41 PM, Omar Antolín Camarena wrote:
> I did do your proposed simplification at the point in which it occurs,
> but now I also switch windows to the target window if possible, so that
> the marginalia annotator for commands shows the correct key bindings.
> 6b9524d
> <https://github.com/oantolin/embark/commit/6b9524d12fc6f8ea12241f402aa498f06936c7c3>
> That sounds more correct to me (of course, once you've closed the target
> window, the annotator is run in the embark collect window and you get
> the wrong key bindings again 😢). Maybe this points to your idea of just
> gathering all the annotations when the embark collect buffer is created
> and not rerunning them afterwards...

Yes, in order to be always correct we have to export the annotations
together with the candidates. The big downside is that it will hurt your
grid view and make it slower. Unfortunately I don't have a good solution
for that. Maybe it is worth paying the price, I don't know. One could
argue that we should optimize for a correct list view and not for an
efficient grid view. We could also avoid the annotation computation for
the grid view and only try to compute the annotations when switching to
the list view, but this sounds like a complication - as complicated as
what we have now and with all the downsides.

oantolin:
As a heavy user of the grid view (it's my second most common way of dealing with completions right after "not displaying them at all"), I do want to avoid computing annotations in grid view.

minad:
> But collecting the annotations with the candidates doesn't make the
> annotations correct either! Key bindings can change, file size can
> change, people can run marginalia-cycle so annotator can easily change...

Sure, but I see the export/collect as a *snapshot* and I am okay if the
buffer does not update if things change. Note that g also doesn't
restart consult-line/consult-ripgrep for snapshot/grep buffers.
Implementing reverting would require quite some effort and I am not sure
if it is worth it. However reverting will certainly involve rerunning
the command in question, such that we get the updated candidate list
inclusive updated annotations.

PR #385, opened 2021-10-18
Additional variable and file actions

minad:


oantolin:
There was an interesting bug with embark-eval-replace that happened if you tried to evaluate code which would open a minibuffer: the expression would be injected at that minibuffer! I have them ignore the target now. c929a79fc2cffed81c7709aa1b8a599ab8ccbfc4

PR #395, opened 2021-10-31
Add ELPA badges

minad:


oantolin:
That works? I never export the README to HTML so I'd naively expect those to never be seen by anyone. Is the trick that GitHub's Org renderer render them even though you aren't exporting to HTML? So this effectively takes advantage of what could be construed as a bug in GitHub's org renderer?

oantolin:
Huh, I just checked that this is how you do it in Consult. OK, then.

oantolin:
Oh, merging was premature: the links to ELPA don't work yet. I'll just leave it like this for now, and soon those links will point to something, I guess.

oantolin:
The badges aren't actually visible, nor here nor on consult, for example. The URL for the SVG does seem to work now, and it certainly looks like an SVG image, but the browser doesn't render it:

```xml
<?xml version="1.0"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="0" height="20" role="img" aria-label="GNU ELPA: consult 0.12">
  <title>GNU ELPA: consult 0.12</title>
  <linearGradient id="s" x2="0" y2="100%">
    <stop offset="0" stop-color="#bbb" stop-opacity=".1"/>
    <stop offset="1" stop-opacity=".1"/>
  </linearGradient>
  <clipPath id="r">
    <rect width="0" height="20" rx="3" fill="#fff"/>
  </clipPath>
  <g clip-path="url(#r)">
    <rect width="0" height="20" fill="#555"/>
    <rect x="-1" width="0" height="20" fill="#bb3955"/>
    <rect width="0" height="20" fill="url(#s)"/>
  </g>
  <g fill="#fff" text-anchor="middle" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" font-size="110" text-rendering="geometricPrecision">
    <text aria-hidden="true" x="-10" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="0">
    GNU ELPA</text>
    <text x="-10" y="140" transform="scale(.1)" fill="#fff" textLength="0">
    GNU ELPA</text>
    <text aria-hidden="true" x="-10" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="0">
    consult 0.12</text>
    <text x="-10" y="140" transform="scale(.1)" fill="#fff" textLength="0">
    consult 0.12</text>
  </g>
</svg>
```

I don't know what's wrong.

oantolin:
Mmh. The MELPA badge doesn't have the initial `<?xml version="1.0"?>` but after that looks pretty similar, maybe that's tripping up Chrome.

iquiw:
The SVG has elements with `width="0"`, `textLength="0"` and negative `x`.
There might be something wrong in SVG generation.

oantolin:
Oh, right, @iquiw! I hadn't noticed.

minad:
Yes, the svg generation process is currently broken on ELPA :(

PR #396, opened 2021-11-03
keymap-prompter: fix persistence of digit-argument overriding map

jakanakaevangeli:
This fixes the following bug:

To insert the keybinding of the command next-line:
1) M-x next-line
2) C-. (embark-act)
3) C-u b (where-is with universal argument)
Pressing a number immediately after this will not insert this number, it will
interpret it as a digit argument instead.


oantolin:
Thanks!

PR #397, opened 2021-11-03
Allow binding abort-recursive-edit in embark maps

jakanakaevangeli:
This reorders two clauses in a `cond` with the following consequence: one can
now bind `abort-recursive-edit` to a key like `<escape>` in an embark map and
then use this key to cancel embark.

Current behaviour without this patch is that by pressing `<escape>`, embark
will throw your minibuffer away and literally execute `abort-recursive-edit`
like any other action, which is rarely useful and usually just signals an
error.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -1007,10 +1007,10 @@ UPDATE is the indicator update function."
          (let ((overriding-terminal-local-map overriding-terminal-local-map))
            (command-execute cmd)))
        (embark-keymap-prompter keymap update))
-      ((guard (lookup-key keymap keys))  ; if directly bound, then obey
-       cmd)
       ((or 'minibuffer-keyboard-quit 'abort-recursive-edit 'abort-minibuffers)
        nil)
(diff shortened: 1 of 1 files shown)


oantolin:
OK, sure. If you really wanted to quit the minibuffer, you just cancel twice, that's not too bad.

oantolin:
Oh, wait, I just remembered someone emailed me a patch with exactly this change. I'm not used to dealing with email patches so I put it off and then completely forgot! I owe that person an email.

jakanakaevangeli:
Omar Antolín Camarena ***@***.***> writes:

> Oh, wait, I just remembered someone emailed me a patch with exactly
> this change. I'm not used to dealing with email patches so I put it
> off and then completely forgot! I owe that person an email.

(That was me) Yeah no problem. By the way, in notmuch, I simply put
point on the attached patch and press ". | git am RET", to pipe this
attachment through git am, making a commit. Other mail readers probably
have similar piping command.

Issue #319, opened 2021-07-31
Make embark-verbose-indicator available for C-h 


minad:
For people who like which-key:

~~~ elisp
(defun embark-bindings (&optional prefix)
  "Explore all current command key bindings with `completing-read'.
The selected command will be executed.  The set of key bindings can
be restricted by passing a PREFIX key."
  (interactive)
  (let ((keymap (if prefix
                    (key-binding prefix)
                  (make-composed-keymap (current-active-maps t)))))
    (unless (keymapp keymap)
      (user-error "No key bindings found"))
    ;;(when-let (command (embark-completing-read-prompter keymap nil 'no-default))
    ;;  (call-interactively command))
    (when-let (command (let ((indicator (funcall
                                         (if (eq embark-indicator #'embark-mixed-indicator)
                                             #'embark-verbose-indicator
                                           embark-indicator))))
                         (unwind-protect
                             (embark--prompt indicator keymap nil)
                           (funcall indicator))))
      (call-interactively command))))
~~~

It seems to me that the indicator/prompter distinction is somehow not meaningful. Somehow this feels conceptually off a little bit. Can we unify these indicator and prompter?

oantolin:
Yeah, some unification would be nice. The minimal indicator makes sense for both the keymap prompter and for the completing-read prompter, but the verbose indicator isn't really a good fit for the completing-read prompter.

minad:
Exactly. We should improve this. as you say there are these three combinations:

* completing-read+minimal
* keymap+minimal
* keymap+verbose

But one also has transitions between these three prompters.

minad:
And it should also still be possible to replace the indicator used by the keymap prompter, such that users can write their individual indicators.

Issue #207, opened 2021-04-16
Make config option to hide embark which-key menus, or put in submenu


bdarcus:
Just thought I'd add, for tracking.

Edit: sorry for the confusion this generated. I meant to say this was focused on `which-key` (though of course it relates to the completing-read-based menu system too).

Given embark already supports simple menus, perhaps this could be a defcustom here; something like `embark-act-menu`, with options for "hide", "submenu", and whatever value would yield current behavior?

_Originally posted by @bdarcus in https://github.com/oantolin/embark/issues/197#issuecomment-813024938_

oantolin:
Do we really need a user option for this? Maybe always putting the general actions in a submenu is OK for everyone.

oantolin:
Note that using `@ <keybinding>` still sees the general actions, so once you've memorized their keybindings you can access them from the completing read prompter without going into the submenu.

minad:
To be honest, I am not a fan. 

1. I think it looks ugly.
2. Flat menus work much better with completing-read than nested menus, e.g. consult-imenu/consult-org-heading which are deliberately flat 
3. I don't like the special casing here, I prefer more generic solutions. Is it possible to configure other keymaps as submenus?

I have an alternative proposal: What about using the `group/title-function` and group the actions according to their keymap? The `group-function` is implemented as `x-title-function` in Selectrum, Vertico and Icomplete-vertical as of now. But I am optimistic that it will be accepted soon in Emacs proper.

oantolin:
At any rate, let me emphasize that this general-action-submenu-commit is just a proposal to address this issue, and I most definitely welcome other ideas.

minad:
Yes, I agree with you about searching vs browsing. But if grouping is enabled, the browsing problem will be mostly gone, since then the groups are clearly separate. From my perspective it is better than using a submenu. As I mentioned, I prefer consult-imenu over imenu.

> the embark code base deals with keymap names only at top level and quickly switches to passing actual (nameless) keymaps around. I'm not saying I wouldn't be willing to make the necessary changes, just commenting on them.

Of course, this is a problem. One has to somehow keep the keymap naming information around. We could add "embark pseudo keys" into the keymap which has the grouping information, like "which-key pseudo keys", but I am pretty sure you dislike this since you prefer to keep the keymaps pure ;)

I am unsure about the best way forward here. Maybe it would be a good idea to implement separate `embark-nested-completing-read-` and `embark-flat-completing-read-prompters`. These could probably also share some code?

[...]

PR #401, opened 2021-11-09
Add embark-help-key, remove embark-keymap-help from embark-meta-map

minad:
Treat the help key specially such that it is always available when
acting with the keymap prompter, even if the action keymap does
not contain such a binding. This special treatment is needed to
ensure that the Embark help is always available. It is possible
to disable the help binding by setting embark-help-key=nil. The
embark-cycle-key has a similar treatment and is always available.

cc @roshanshariff

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -216,6 +216,13 @@ This is the key representation accepted by `define-key'."
 If the key is set to nil it defaults to the global binding of
 `embark-act'.  The key must be either a string or a vector.  This
 is the key representation accepted by `define-key'."
+  :type '(choice key-sequence (const :tag "Use embark-act key" nil)))
+
+(defcustom embark-help-key "\C-h"
+  "Key used for help.
+
+The key must be either nil, a string or a vector.  This
+is the key representation accepted by `define-key'."
   :type '(choice key-sequence (const :tag "None" nil)))
 
 (defcustom embark-keybinding-repeat
@@ -889,6 +896,8 @@ If CYCLE is non-nil bind `embark-cycle'."
      (define-key map [13] (embark--default-action type))
      (when-let ((cycle-key (and cycle (embark--cycle-key))))
        (define-key map cycle-key #'embark-cycle))
+     (when embark-help-key
+       (define-key map embark-help-key #'embark-keymap-help))
      map)
    (symbol-value (or (alist-get type embark-keymap-alist)
                      (alist-get t embark-keymap-alist)))))
@@ -1960,11 +1969,14 @@ See `embark-act' for the meaning of the prefix ARG."
 
 (defun embark--become-keymap ()
   "Return keymap of commands to become for current command."
-  (make-composed-keymap
-   (cl-loop for keymap-name in embark-become-keymaps
-            for keymap = (symbol-value keymap-name)
-            when (where-is-internal embark--command (list keymap))
-            collect keymap)))
+  (let ((map (make-composed-keymap
+              (cl-loop for keymap-name in embark-become-keymaps
+                       for keymap = (symbol-value keymap-name)
+                       when (where-is-internal embark--command (list keymap))
+                       collect keymap))))
+    (when embark-help-key
+      (define-key map embark-help-key #'embark-keymap-help))
+    map))
 
 ;;;###autoload
 (defun embark-become (&optional full)
@@ -3323,7 +3335,6 @@ and leaves the point to the left of it."
 (embark-define-keymap embark-meta-map
   "Keymap for non-action Embark functions."
   :parent nil
-  ("C-h" embark-keymap-help)
   ("-" negative-argument)
   ("0" digit-argument)
   ("1" digit-argument)


roshanshariff:
Excellent, thanks! Looks good to me. For the record, the solution I was trying to get across in bdarcus/citar#374 was (in your own words)

> So I think I could agree with a proposal of always activating the meta map and taking it completely out of the inheritance hierarchy. Then the indicators wouldn't also have to bother with the commands registered there.

...and also moving the `digit-argument` binds to `embark-general-map`. Then you could also get rid of the help-key and cycle-key entirely (as well as the slightly magical `embark--cycle-key` function) by doing this:

```elisp
(embark-define-keymap embark-meta-map
  :parent nil
  ("?" embark-keymap-help)
  ("C" embark-cycle)
  ([remap embark-act] embark-cycle))

(defun embark--action-keymap (type _cycle) ; can remove the _cycle parameter later
  "Return action keymap for targets of given TYPE."
  (let* ((action-map (symbol-value (or (alist-get type embark-keymap-alist)
                                      (alist-get t embark-keymap-alist))))
         (map (make-composed-keymap embark-meta-map action-map)))
    (define-key map [13] (embark--default-action type))
    map))
```

I'm not arguing for doing it this way, but I personally find it quite a lot simpler. Now `embark-cycle` and `embark-keymap-help` are very much like regular commands, that I can run with `M-x`. Also running `describe-function` on them tells me where they're bound. It's also very clear from looking at `embark-meta-map` that the `embark-act` key will now run `embark-cycle` instead, and the user can remove or change that if they want. And finally, if I want to bind some keys that are always active whenever `embark-act` runs, I can bind them in `embark-meta-map`.

But anyway, I'm just going to leave this here for future reference.

minad:
@roshanshariff Your solution looks better and I think I almost agree within it. Configuration via keymaps is my preference. However the meta map also acts as the parent keymap of the become keymaps. Furthermore I am not sure where the digit keys should go - the general map does not look like a good place, also because we would lose the keys for become.

Option 1:

1. Add an embark-action-map with cycle and help.
2. Add an embark-become-map with help only.
3. Rename the meta map to digit map (only digit keys). Do not auto activate this map, but use it as parents in the keymap hierarchy.

Option 2:

1. Add an embark-mode-map with help and cycle which is always on for act and become.
2. Rename meta map to digit map and keep it in the hierarchy.
3. For become magically unbind the cycle key?

Note that we also magically bind RET for actions, so we cannot get by without any magic. So I am not sure if the keymap solution is indeed simpler, since we also have to introduce another keymap. 🤷 

I hope Omar comes up with the best solution :)

oantolin:
I don't think I have the best approach yet, but here's an idea:

1. We use `embark-meta-map` as a place to store these special key bindings, as suggested by @roshanshariff. It will be used instead of the variables `embark-help-key` (used in this PR), `embark-keymap-prompter-key`  and `embark-cycle-key`. We don't actually compose the keymap with the action map, since all of the bindings aren't used at once: the keymap prompter doesn't use `embark-keymap-prompter-key`; the completing-read prompter doesn't use `embark-help-key`; becomes doesn't use `embark-cycle-key`. The structure of the code stays the same, we just lookup the necessary key bindings in the meta map instead of in those variables whenever we are about to bind them.

2. For the digit and negative arguments we go ith @minad's suggestion: put them in a new map that isn't auto-activated, it just takes the current `embark-meta-map`'s place in the hierarchy.

minad:
There is an issue though - you have to bind the keys like C-h since otherwise the original C-h prefix map is used and the prompter will not return after pressing C-h. Instead it enters the help map. But since you said the structure stays the same you probably meant that? I kind of like the idea of using the keymap for configuration and not actually using it. It is a bit wild :-P

You could also add a binding of a magic embark-default there to reverse lookup the 13 or alternatively remap exit-minibuffer for additional confusion.

minad:
Hmm, Omar there is another serious issue with your idea of using the map only for configuration. The user may assume that more commands could be bound there which will then be silently ignored.

roshanshariff:
Yes, exactly, I really think the keymaps have to be composed. The composition is just getting around the fact that we only have one `overriding-terminal-local-map`, so we have to do something to put multiple maps on top of the active keymap stack. It's just an implementation detail, but it's very much in the spirit of the built-in Emacs command loop.

You can actually use remap bindings to deal with RET as well, but I was hesitant because it's not a perfect solution. The RET binding is the only situation in which embark looks "inside" the action map for any reason, so some amount of magic is unavoidable anyway. The above is the cleanest solution I could think of, and it's still very small magic :slightly_smiling_face:.

roshanshariff:
May I ask why the composition doesn't appeal to you? I would like to know if it has downsides I'm not aware of.

oantolin:
Silently ignoring other command bound in the map does sound bad, @minad.

I think the only downside of composing the meta map, @roshanshariff, is that that binds all of the keys in it whether they are relevant or not. Unfortunately become, the keymap prompter and the completing read prompter all have slightly different needs.

roshanshariff:
> Unfortunately become, the keymap prompter and the completing read prompter all have slightly different needs.

Oh, good, that's a fine reason, but also easy to solve! (I was thinking about solutions but was hesitant to propose them until I understood whether or not you agreed with the idea in general).

So I think embark-act should put three keymaps on top of the stack: meta-map on top, the action-map in the middle, and general-map at the bottom. These are analogous to before and after advice for functions, they cover all the possible cases.

The next question is what should meta-map and general-map be. For these we just have more keymaps, of which some subset will be composed together on the fly! I would have:
* embark-meta-map, containing the help command and always included in meta-map.
* embark-act-map, used when embark-acting, empty by default 
* embark-become-map, used when embark-becoming, empty by default
* embark-completion-map, for when embark is invoked from a completion session, containing bindings for embark-become, snapshot and collect. (Though I'm not sure about the capabilities of collect, so this may change.)
* embark-buffer-map, for when embark is invoked from a regular buffer, containing embark-cycle.

It doesn't really matter whether these form an inheritance hierarchy or not. I would prefer to just have a function that composes the right ones together as needed.

I know this is a lot of keymaps, so you can see why I was hesitant to say it out loud :slightly_smiling_face:. And you would need a similar set of general-maps, which should contain things that are okay to be shadowed by some action map. (Such as convenience functions, digit-argument bindings, etc. The meta-map bindings should be absolutely minimal.) And, of course, perhaps some combinations don't make sense and can be omitted, but this is the general idea.

Finally, the prompters: As far as I can tell, the problems go away if the prompters have access to the original, unmodified, uncomposed action map. The keymap prompter is in some sense the most important, and it will use the composed keymap above. The other prompters (just the completing-read prompter for now, I guess) can do their own thing as needed (since, of course, it doesn't make sense to bind "?" or "C" or anything when using that prompter).

That leaves the "@" key in the completing read prompter. I think this needs to be implemented by passing control back to the keymap prompter. I thought this when I was looking at that code earlier to implement prefix-help-char. There's no reason to duplicate all that logic. And while the keymap prompter is active, the completing read prompter just acts as an indicator, narrowing the displayed candidates based on what's pressed. This also fits in with the prompter/indicator unification you've mentioned before.

I know all this seems like a big change, but I really do think it'll greatly simplify the Embark codebase. And I'm very eager to help implement things as soon as I get my FSF papers ready.

oantolin:
That is a lot of keymaps! :)

And spending on what goes into the general map it's either a loss of functionality or a loss of flexibility too: if it doesn't include the things it currently does like saving to the kill ring or insert, it's definitely a loss of functionality (I use those actions from both the minibuffer and regular buffers many times a day); if does contain the usual suspects, some users su H as the citar package will want to disable the inclusion of for example, is not wanted in citar.

oantolin:
I do agree that the @ key in the completing-read prompter should call the keymap prompter and have no idea why I didn't do it that way to begin with.

oantolin:
> As far as I can tell, the problems go away if the prompters have access to the original, unmodified, uncomposed action map

I'm getting a little lost: which problems go away when you do that?

roshanshariff:
I think a loss of functionality is unacceptable, and the large number of keymaps is precisely so there is a reasonable place to put everything. So if you (either as the Embark developer or an Embark user) want to bind something, you have to ask yourself a few questions, and you will always find the appropriate keymap to bind it in:
1. Is it acceptable for this binding to be shadowed by some action map. If yes, it should go in the general-map side, otherwise in the meta-map side. You've said that your workflow depends on Collect, for example, so that should definitely go in the meta side.
2. Does this action make sense only during completion, only outside completion, or both? If both, put it in one of the act-maps, otherwise either in completion-map or buffer-map.
3. Does this make sense only for embark-act, or embark-become, or both? This is a different kind of dichotomy, as in the "become" case there are no targets and you're always in a completion session, so the number of choices is much smaller.

> which problems go away when you do that?

I'd like to answer this question at length, so please indulge me :)

You've alluded to this before, but the idea of a prompter is not *really* exclusive to Embark. So, for example, the keymap prompter is just an implementation of the built in Emacs command loop, whereas the completing-read prompter is a reimplementation of M-x completion. (Let's ignore indicators for now, because perhaps the distinction is irrelevant and they fit in naturally.)

So the main question: what is the extra ingredient that makes an embark prompter different from those two things? The emacs command loop and the M-x completer only know that the user is going to select some Emacs command, either via its keybinding or by name. The Embark prompter, on the other hand, knows that there is some small set of privileged commands (those in the target/become keymaps) in addition to the entire set of emacs commands.

So how do these prompters use this extra knowledge? The keymap prompter doesn't do anything special, just binds the embark actions on top of the stack and calls read-key-sequence. The completing-read prompter should behave just like M-x completion, with the embark actions added to the set. And, importantly, it starts off narrowed to those actions (and perhaps this means there should be a way to widen it to include all commands).

The prompter also knows that not all the embark actions are equally relevant. Some of them come from the meta/general maps, and others come from the become/target maps. Again, the keymap prompter doesn't really care about this. But the completing-read prompter can either narrow to the union of these three maps (for new users) or just the become or target-specific actions (if you're familiar with Embark). In principle, the completing-read prompter can have three levels of narrowing: all emacs commands, meta+action+general commands, or just action commands. And to do this the prompter needs to have those keymaps available separately, not combined into one undifferentiated keymap.

Of course, the contents of the meta and general maps can be obtained by the prompter independently, since they don't depend on the particular target type or become-group. But the target/become maps do need to be passed into the prompter by embark-act/embark-become.

So that's why I said that the prompter only really needs access to the target/become keymaps as part of the prompter protocol! *whew* :grin:

minad:
Oh no no no! Sorry for jumping into the parade again!

@roshanshariff This PR solves the exact problem we are having of entirely decoupling the action maps (which are up to the user) from the fundamentally required commands (help and cycle). I suggest we go with my proposal now. It is not that bad actually, since we had used the "magic solution" for the cycle key before. Now we only add another "magic solution" for the help key.

You instead propose to add multiple new keymaps. This is NOT a simplification. Maybe it is conceptually simpler since we only deal with keymaps, but please keep the other criteria in mind. Does it make the code base simpler? Does it make the code base shorter? Very primitive complexity measures can also help for judgement. Let us not go into direction of overengineering!

Now you are mixing a prompter discussion in. Of course all these things are mixed together and there is some cleanup. I've also opened an issue about unification of prompters and indicators (https://github.com/oantolin/embark/issues/319). I would appreciate it we take smaller steps here. Embark is used in the wild and people rely on it, we should not just refactor it for the sake of some minor architectural annoyance.

minad:
@roshanshariff I must also say that I am getting annoyed by your approach here.

> (I was thinking about solutions but was hesitant to propose them until I understood whether or not you agreed with the idea in general).

Of course I am not the one making the decision here. But this is not how decision making or a discussion should work. If there is some agreement on some specific point it does not mean that there is general agreement on something. I am only talking about myself - but I would not unconditionally agree on the keymap solution no matter the costs. I only agree that it looks nicer but I immediately pointed out deficiencies with it leading to MORE complexity. You said yourself that proposals should be laid out more clearly in the citar discussion. Please let's do that and split up the issues, create separate discussions with small steps. This also ensures that we don't overthrow everything.

I find it very important now to not overthrow everything since Embark is somehow proven already. Of course there is some personal development or development of perspective involved here. When I started using Embark and when Omar and I started to work on fitting Marginalia/Embark/Consult/Selectrum together there was a lot of discussion of how Embark is architected and there was a lot churn in the following months. But even worse when I started with Embark I had many architectural concerns with it - I disagreed with many of the choices Omar made here. I wasn't sure if completion categories are indeed the right classification criterion to bind the actions. I mostly focused on the minibuffer and disliked the all encompassing nature of Embark which in my opinion complicated things. I was proposing numerous times of splitting up Embark in more fine grained components and so on. Omar said no to most things and I am glad he did since I think the current state of Embark is pretty good and many of the design decisions make a lot of sense now.

minad:
While I think we should just get over this discussion, I want to single out one specific proposal you've made here: You proposed the introduction of multiple maps (embark-meta-map, embark-act-map, embark-become-map, embark-completion-map, embark-buffer-map).

Here your goal is to achieve full composability at the level of keymaps. But this is unnecessary since we can already recompose and mix on the level of commands. If you disagree with a certain keymap, just create your own and bind only the commands you are interested in. 

Furthermore I argue that achieving full composability at the level of keymaps is undesirable since it adds one more indirection and complexity. I am sure most users are aware on how to bind keys, if we keep it to that it is better than telling the user that they have to remix a bunch of keymaps. Furthermore keymaps are also a unit to navigate the project. If we split the up on a too fine grained level we lose that advantage. 

As I see it Embark does not aim for the perfect keymaps (which fit precisely to the context) but it approximates a bit. This makes sense since all the other keybinding are still available C-x/M-x/... The approximative nature of the Embark keymaps is done for the sake of simplicity and to avoid redundancy. For example you proposed an embark-buffer-map here. The natural conclusion then would be to also separate the minibuffer and buffer maps for the actions and so on and so forth. While this may make sense at first, it is such a complication that one loses the introspection aspect somehow. We will get a tangled mess of keymaps. For me Emacs is about having access to the inner workings on the Elisp level - I see the Elisp as part of the interface to the user and this is what leads to the appeal for me. So as I see it - Embark is over-unifying a bit for the sake of a more concise code base.

roshanshariff:
> I must also say that I am getting annoyed by your approach here.

I'm terribly sorry, and I sincerely apologize! My intention was just to get feedback into a particular direction before going deeper into details in that direction. So, for example, in the beginning I presented a low-complexity piece of code with (as I see it) low cost, but it might well not have been acceptable to you or @oantolin.

Then I outlined a different approach, with a different cost-benefit tradeoff. Again, I wanted to get a sense of what you think about this. And I actually agree with you that perhaps that goes too far, trading off one kind of complexity for another kind. Now knowing this, I have a better idea of what you value that I'll take into account in the future.

Of course, at any point you or @oantolin are free to shoot down any of these proposals for any reason, or change your mind after you've tentatively agreed to some point. I appreciate the time you've put into maintaining this very nice software and am hesitant to impose on it any further.

And I realize that embark, being widely used and relied-on software, cannot be re-architected on a whim (that's not how you get software that lasts 40 years). So of course, even if some architecture seems abstractly desirable, that doesn't mean there's any reason to go and rewrite everything or even anything. But I still think it's useful to have these sorts of discussions (and in public) so there's a shared understanding of the goals and overall structure of the project. I know I learned a lot by reading old issues and pull requests. But before one actually writes or changes code, of course we must be very conservative and take small steps.

Thanks again @oantolin and @minad for engaging with me so constructively, and I apologize again for overstepping here.

minad:
> I'm terribly sorry, and I sincerely apologize! 

No need to apologize! I should rather apologize for taking such a defensive/conservative stance here and shooting too much against your ideas! This is not the picture I want to paint - it is great if you come with new ideas and I don't want to discourage you.

All I am asking for is to take smaller steps and not mixing together orthogonal discussions. Everyone does that occasionally and it is totally naturally to do this if you have some stream of thought. It is just my experience that we can reach the goal more quickly if we agree on small things, implement them, get over it and revisit later. You can very well rebuild a better more overarching architecture on top of a small-step solution as proposed in this PR.

While I tried to bring across my perspective of how I see Embark and its architecture - I would like to make it clear that I speak for myself. Omar may have a different perspective!

> And I realize that embark, being widely used and relied-on software, cannot be re-architected on a whim (that's not how you get software that lasts 40 years). So of course, even if some architecture seems abstractly desirable, that doesn't mean there's any reason to go and rewrite everything or even anything. 

True. Also Embark has to make compromises to unify and make many Emacs features accessible in its framework. The compromise is at the cost of the cleanliness of the menus (in the Embark default configuration). But as a user you are of course free to change this. In particular if this PR gets merged - the full freedom is available to the developer of Embark keymaps.

> But I still think it's useful to have these sorts of discussions (and in public) so there's a shared understanding of the goals and overall structure of the project. I know I learned a lot by reading old issues and pull requests. But before one actually writes or changes code, of course we must be very conservative and take small steps.

Of course! However in these long discussions a lot of things get lost unfortunately. Therefore I am so much in favor of doing the small steps and also splitting up the discussions in separate issues.

oantolin:
Woah! Lots happened overnight (it's early morning where I am)!

I had held off on merging this in case we quickly discovered a close-to-perfect solution, but it's clear now this requires more discussion. The nice thing about software is you can always change it again later. So I'm merging this now, but I'd like it if we continue discuss other options too. 

With this change the keymaps in Citar don't need to inherit from anyone and can be kept as something you might want to bind to a prefix key. The ability to make a standalone keymap seems pretty desirable and we already achieve it with this, and in the future it looks like we might be able to simplify the Embark source without losing this property.

After merging, I'll read everyone's comments and respond.

oantolin:
So first of all let me say, @roshanshariff, that proposals are very welcome in general and that you don't need to apologize for "overstepping" just for proposing something, that's definitely not overstepping! At the same time, it's wise to expect some pushback. :)

oantolin:
Now let me say something about the big picture here. I think the main goal of the changes we are discussing is to have fewer key bindings available, right?, so that we only have bindings for the most relevant commands? But I don't view having a binding for a command I won't use as a terrible thing: it happens absolutely all the time in Emacs! What fraction of the default key bindings do I use? It's probably a pretty damn small fraction!

I think the only reason one might want to have _fewer_ key bindings is if one is a which-key user (or also, in Embark, a verbose indicator user). :P In my opinion this "problem" of having too many key bindings is adequately solved by simply not looking at them most of the time, and using the completion system to narrow a list of key bindings when you do need to look at them. For example, it used to bother me than Emacs put rectangle, bookmark and register commands all under the same prefix, `C-x r`. But that was when I briefly tried which-key, or when I used the built-in `C-h`; now with `embark-prefix-help-command`, if I forget a binding I press `C-h` and type `rec`, `bo`, or `reg`.

At the same time I understand that which-key is very popular, that it serves a real purpose in aiding discoverability, and I even think it's a good idea to have the mixed indicator on by default (just like I think it's good that Emacs's menu bar is on by default, but I personally turn it off). Now, which-key users have been dealing with this too-many-bindings problem (which, I repeat is actually _caused_ by which-key :P) for longer than Embark users, and we should take advantage of the lessons they've learned. So, how do which-key user reduce clutter: do they unbind keys or do they keep the bindings and just filter them out from the which-key display? I suspect filtering is the preferred solution.

oantolin:
So, I think we only really need to hide the unwanted bindings in the UIs that display them: the completing-read prompter, the verbose indicator and the which-key indicator. Maybe a single exclude list shared by all three would be enough (such a list already exists in `embark-verbose-indicator-excluded-actions` which could be renamed to something like `embark-hidden-actions`).

I think having the keymap-prompter stich together two or three different keymaps just so some things are unavailable complicates the docstrings without a commesurate benefit. But maybe the circumstances under which different keymaps would be activated are not as complicated as I think. Probably to evaluate the complexity of the many-keymap approach correctly, we'd need a more detailed description of it.

oantolin:
There also the matter of the special key bindings, which are just 3, I think:

- for switching from the keymap prompter to the completing-read prompter,
- for switching from the completing-read prompter to the keymap prompter,
- for cycling between targets.

No two of them are used in the same circumstances, so if they were to fit into the many-keymaps proposal, I think they'd have to be in 3 separate keymaps, each of which would probably only contain that single keybinding, right? So it seems about the same as having three variables to customize the key bindings.

oantolin:
And let's look at the general actions:

- `delete-region` and `mark`: useful for all buffer targets that report bounds
- `embark-isearch`: probably only useful for buffer targets that appear literally in the buffer (I don't think if point is on `foo` in `#include <foo>` I really want to search for occurences of the *full path to* `foo`)
- `embark-become`: only useful in the minibuffer (whether in completion or not)
- `embark-collect-snapshot`, `embark-collect-live` and `embark-export`: mostly useful in minibuffer completion, can be used in regular buffers.
- `kill-new`, `embark-insert`: super useful always, I've only heard one claimed counterexample: Citar!
- `consult-line-multi`, `consult-line`, `consult-imenu-multi`, `consult-imenu`, `consult-outline`, `consult-locate`, `consult-find`, `consult-git-grep`, `consult-ripgrep`, `consult-grep`: Same as `embark-isearch`, probably only useful for buffer targets that appear literally as text in the buffer.

So I think if we wanted to segregate these it would be tricky, we'd need keymaps for:

- buffer targets with bounds
- buffer targets that appear literally in the buffer
- the minibuffer, and maybe a separate one for minibuffer completion

I guess that's doable, but I'm not sure it's worth it.

minad:
> So I think if we wanted to segregate these it would be tricky, we'd need keymaps for: buffer targets with bounds, buffer targets that appear literally in the buffer, the minibuffer, and maybe a separate one for minibuffer completion. I guess that's doable, but I'm not sure it's worth it.

I would like to point out that we also have hooks to transform the targets or setup the minibuffer for the action. So this makes it possible to prepare targets for literal searches. This should alleviate the need for an overly fine grained keymap hierarchy. I kind of like that Embark over generalizes a bit and is conceptually a bit simpler in the end.

As probably became clear in my responses here, overall I agree with you @oantolin that this is first and foremost an indicator issue (https://github.com/bdarcus/citar/pull/374, https://github.com/bdarcus/citar/issues/389, https://github.com/oantolin/embark/issues/207). The issue is in particular relevant for the which-key and verbose-indicator. While these indicators are great for discoverability they are lacking as soon as you are actually searching for something. Fixing this from the side of the keymap is starting at the wrong end. Of course discussing and refining the overall keymap hierarchy can be discussed independently!

Furthermore the indicators or the Embark design is not that great if your goal is to create clean hydra/transient-like menus. So I think this is something which is also worth some consideration - why not reuse the transient/hydra UI to create *yet another indicator* which is Citar specific? Fundamentally I think Transient and Embark sit at different sides of the spectrum - either you optimize a fixed UI (Transient) or you optimize the flexibility regarding remixing/reusing existing commands as actions freely (Embark).

bdarcus:
I did play with transient earlier, but decided to drop it, possibly for someone else to pick up if they had interest.

https://github.com/bdarcus/citar/pull/186

I think with this change which-key should be (I haven't tested it yet) fine though.

I also mentioned on that thread this package, though I haven't gotten around to testing it:

https://gitlab.com/jjzmajic/hercules.el

minad:
> https://gitlab.com/jjzmajic/hercules.el

I played with hercules for a while when I still used which-key. From my perspective these action-related UIs can be categorized as follows:

1. Transient, Hydra: Leans towards fixed keymaps/bindings, optimized formatting of the UI. The formatting is up to a certain extent automatic but can be influenced by the author of the Transient.
2. Embark verbose indicator, which key: Leans towards arbitrary keymaps, potentially very cluttered. The priority is to allow easy modification of the keybindings. Fully automatic formatting. No user intervention needed. Unfortunately the user also has only limited influence on the formatting.
3. Hercules: Reuses which-key for transient keymaps. Sits between the two other types of UIs and from my perspective it combines the disadvantages of both. Fixed keymaps, no flexibility, limited possibility to influence the formatting.
4. Embark completing-read prompter, embark-prefix-help-command, helm-descbinds, counsel-descbinds: Reuses completing-read for the UI.

Of course hydras/transients can also be adjusted but the intention generally is to offer a fixed set of actions. The Embark, which-key and completing-read UIs in contrast rather optimize for an arbitrary and extensible set of actions. This sounds maybe a bit fuzzy and there is certainly overlap.

roshanshariff:
Thanks for your comments @oantolin! I'm fully expecting (and even hoping for) pushback to anything I propose, because of course there are often big flaws I didn't foresee. The main takeaway I got from @minad was to try to keep discussion focused, and to use the appropriate issue/PR (or open a new one) to discuss things, which I will try to be better at in the future. And the other takeaway is that it's always good to present at least a small code prototype to illustrate proposals, so we understand what exactly we're talking about.

> Probably to evaluate the complexity of the many-keymap approach correctly, we'd need a more detailed description of it.

Yes, I agree, I will try to come up with a small prototype when I have time, just to discuss the idea and with no expectation that it'll actually be used.

> Maybe a single exclude list shared by all three would be enough

Yeah, I think this is the best medium-term solution. Currently the verbose indicator does filtering on formatted text, but instead if we have a function that does filtering on keymaps (so produce a new keymap with some bindings removed) then it can be used by both the verbose and which-key indicators.

> first and foremost an indicator issue

In the longer-term, I think the prompter/indicator protocol can be improved so that less/no filtering is needed; the goal is to make small modifications to Embark that allow indicators to better deal with the issue. And this goal doesn't need fewer keys to be bound, just for the indicators to be given a little more information. But I will resume the discussion on this when I have a prototype to point to. And, of course, I might encounter problems or find that the prototype has unnecessary complexity, in which case I will drop the idea.

And, secondly, there is one case where it's not just an indicator issue to minimize the number of bindings: when those bindings shadow. That was the original impetus for my many-keymaps proposal: that short snippet of code I gave in comment 1 unnecessarily blocked the `embark-cycle` key from being used when running `embark-become`, even though there's no cycling in that context. But perhaps the many-keymaps idea was too drastic. Again, I will discuss this further when I have a prototype that alleviates the problem in a nice composable way without quite that many keymaps :)

PR #404, opened 2021-11-14
Remove from history and minor changes

minad:
I know you don't use UIs which sort, but for some people this matters. ;)

Sometimes I accidentally execute some command which is then sorted closer to the top. I added a command `embark-history-remove`, which allows to remove the candidate from the minibuffer history. This command is similar to `embark-recentf-remove`. I wasn't so bold to put this command into the general map. Instead I put it into the symbol map, but one could be even more conservative and only put it in the command map. I am mostly bothered by this for M-x.

I changed the keybindings to backslash, which is better since it avoids the conflict with the reserved "-" binding. Furthermore I feel the `embark-recentf-remove` and `embark-history-remove` commands are not important enough to put them on "-".

If you are not convinced - a more important use case: If you played `tetris` for a while, it is better to remove the command afterwards from the history. The next time you demonstrate something with your Emacs everybody is convinced that you use the editor only for serious work. Unfortunately this problem is more severe for poor guys who use in-your-face UIs like Vertico. Damn.

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -322,6 +322,7 @@ that is a Consult async command."
   "Collect all imenu items in the current buffer."
   (cons 'imenu (mapcar #'car (consult-imenu--items))))
 
+(setf (alist-get 'imenu embark-default-action-overrides) #'consult-imenu)
 (add-to-list 'embark-candidate-collectors #'embark-consult-outline-candidates 'append)
 
 (provide 'embark-consult)
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -317,13 +317,10 @@ that is a Consult async command."
   "Collect all outline headings in the current buffer."
   (cons 'consult-location (consult--outline-candidates)))
 
+(autoload 'consult-imenu--items "consult-imenu")
 (defun embark-consult-toc-imenu ()
   "Collect all imenu items in the current buffer."
-  (cons 'consult-location
-        (mapcar (pcase-lambda (`(,item . ,pos))
-                  (propertize item 'consult-location
-                              (cons pos (line-number-at-pos pos))))
-                (consult-imenu--items))))
+  (cons 'imenu (mapcar #'car (consult-imenu--items))))
 
 (add-to-list 'embark-candidate-collectors #'embark-consult-toc-outline 'append)
 
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -313,22 +313,19 @@ that is a Consult async command."
 
 ;;; Tables of contents for buffers: imenu and outline candidate collectors
 
-
 (defun embark-consult-toc-outline ()
-  "Collect all outline headings in the current buffer." 
+  "Collect all outline headings in the current buffer."
   (cons 'consult-location (consult--outline-candidates)))
 
 (defun embark-consult-toc-imenu ()
-  "Collect all imenu items in the current buffer." 
+  "Collect all imenu items in the current buffer."
   (cons 'consult-location
         (mapcar (pcase-lambda (`(,item . ,pos))
                   (propertize item 'consult-location
                               (cons pos (line-number-at-pos pos))))
                 (consult-imenu--items))))
 
-(unless (memq 'embark-consult-toc-outline embark-candidate-collectors)
-  (setq embark-candidate-collectors
-        (append embark-candidate-collectors '(embark-consult-toc-outline))))
+(add-to-list 'embark-candidate-collectors #'embark-consult-toc-outline 'append)
 
 (provide 'embark-consult)
 ;;; embark-consult.el ends here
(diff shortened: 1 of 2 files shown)


oantolin:
This all looks good, except possibly for using the `imenu` category. You lose the occur export (though I guess I could implement that).

minad:
> This all looks good, except possibly for using the |imenu| category. You
> lose the occur export (though I guess I could implement that).

Why do you want an occur export? The imenu is not about matches found in
a file and not about lines. For example writing in such an imenu buffer
won't work. If we really want occur we should also change the category
in consult-imenu from imenu to consult-location. But then we lose the
annotators and potential imenu specific actions. Currently there are no
imenu specific actions, but there could very well be some. In Marginalia
the imenu candidates are only annotated in emacs-lisp-mode.

oantolin:
Using the imenu category is a little worse than that: pressing RET on the items doesn't actually jump to them! (You didn't test it, did you? :P). I remember running into this problem before, which is why I decided to turn them into `consult-location`s instead. Another way to fix it would to add a default action override for the `imenu` category, but setting it to `consult-imenu` would mean embark-collect-snapshot wouldn't work from `M-x imenu`, and viceversa.

I do agree with your reasons for changing it, and think we can do without the occur export (and thus without the slow-to-compute line numbers), but we should fix this. Maybe the best way is to just post-process `consult-imenu--items` to put the position marker into a text property, and make a new category with a default action override.

By the way, in the commit message you said "Preserve the `consult-imenu` actions and annotations", and certainly the annotations are great, but which actions did you mean? I don't think we have any special actions for them, do we?

oantolin:
> I know you don't use UIs which sort, but for some people this matters. ;)

I actually do now, I've been using `embark-sorted-minibuffer-candidates` instead of `embark-minibuffer-candidates` for a few weeks.

oantolin:
Sorry, I wrote some comments above before noticing you had replied: I agree the occur export is wrong (occur-edit-mode not working convinced me), and you answered my question about imenu actions: there are currently none, but we may want to add them in the future.

minad:
> By the way, in the commit message you said "Preserve the consult-imenu actions and annotations", and certainly the annotations are great, but which actions did you mean? I don't think we have any special actions for them, do we?

I said potential actions ;)

> Using the imenu category is a little worse than that: pressing RET on the items doesn't actually jump to them! (You didn't test it, did you? :P). I remember running into this problem before, which is why I decided to turn them into consult-locations instead. 

Indeed. I obviously did not. But I tested the outline toc! :) Now I understand why you attach the location information, sorry!

> Maybe the best way is to just post-process consult-imenu--items to put the position marker into a text property, and make a new category with a default action override.

Makes sense! The imenu candidates in consult have the format of usual imenu entries (cons cells). So maybe it is time to change this now. There is no reason in keeping this format, since imenu entries are not consumed anywhere else in this format.

oantolin:
You know, actually adding `(setf (alist-get 'imenu embark-default-action-overrides) #'consult-imenu)` isn't so bad: it works fine for the imenu TOC, and for `consult-imenu`, of course, but it's even sort of OK with imenu --> embark-collect-snapshot. It might get confused if a function and variable have the same name, but it's not like using `imenu` as the default action for `M-x imenu` is perfect either: the `*Rescan*` and `Variables` items don't work.

Maybe we should just give `imenu` and `consult-imenu` different categories? We can simply use the same annotator for both, and use embark-default-action-overrides for just `consult-imenu`.

oantolin:
Another solution that might be worth considering is just dropping the imenu TOC. Maybe outline is just better?

minad:
Yes, I am thinking the same - I don't want to spent time on this now. Imenu is pretty complicated actually, you have to use `imenu` or `consult-imenu--jump` to jump to the item since there can also be special imenu items which call functions.

oantolin:
.Sorry for suggesting all these slightly different ideas, but here's another: if either consult-imenu or this candidate collector puts the location information in a text property, we can keep the same category for `imenu` and `consult-imenu`, but add a default action override that checks for the text property and use it if found, and if not knows it should call imenu on the entry. We can even be able to arrange for the minibuffer to not quit (use that `unread-events` thing instead), so that `*Rescan*` and `Variables`, etc., let you keep typing.

minad:
Oh, I didn't read your answer before - so yes if you reroute through `consult-imenu` it should work well. This is the usual Embark approach without touching internal structure of the candidate. The downside is only that it will be slow since the imenu is regenerated every time. But Consult caches the imenu, so we are good!

minad:
I added the consult-imenu overwrite for now. Is this PR okay like this?

oantolin:
Yes, this seems fine for now. I'll merge instead of holding up the "remove from history" bit just because I'm slightly uncertain about the imenu candidate collector. That can be fixed later. Plus, we'll see if anyone actually uses `embark-collect-snapshot` with `imenu`. I suspect the answer is no: everyone using embark prefers `consult-imenu`.

minad:
Thanks!

> I suspect the answer is no: everyone using embark prefers |consult-imenu|.
Probably, since this only matters as long as you have embark-consult
installed.

PR #409, opened 2021-11-17
Remove occur-highlight-regexp

minad:
This variable is neither present on Emacs 27 nor 28.

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -155,8 +155,7 @@ The elements of LINES are assumed to be values of category `consult-line'."
             (setq last-buf this-buf))
           (insert (concat lineno contents nl))))
       (goto-char (point-min))
-      (occur-mode)
-      (setq-local occur-highlight-regexp "^.*$"))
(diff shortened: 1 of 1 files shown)


oantolin:
I think that's there for a good reason, I'll try to remar what it's for and comment instead.

minad:
The variable does not exist, I think it has been removed from Emacs 28.

oantolin:
I found it [Emacs master](https://github.com/emacs-mirror/emacs/blob/3af9e84ff59811734dcbb5d55e04e1fdb7051e77/lisp/replace.el#L775), and in my Emacs, but not in the Emacs 26, 27 or 28 branches of that mirror. Weird. I guess I'll just set in my personal configuration instead, and remove it once I update Emacs.

minad:
It is also not present in Emacs master and the upcoming Emacs 29. I looked at the git log and it seems that this variable had been introduced for a short time during the development of Emacs 28 and then removed again after a while.

PR #410, opened 2021-11-17
embark-export-customize-variable: Check if the variable can be properly deserialized

minad:
Currently I get an error, when exporting `corfu--frame`, since this is a `#<frame ...>` object. Maybe you have a better idea on how to perform this check?


oantolin:
That's the only approach I can think of.

PR #411, opened 2021-11-17
Instead of a serialization check, overwrite `read`

minad:
I thought again about #410. It seems to me that the approach used here is better. It is more performant and variables are not filtered out unexpectedly. But it seems that the effect is that nonserializable variables are displayed as "EDITED", so maybe this has other bad side effects.

(EDIT: Downside is now that we cannot save the widget view anymore because of this edited display. Hmm.)

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -2904,16 +2904,22 @@ PRED is a predicate function used to filter the items."
 
 (defun embark-export-customize-variable (variables)
   "Create a customization buffer listing VARIABLES."
-  (embark--export-customize
-   variables "Variables" 'custom-variable
-   (lambda (sym)
-     (and (boundp sym)
-          (condition-case nil
-              ;; Check if variable can be properly deserialized.
-              ;; The customization widget relies on this.
-              (let ((val (symbol-value sym)))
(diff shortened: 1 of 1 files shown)


oantolin:
Woah, how does this work? I thought the problem was that read reported an error and you fix that by ... reporting an error?

minad:
Please give it another look. It is quite a hack, maybe we should just keep the filtering. But I made a mistake in #410, instead of t it should be error - I always forget the syntax of condition-case.


~~~ elisp
(condition-case nil
    ...
  (error nil)) ;; error instead of t here
~~~~

oantolin:
I think this is probably as good as it gets. If you edit one of these unreadable values by accident you can still  try to undo the edit (by memory, because the customize buffer does seem to store undo information!) and it gets set to the value it already had. In the common case where you don't even attempt to modify those values, this is perfect.

oantolin:
Thanks for figuring this out!

oantolin:
Gotta love dynamic scope, huh? Whenever people say Emacs should be written in Python or JavaScript my first objection is always dynamic scope and the advice system: they're indispensable.

PR #416, opened 2021-12-09
Add heading target and keymap based on outline mode

minad:
We could also reuse the outline-mode-prefix-map directly instead of the embark-heading-map. However Embark mostly uses keys without modifiers.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -1905,7 +1905,9 @@ target."
                            new-targets
                            (or (cl-position-if
                                 (let ((desired-type
-                                       (if (eq action 'mark)
+                                       ;; TODO Introduce customizable variable,
+                                       ;; instead of hard-coding the mark commands.
+                                       (if (memq action '(mark outline-mark-subtree))
                                            'region
                                          (plist-get (car targets) :type))))
                                   (lambda (x)
--- a/embark.el
+++ b/embark.el
@@ -500,7 +500,8 @@ arguments and more details."
 
 (defcustom embark-repeat-actions
   '(embark-next-symbol embark-previous-symbol backward-up-list
-    backward-list forward-list forward-sexp backward-sexp mark
+    backward-list forward-list forward-sexp backward-sexp
+    mark outline-mark-subtree
     outline-next-visible-heading outline-previous-visible-heading
     outline-up-heading outline-forward-same-level outline-backward-same-level
     outline-show-subtree outline-hide-subtree outline-cycle
--- a/embark.el
+++ b/embark.el
@@ -504,6 +504,7 @@ arguments and more details."
     outline-next-visible-heading outline-previous-visible-heading
     outline-up-heading outline-forward-same-level outline-backward-same-level
     outline-show-subtree outline-hide-subtree outline-cycle
+    outline-move-subtree-up outline-move-subtree-down
     transpose-sexps transpose-sentences transpose-paragraphs
     forward-sentence backward-sentence forward-paragraph
     backward-paragraph)
--- a/embark.el
+++ b/embark.el
@@ -499,16 +499,25 @@ arguments and more details."
                "0.12")
 
 (defcustom embark-repeat-actions
-  '(embark-next-symbol embark-previous-symbol backward-up-list
-    backward-list forward-list forward-sexp backward-sexp
-    mark outline-mark-subtree
+  '(mark
+    ;; outline commands
     outline-next-visible-heading outline-previous-visible-heading
-    outline-up-heading outline-forward-same-level outline-backward-same-level
-    outline-show-subtree outline-hide-subtree outline-cycle
-    outline-move-subtree-up outline-move-subtree-down
+    outline-forward-same-level outline-backward-same-level
+    outline-demote outline-promote outline-mark-subtree
+    outline-show-subtree outline-move-subtree-up outline-move-subtree-down
+    outline-up-heading outline-hide-subtree outline-cycle
+    ;; org commands (remapped outline commands)
+    org-forward-heading-same-level org-backward-heading-same-level
+    org-next-visible-heading org-previous-visible-heading
+    org-demote-subtree org-promote-subtree org-mark-subtree
+    org-show-subtree org-move-subtree-up org-move-subtree-down
+    ;; transpose commands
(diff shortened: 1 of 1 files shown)


minad:
> Isn't outline-regexp also from outline? If that variable is available then outline has already been loaded, hasn't it?

Good point. I misremembered this and thought that the variable is always defined via a crazy autoload. But I just checked and this is not the case. I used the same trick as in Consult now.

oantolin:
This looks good to me! I have some stuff to do right now, but tomorrow morning (Mexico City time) I'll test and merge.

minad:
Great, thanks! You may want to look into some of the other outline actions bound in the outline-mode-prefix-map for showing and hiding bodies/leaves/children. I have only bound the subtree commands since I think hiding on a finer level is confusing. But there is also some value in replicating the outline-mode-prefix-map exactly (except for the C- modifiers).

hmelman:
> I have no idea why I keep leaving trailing whitespace all over. I guess I never pressed RET on that line?

I've been doing similar lately.  See the Emacs manual node "Useless Whitespace" for some things emacs can do to help.

oantolin:
@minad

> You may want to look into some of the other outline actions bound in the outline-mode-prefix-map for showing and hiding bodies/leaves/children. I have only bound the subtree commands since I think hiding on a finer level is confusing. But there could also be some value in replicating the outline-mode-prefix-map exactly.

I think starting small is better, this looks like a good selection of commands. We can add more later if they are requested.

I have a few questions:

1. Should `outline-move-subtree-up` and `outline-move-subtree-down` be repeatable? The transpose commands are repeatable and these feel similar.

2. For some reason the motion commands (next/previous heading, forward/backward same level) don't repeat in Org mode! It's weird, they do correctly repeat in Emacs Lisp buffers, for example.

3. It would be nice if the `outline-mark-subtree` action got the same treatment as `mark`: that embark-act is kept alive and it automatically cycles to the region target. This requires changing `embark--act`, since that behavior for `mark` is hard-coded. Maybe instead of selecting the next target via`(if (eq action 'mark) 'region ...)` we could use `(if (use-region-p) 'region ...)`?

oantolin:
Thanks for the tip, @hmelman. I think I might try `(setq show-trailing-whitespace t)` for a few days to see if I can get used to it.

minad:
> Should outline-move-subtree-up and outline-move-subtree-down be repeatable? The transpose commands are repeatable and these feel similar.

Good idea.

> For some reason the motion commands (next/previous heading, forward/backward same level) don't repeat in Org mode! It's weird, they do correctly repeat in Emacs Lisp buffers, for example.

Indeed. I see this too. This seems like a bug in some other part of Embark. Do you have an idea what causes this? (However we should probably provide embark-org targets later on, then we will get an org-heading target and the embark-org-heading keymap will inherit from the embark-heading keymap).

minad:
> It would be nice if the outline-mark-subtree action got the same treatment as mark: that embark-act is kept alive and it automatically cycles to the region target. This requires changing embark--act, since that behavior for mark is hard-coded. Maybe instead of selecting the next target via(if (eq action 'mark) 'region ...) we could use (if (use-region-p) 'region ...)?

This sounds reasonable. I remember that we discussed about this a while ago and I think we even discussed the possibility to detect the region. But in the end you went with the explicit mark check. I don't recall the reasons :thinking:

oantolin:
I'll investigate the Org heading problem.

In the meantime I did some light testing on providing the automatic cycling to region targets for any action that is repeatable and that leaves an active region (just using `(if (use-region-p) 'region ...)` as I said above.). It seems to work well.

oantolin:
Mmh. The problem with detecting the region is that some people might now use transient-mark-mode...

oantolin:
> But in the end you went with the explicit mark check. I don't recall the reasons 🤔

I don't remember either.

minad:
> Mmh. The problem with detecting the region is that some people might not use transient-mark-mode...

You could check if the `(mark)` is non-nil. But there may be other bad effects? I don't mind for now if we would hard code it, or introduce an `embark-mark-actions` variable.

oantolin:
I think `(mark)` is basically always non-nil. It's only nil if in a given buffer you have never set a mark!

oantolin:
Maybe we could add a general way for actions to tell Embark what the next target should be? It would be good for marking, but probably also for other things.

minad:
> I think (mark) is basically always non-nil. It's only nil if in a given buffer you have never set a mark!

Right. So what is so bad about using the `region-active-p` check? The only downside is that the marking does not work if transient-mark-mode is off, but this seems acceptable to me. If you disable transient mark mode it is less obvious that we cycled to a region target.

> Maybe we could add a general way for actions to tell Embark what the next target should be? It would be good for marking, but probably also for other things.

Okay, it is a reasonable generalization to introduce an `embark-repeat-target-type` alist. So far, most actions keep their target as is (the default) and the mark commands are the only exception. Therefore I would not yet go with this generalization for only these two cases. But as soon as another target type switching command pops up we should generalize.

minad:
For now I hard-coded the `outline-mark-subtree` command and added a TODO comment. The org problem is caused by org mode remapping of the outline commands. I added the relevant org commands to `embark-repeat-actions` to fix the issue.

oantolin:
I think I'll go with `use-region-p` for now: I just realized the target finder for region only looks for active regions anyway.

But for the future, if we come up with non-`region` applications, the `embark-repeat-actions` command could be generalized: in addition to the action entries we have now, we could also allow `(action . next-target-type)` entries. It makes sense to store the information in that variable, since this question of where to cycle to is only relevant for repeatable actions.

oantolin:
Thanks for everything, in particular for figuring out the org business: I hadn't had time to look into it yet.

minad:
Thanks for merging! I opened the issue https://github.com/oantolin/embark/issues/418 to keep this repeat type changing business in mind.

Maybe I will find in a few days to implement some embark-org target finders. Then we could in particular specialize the heading type to an org-heading type.

oantolin:
I found the reason I was explicitly checking for `mark` in a [reddit comment](https://www.reddit.com/r/emacs/comments/p3646u/comment/h9invgd/?utm_source=share&utm_medium=web2x&context=3). See its parent too.

PR #417, opened 2021-12-09
Profile

minad:


oantolin:
😮 There's a second profiler in Emacs I completely forgot about!

minad:
Yes, I've learned about it by @dgutov in his Emacs conf talk! I didn't know about elp before.

oantolin:
> Yes, I've learned about it by @dgutov in his Emacs conf talk! I didn't know about elp before.

I could remember where I learned about it, but it was also from that talk. So I've only known about it a short while.

oantolin:
Why did you pick `%` for the binding?

minad:
Hmm, does not make so much sense for an instrumenting profiler. Feel free to change it to something else - a letter.

minad:
Omar, I changed the keybinding to "s" and "S" in https://github.com/oantolin/embark/pull/417/commits/55cf25713ab9974f0a629d6f1ae459a7309b8a2d.

PR #420, opened 2021-12-10
∀

minad:
:smile:


oantolin:
I love it! But... I do worry about font support. Will this appear as tofu for some people?

minad:
I would take the chances :) Many terminals support unicode fonts now. One could add a check if the character is displayable but users can also overwrite the label function if they use a terminal/font without unicode support.

oantolin:
Let's merge and wait for complaints. I did also like the Kleene star you had previously, a math pun too.

PR #425, opened 2021-12-11
..../---/.-- -.-./---/..-/.-../-.. .--/. ..-./---/.-./--././- -/..../…

minad:
…./.../. ...-/./.-./-.-- ../--/.--./---/.-./-/.-/-./- .-/-.-./-/../---/-./.../..--..

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -3562,6 +3562,16 @@ and leaves the point to the left of it."
   ("SPC" mark)
   ("DEL" delete-region))
 
+;; TODO add more encode actions, see M-x or C-h f encode region
+(embark-define-keymap embark-encode-map
+  "Keymap for Embark region encoding actions."
+  :parent nil
+  ("r" rot13-region)
+  ("." morse-region)
+  ("-" unmorse-region))
+
+(fset 'embark-encode-map embark-encode-map)
+
 (embark-define-keymap embark-sort-map
   "Keymap for Embark actions that sort the region"
   :parent nil
@@ -3574,7 +3584,6 @@ and leaves the point to the left of it."
   ("n" sort-numeric-fields))
 
 (fset 'embark-sort-map embark-sort-map)
-
 
 ;; these will have autoloads in Emacs 28
 (autoload 'calc-grab-sum-down "calc" nil t)
@@ -3599,9 +3608,6 @@ and leaves the point to the left of it."
   ("f" fill-region)
   ("p" fill-region-as-paragraph)
   ("$" ispell-region)
-  ("r" rot13-region)
-  (".." morse-region)
-  (".-" unmorse-region)
   ("=" count-words-region)
   ("s" whitespace-cleanup-region)
   ("t" transpose-regions)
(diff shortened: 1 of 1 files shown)


oantolin:
I mean, I did add a binding for `rot13-region` (which I do actually use sometimes).

minad:
More seriously - we could introduce an embark-encode-map (similar to the sort map) which could have such encodings and maybe even some practical ones (url encode for example).

oantolin:
That's a good idea. We have no reason to avoid prefixes anymore, now that the indicators can handle them correctly.

oantolin:
Are there some other encoding commands it would be useful to have here? I guess at least base64 and url-encode-url (which requires a small wrapper, I guess). There is something called base64url as well.

minad:
There are also `format-encode-region` and `format-decode-region` which ask the user for the desired format. See also `format-alist`.

oantolin:
If we really wanted to avoid shadowing `embark-collect-snapshot` and `embark-export` we could bind `whitespace-cleanup-region` to `SPC` (shadowing `mark` which is useless for region targets!), and that leaves `s` open for the sort map. We could bind the encode map to `>`. What do you think?

minad:
> What do you think?

Sounds good! The base64 actions are a bit questionable imo since they are more useful when operating on binary.

oantolin:
Of course! Let's not add base64, then.

minad:
Ah, I forgot. "e" is eval, which is important!

minad:
Okay, from my side this is ready.

oantolin:
I think I might add a couple of more already before merging. I also want to change the binding of `reverse-region` to lowercase `r` which is now free since we moved `rot13-region` under the prefix.

minad:
Sure, please do! Also the `ispell-word` binding is questionable since there is `M-$`.

oantolin:
> Sure, please do! Also the `ispell-word` binding is questionable since there is `M-$`.

Let the which-key users complain about clutter. 🙃

oantolin:
Should the url-encode action replace the buffer text or save the encoded version on the kill-ring?

minad:
> Should the url-encode action replace the buffer text or save the encoded version on the kill-ring?

I think the commands which encode/decode should directly manipulate the buffer. For non invertible commands it makes more sense to save to the kill ring, like the hash commands do.

oantolin:
I agree, I guess the use case is simply writing a URL: you write it with spaces and whatnot, mark it, encode it.

embark.el
```diff
@@ -3598,7 +3598,7 @@ and leaves the point to the left of it."
   ("m" embark-hash-md5)
   ("1" embark-hash-sha1)
   ("2" embark-hash-sha256)
-  ("3" embark-hash-sha356)
+  ("3" embark-hash-sha384)
```
minad, line 3601: Oops :laughing:

minad:
Yay, thanks for merging. Is there also a decode-url command? I haven't found such a function, but I didn't dig deeply. Eww must have something like this?

oantolin:
There is a function but it's called url-unhexify-string. I added the corresponding action. bc460d9654c1728b7f8ffd5df8939c1f0054c32f

oantolin:
Thanks for doing this. I think it's worth it even if just for publicizing that Emacs has all those hash functions. 😃

minad:
You are welcome! Yes, I like Embark also for its exploratory aspect.

oantolin:
I USED BOTH SHA256 AND Base64 TODAY!

minad:
Unbelievable! I hope you could use it multiple times repeatedly such that you *really* profited from the nice Embark integration. I didn't even remember the keybinding, but > is a nice one!

PR #428, opened 2021-12-12
Add minibuffer indicators for Vertico and Selectrum (Fix #427)

### LINKED ISSUE #427: Vertico/Selectrum etc indicators
Do you mind if we add the UI indicators to Embark for a better ootb experience?

~~~ elisp
(defun +embark-vertico-indicator ()
  "Embark indicator which highlights the current Vertico candidate."
  (let ((fr face-remapping-alist))
    (lambda (&optional keymap _targets _prefix)
      (when (bound-and-true-p vertico--input)
        (setq-local face-remapping-alist
                    (if keymap
                        (cons '(vertico-current . embark-target) fr)
                      fr))))))
~~~

There are at least two configs which have it, Doom and my config ;) But it is only a gimmick and not strictly needed for the integration so it would be understandable if you say no the addition here. I think we discussed about this before, but back then I considered this highlighting experimental. Now I think I would miss it.

minad:


oantolin:
Great, thanks! Will you let the Doom people know?

PR #429, opened 2021-12-12
Add a few more region actions

minad:
I also added base64 now, at least decoding can be sometimes useful. And one could still try to continue to operate on the binary data, even if it is not a strength of Emacs. Anyways these are just suggestions :)


embark.el
```diff
@@ -3919,7 +3919,7 @@ and leaves the point to the left of it."
   ("$" ispell-region)
   ("f" fill-region)
   ("u" upcase-region)
-  ("d" downcase-region)
+  ("l" downcase-region)
```
oantolin, line 3922: Oh my god, you're right: `d` for `downcase` was ironically *not* mnemonic at all!

embark.el
```diff
@@ -3623,13 +3630,14 @@ and leaves the point to the left of it."
   ("SPC" mark)
   ("DEL" delete-region))
 
-;; TODO add more encode actions, see M-x or C-h f encode region
+(autoload 'message-caesar-region "message" nil t)
 (embark-define-keymap embark-encode-map
   "Keymap for Embark region encoding actions."
   :parent nil
-  ("r" rot13-region)
+  ("r" message-caesar-region) ;; default rot13-region
```
oantolin, line 3637: This one I'm not so sure about: I think the only case of the Caeser cipher people actually use is Δ=13. I think I'd rather not load `message.el` for what is most likely to be `rot13` anyway.

embark.el
```diff
@@ -3638,8 +3646,12 @@ and leaves the point to the left of it."
   ("5" embark-hash-sha512)
   ("f" format-encode-region)
   ("F" format-decode-region)
+  ("b" base64-encode-region)
+  ("B" base64-decode-region)
   ("u" embark-encode-url)
-  ("U" embark-decode-url))
+  ("U" embark-decode-url)
+  ("c" epa-encrypt-region)
+  ("C" embark-epa-decrypt-region))
```
oantolin, line 3654: I'm not familiar with these functions: why is `epa-encrypt-region` OK as-is but `epa-decrypt-region` requires a little wrapper?

embark.el
minad: Oh, I thought you like the generalization :laughing: 

If I apply a Caesar chipher  to a region I guess I wouldn't bother if message.el is loaded. Is it that heavy?

embark.el
minad: It is an inconsistency in Emacs. encrypt always replaces but decrypt asks by default :shrug:

embark.el
minad: Actually I think this should be fixed in Emacs itself. message-caesar-region should be renamed to rot13-region and removed from message.el. There is no point in having both these commands. But I don't dare to report a bug for this :smile:

embark.el
minad: I am not sure I get your joke here, but you used "l" at another place for downcase. And we use "d" relatively consistently for delete.

oantolin: It wasn't a joke: in general Embark tries to follow Emacs's letter choices for keybindings, and Emacs uses `l` for downcasing: `M-l` and `C-x C-l` have an `l` despite being bindings for commands that start with `downcase`. So, if I followed my own guideline I would have bound this to `l` long ago.

minad:
Do you want me to change anything here, the rot13? Or do you apply the changes yourself before merging the PR?

oantolin:
I'll make the change. You don't mind losing the Caeser cipher, do you?

minad:
No, I don't mind it.

oantolin:
Done. Thanks again! This seem silly, but I am grateful especially for changing the binding of `downcase-region` to `l`.

minad:
Thanks, but poor caesar. I should submit the bug report :-P

PR #430, opened 2021-12-12
Minor fix for embark-target-file-at-point

localauthor:
I was getting "wrong-type-argument" errors. This seems to fix that.


oantolin:
Oh, that's a bad typo to have! Thanks for fixing it!

PR #432, opened 2021-12-13
Add candidate collector for consult-completing-read-multiple

minad:
See also https://github.com/minad/consult/commit/1cfc06800bb5663f0e516cd74179256bd738b38d and https://github.com/minad/consult/commit/8cadf148239be7f4a5b0350068d17fe07739fc3b

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -320,5 +320,14 @@ that is a Consult async command."
 (setf (alist-get 'imenu embark-default-action-overrides) #'consult-imenu)
 (add-to-list 'embark-candidate-collectors #'embark-consult-outline-candidates 'append)
 
+;; consult-completing-read-multiple
+
+(defun embark-consult--crm-selected ()
+  "Return selected candidates from `consult-completing-read-multiple'."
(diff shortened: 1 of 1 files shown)


oantolin:
It would be fun to have a way to temporarily turn any `completing-read` call into a `consult-completing-read-multiple` call, just so you can do ad hoc candidate marking and use `embark-act-all`. This could be an alternative to `embark-select-one` or a Vertico-specific candidate selection UI.

minad:
Hmm, then I would rather implement multi marking in Vertico or the respective UI directly. The problem is that `completing-read` cannot always be transformed into `completing-read-multiple`, since `completing-read-multiple` only works with completion tables without completion boundaries. For example you cannot do `completing-read-multiple` with `read-file-name-internal`. 

To be honest, in my opinion `completing-read-multiple` is a broken and underspecified API with numerous edge cases and problems. I would not throw it into the mix if it can be avoided. For this reason I am also sufficiently happy with `consult-completing-read-multiple` and did not bother to implement rich support in Vertico directly.

PR #434, opened 2021-12-13
embark-face-map: Add describe-face

minad:


oantolin:
How on earth was that missing!? Thanks!

PR #440, opened 2021-12-29
Support generic multi-category

minad:
The consult-multi category can be useful for other packages besides
Consult. For example the Citar package has a completion command which
offers both files and urls. We keep consult-multi support for a short
transition period.

See also https://github.com/minad/marginalia/pull/119 and https://github.com/minad/consult/pull/488

Omar, what do you think about the proposed generalization? Also we may want to start a name bikeshedding discussion. ;)

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -201,19 +201,6 @@ The elements of LINES are assumed to be values of category `consult-line'."
 (setf (alist-get 'consult-grep embark-collect-initial-view-alist)
       'list)
 
-;;; Support for consult-multi
-
-(defun embark-consult--multi-transform (_type target)
-  "Refine `consult-multi' TARGET to its real type.
-This function takes a target of type `consult-multi' (from
-Consult's `consult-multi' category) and transforms it to its
-actual type."
-  (or (get-text-property 0 'consult-multi target)
-      (cons 'general target)))
-
-(setf (alist-get 'consult-multi embark-transformer-alist)
-      #'embark-consult--multi-transform)
-
 ;;; Support for consult-isearch
 
 (setf (alist-get 'consult-isearch embark-transformer-alist)
--- a/embark.el
+++ b/embark.el
@@ -178,7 +178,10 @@ bounds pair of the target at point for highlighting."
     (symbol . embark--refine-symbol-type)
     (embark-keybinding . embark--keybinding-command)
     (project-file . embark--project-file-full-path)
-    (package . embark--remove-package-version))
+    (package . embark--remove-package-version)
+    (multi-category . embark--refine-multi-category)
+    ;; TODO: `consult-multi' has been obsoleted by `multi-category'. Remove!
+    (consult-multi . embark--refine-multi-category))
   "Alist associating type to functions for transforming targets.
 Each function should take a type and a target string and return a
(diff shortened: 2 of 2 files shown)


oantolin:
The proposed generalization is just a renaming in recognition of the fact that other clients such as citar may want this functionality so its not really consult specific, right? That sounds good to me. If you want to bikeshed names, let me say the thing I don't quite like about `multi-category` is that each individual target only has a single category, not multiple categories. I think a name like `proxy` might make more sense. But I'm also OK with `multi-category` because it does make sense at the `completing-read` call site.

minad:
Yes, it is just a renaming and also the additional guarantee that the API is sufficiently good for other packages and not Consult-private. I meant the name in the sense that the candidate has a "multi" category, not multiple categories and I didn't want to use only `multi` to avoid name clashes. I also like to use the same symbol for the metadata category and the candidate text property.

minad:
But now is the time to come up with a better name :grinning:

oantolin:
I think `multi-category` is fine, but if you want proposals, how about `proxy-category`, `indirect-category` or `dispatch-category`?

minad:
Hmm, I like multi-category better than the other proposals. Not that I disagree with your criticism, but the multi at least coveys that we have candidates of multiple types.

minad:
mixed-category?

oantolin:
Let's stick with `multi-category` then. Since you included backward compatibility in the PRs, we should just merge now.

oantolin:
Ah, I hadn't thought of completing-read-multiple.

oantolin:
It's not too bad of a clash. At least crm has "multiple" spelled out in full and this is just "multi".

oantolin:
I think you probably do have some `consult--multi` callers out there in the wild and even though the name has a double-dash, I don't think we need to disturb them just for this.

minad:
> I think you probably do have some consult--multi callers out there in the wild and even though the name has a double-dash, I don't think we need to disturb them just for this.

Okay, let's proceed with multi-category then. But I actually like mixed-category more now. This would be a good opportunity to remind the `consult--multi` users that `consult--multi` is double-dashed ;)

oantolin:
Well `consult--multi` has multiple sources, but I'm not sure I'd say it mixes them: the sources are like water and oil which famously don't mix and sit atop each other in a fixed order.

minad:
You can use `flex` to create an emulsion. But the UI may demix the candidates again.

oantolin:
But crm is a good argument against `multi`. You decide. If you want mixed-category let's just do it now and you can include a deprecation warning for `consult--multi` in Consult.

minad:
Another word is "merge". For example there is `completion-table-merge`. I am really not that happy with `multi` anymore since it is too generic and overused. merge or mixed would be better and then I will see if I rename consult--multi. But then we have the unnecessary churn. I am annoyed that I didn't think earlier of this when I chose the name `consult--multi` :(

minad:
Let's keep multi-category. It would also sound bad to create a command of mixed sources. It is not really mixing.

I merged https://github.com/minad/marginalia/pull/119.

hmelman:
I don't have an opinion but would `heterogeneous` or `various` work?

PR #441, opened 2021-12-30
Add occur action to identifier map

minad:


oantolin:
Sure, why not? I know it's a useful action on identifiers because I actually use it a lot with the default key binding of `M-s o`. It's probably worth skipping the `M-s`.

minad:
`M-s o RET` vs `C-. o`

minad:
Of course I also used `M-s o RET`, but I just saw that the Emacs 28 context menu adds occur on identifiers, see https://ruzkuku.com/texts/emacs-mouse.html.

oantolin:
I think the default for occur is the last regexp you searched for, and the symbol at point is the *next* element of future history, so I think to search for the symbol at point you need `M-s o M-n M-n RET` with two `M-n`s! Or, with Embark, `C-. M-s o` (which is what I said I used to do), or, now thanks to this PR `C-. o`.

oantolin:
You can also do `M-s . M-s o C-g` (the `C-g` is to exit isearch).

oantolin:
So even without the `o` binding, I think `C-. M-s o` was still the best way to run occur for the symbol at point.

minad:
Hmm, when I press `M-s o` it says `(default: <symbol at point>)`.

oantolin:
Is that prior to running any occurs in that buffer?

minad:
Oh sorry, that's a custom occur override.

oantolin:
Oh, OK. Sounds like a useful bit of configuration if you don't have Embark. 😛

oantolin:
Another one I use a lot as an action on identifiers is `query-replace`. I use it via the default binding of `M-%`, maybe it's also worth giving it a simpler binding.

minad:
> Another one I use a lot as an action on identifiers is query-replace. I use it via the default binding of M-%, maybe it's also worth giving it a simpler binding.

Sounds good. I also use the default binding for that.

oantolin:
This made me remember I sometimes want to run `occur` on something that's not quite an indentifier, like maybe an indentifier in quotes or just a prefix like `consult-imenu-`. So it would be nice to mark what I want to search for and use `occur` as an action. The problem is that if the region is active commands like occur, isearch or query-replace restrict their search to the active region, so I think I'll add an `embark--unmark-target` pre-action hook for those commands.

minad:
> So it would be nice to mark what I want to search for and use occur as an action. 

I agree. Also thought about this but I was too lazy to implement it properly :-P Both occur and replace could profit from such a treatment.

oantolin:
Done: b1aee4b72ed855837de63d16eb604ba2ec73b132

minad:
Perfect! The question is if you want to bind this generally, similar to the consult map we could introduce a search map? But then we lose the short keybinding.

oantolin:
I don't think we should add any key bindings to `embark-region-map` for things like `occur` or `query-replace`, since it's probably not used often enough. Plus those things are very basic commands people probably have strong muscle memory for. That's also why I'm not certain I really want a binding for `query-replace` in `embark-identifier-map`.

minad:
I agree. These are basic. But for identifier occur `C-. o` already replaced `M-s o RET` for me ;)

oantolin:
I'll have to see if I can retrain my brain from `C-; M-s o` to `C-; o`. It'll take a few days, I suspect.

Changing topics, I'm really happy with how powerful the pre-action/target-injection/post-action hooks are. They can really smooth out all sorts of little issues that keep pre-existing commands from being good actions.

oantolin:
I'm still on the fence about a simpler binding for query-replace in the identifier-map. Maybe instead we should bind a tiny wrapper that does the replacement for `\<_target\_>`? It's a little annoying that the replacement isn't done in symbol mode.

Oh, and query-replace could also use `embark--beginning-of-target` as a pre-action hook, so you don't miss the occurrence you acted on  if you were somewhere in the middle!

oantolin:
OK, I added the `embark--beginning-of-target` pre-action hook for the replace operations. 74c1e74503c2007c157e0a05c8127d2d5b47f8f0

oantolin:
Forget the wrapper that replaces only at symbol boundaries: I forgot that the prefix argument to `query-replace` makes it search only at word boundaries. That's close enough.

minad:
> Changing topics, I'm really happy with how powerful the pre-action/target-injection/post-action hooks are. They can really smooth out all sorts of little issues that keep pre-existing commands from being good actions.

Yes, Embark converged to a good design!

PR #446, opened 2022-01-11
Minor stuff

minad:


minad:
> Guess I hadn't run package-lint in a while. Thanks.

Actually I had it broken once again for a longer time in my config. I even submitted a broken package to MELPA (https://github.com/melpa/melpa/pull/7850#issuecomment-1008170097), how embarrassing! :-P

PR #449, opened 2022-01-23
Don't quit by default when an unexpected key is pressed

### LINKED ISSUE #448: Allow keeping (don't quit) the minibuffer when the pressed key isn't presented in the indicator
Steps to reproduce:
1. Type a complex and long input in the minibuffer prompt
2. Invoke `embark-act`
3. Press a key that isn't recognized, such as `<down>` as `vertico-next`

Expected: 
While not quitting the minibuffer, either:
- The indicator disappears, `vertico-next` is either executed or not
- The indicator persists, `vertico-next` is not executed

Observed: 
The minibuffer command is aborted, typed input is lost.

daanturo:
Fixes #448 by introducing `embark-quit-other-actions` (default to `t`).

Set it to `nil` to prevent quitting.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -1998,16 +1998,14 @@ target."
     (unwind-protect
         (while
             (let* ((target (car targets))
+                   (keymap (let ((embark-default-action-overrides
+                                  (if default-done
+                                      `((t . ,default-done))
+                                    embark-default-action-overrides)))
+                             (embark--action-keymap (plist-get target :type)
+                                                    (cdr targets))))
                    (action
-                    (or (embark--prompt
-                         indicators
-                         (let ((embark-default-action-overrides
-                                (if default-done
-                                    `((t . ,default-done))
-                                  embark-default-action-overrides)))
-                           (embark--action-keymap (plist-get target :type)
-                                                  (cdr targets)))
-                         targets)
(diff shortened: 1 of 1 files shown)


minad:
I don't think we need an option for that? We try to avoid adding more and more options to Embark if not strictly necessary, since there are already so many.

daanturo:
So can we make not quitting the default? Arguably abruptly exiting isn't a good user experience.

minad:
Yes. going back to the minibuffer should be the correct behavior as I see it. You've either discovered a bug in Embark or in your config which should be fixed :)

oantolin:
I'm not merging this because it's based on a misunderstanding of Embark: in Embark the keymaps are just for convenience and you can run any command as an action, I do not wish to limit action to those listed in the keymaps.

minad:
What about undefined keys? I thought in this case it makes sense to recover to the prompt. But afaik Embark already does this and shows "Not an action" or "Canceled".

oantolin:
For undefined key you already get "Canceled", and you get "Not an action" for any key bound to `self-insert-command`. Maybe that should be generalized so it doesn't apply only to `self-insert-command` but to a customizable list of commands.

PR #456, opened 2022-02-08
bookmark-delete, package-delete: Add confirmation

minad:


oantolin:
I guess that does feel more consistent.

PR #462, opened 2022-02-15
Tab actions

minad:
Maybe the README should be rewritten slightly to make it clear that the tab example has been added to Embark?

### DIFF
--- a/README.org
+++ b/README.org
@@ -597,11 +597,14 @@ similar situations where the easiest option is not available.
 
 *** New minibuffer target example - tab-bar tabs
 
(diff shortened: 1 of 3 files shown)


oantolin:
Thank you!

oantolin:
Tabs are pretty nice, I'm starting to use them myself instead of putting window configurations in registers. For me the key difference is that window configuration also store the point location in each buffer which is something I usually don't want restored.

minad:
Thanks for merging!

Yes, I also started to use them as a replacement for winner-mode, inspired by @protesilaos. Instead of having one window configuration history, I have multiple now for each tab. I had always been turned off by tab-bar-mode because I assumed that it implies that you need an actual visual tab bar. Turns out tab-bar is nice without the visuals. I've also started to play around with persisting the tab configurations, see https://github.com/minad/tab-bookmark/.

PR #463, opened 2022-02-17
consult-register uses multi-category now

minad:
See https://github.com/minad/consult/commit/a10d59436f087f1bc79b009cb4dfb155a7ad5dea

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -206,11 +206,6 @@ The elements of LINES are assumed to be values of category `consult-line'."
 (setf (alist-get 'consult-isearch embark-transformer-alist)
       #'embark-consult--target-strip)
 
-;;; Support for consult-register
-
-(setf (alist-get 'consult-register embark-collect-initial-view-alist)
-      'zebra)
-
 ;;; Bindings for consult commands in embark keymaps
 
 (define-key embark-file-map "x" #'consult-file-externally)


oantolin:
This seems like a loss of functionality for not much gain. Now there's no easy way to tell embark-collect to use zebra mode for registers and it seems to me that the only thing gained (from Embark's point of view) by the move to multi-category is that now file registers refine to files and thus get file actions. Is that right, are file registers the only kind of register that has an associated refined target type?

oantolin:
Probably the loss of zebra mode is not a big deal, I expect almost nobody ever runs an embark-collect from consult-register. And gaining file actions is nice, of course.

I like the idea of that consult-buffer file register source. I can imagine creating a bunch of file registers but later forgetting I defined some of the older ones. It's nice that consult-buffer can remind that they also exist as a register.

minad:
Yes, I also doubt that one exports the registers to Embark, therefore I didn't mind the zebra mode problem. Maybe I will add this source to Consult then if you also like it. I have to experiment with it a bit more. I think it is reasonable to restrict it to file/buffer registers.

oantolin:
I should make the explicit caveat that while I like the _idea_ of that source, I'm unlikely to benefit from it myself since I don't (currently?) use file registers!

minad:
I added it to the wiki for now: https://github.com/minad/consult/wiki#file-register-source

Issue #166, opened 2021-02-23
Allow to mark, and act on, multiple candidates in collect buffer


bdarcus:
The conversation [here](https://github.com/tmalsburg/helm-bibtex/pull/355#issuecomment-784177382) has me wondering:

Would it be possible, and wise, to allow one to mark, and act on, multiple candidates in a collect buffer, as in `helm`?

Use case:

I want to open multiple files, so I:

1. narrow the candidates
2. run `embark-collect-snapshot`
3. mark the three I want, using some keybinding
4. hit return for the default action (open the files)

oantolin:
To be clear: we are only talking about applying the same single-argument action to multiple candidates, right? We aren't talking about a new kind of actions that receives multiple candidates at once?

I have considered a feature to mark candidates and apply the same action to each one, and am of two minds about it.

If the only way to mark candidates is one at a time, like, you move to a candidate and press a key to mark it, then move to another, etc., then it's probably not worth it, since it hardly saves any keystrokes: you could just activate `embark-collect-direct-action-minor-mode-map`, which binds the actions directly without needing to call `embark-act` first, and go down the list of candidates running the actions.

If I added more ways to mark candidates, say, mark all, marking all that match a regexp, toggle marked/unmarked, etc. I'd be reimplementing politza's great [tablist](https://github.com/politza/tablist) package. So I'd be inclined to instead make a small embark-tablist package, that depends on both embark and tablist, and provides an `embark-tablist-act-on-marked` command.

By the way, even without the embark-tablist package existing yet, I think it's already worth using tablist with embark collect buffers.

oantolin:
Note that for example Ivy doesn't (or rather, didn't back when I used it) have a way to mark a bunch of items first and act on them later. Instead, it just has a way to set an action that will take a single keystroke to execute, and then you march down the list and run it on the items you want.

minad:
You mentioned this tablist package multiple times already. How is it useful for embark snapshots? Do you enable some tablist-mode by default in your collect buffers? I need a simple example use case to understand what you use it for right now, since this embark-act-on-marked command does not yet exist :)

But it seems that such an embark-act-on-marked would satisfy the use case by @bdarcus perfectly.

oantolin:
[...]

Issue #466, opened 2022-02-23
Proposal: remove embark-collect-live and embark-collect-completions


oantolin:
First of all, let me emphasize that `embark-collect-snapshot` is here to stay! (Although I guess that if the live and completions variants are removed, then we might as well rename `embark-collect-snapshot` to simply `embark-collect`.)

I think that probably very few people use `embark-collect-completions` or `embark-collect-live`, and removing them would let me significantly simplify the code for `embark-collect-snapshot`. I myself never used `embark-collect-live` and I bet I was probably one of the last holdouts using `embark-collect-completions`.

It seems like a reasonable philosophical stance to say that `embark-collect-snapshot` and `embark-export` are of a very different nature than `embark-collect-live` and `embark-collect-completions` and that the latter two are really out of scope for Embark.

In practical terms I was still using `embark-collect-completions` for its features. Of course its main deficiency is how slow (#101) it is, and for a long time I longed to switch to Vertico (which is what I use now), but did not switch because other than being slow, `embark-collect-completions` had a lot to offer:

- It has both a list view (with annotations) and a compact grid view (withoot annotations).
- It's easy to configure it to not be on all the time, to only popup on demand.
- You can configure the initial view per completion category.
- It can display multiline completion candidates, and has zebra stripes to make it clear where one ends and the next one begins.
- It nicely aligns annotations for you.

But with (relatively) recent improvements to Vertico and Marginalia you can get almost of that in Vertico now, and Vertico has always been much, much faster. Only multiline display of multiline candidates (and zebra stripes) is missing from Vertico, but I've found I don't really need that at all! Also, let me say that if you want completions displayed only on demand like I do, `vertico-unobtrusive-mode` is much better than truly displaying *zero* completions. 

I've never really recommended `embark-collect-completions` and would now actually recommend *against using it*, I mean, why wouldn't you use Vertico instead?

I'm creating this issue to give people a chance to protest or in general give their opinion before I do anything about this.

minad:
I am in full support of removing this feature, since it narrows the focus of the package as a package centered around actions. Also having shorter command names is better and there has been confusion before about the difference of `embark-collect-snapshot` and `embark-collect-live`.

[...]

PR #467, opened 2022-03-03
Marking candidates

minad:
See the discussion in #166

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -718,6 +718,10 @@ There are three kinds:
 - :live, which does
 - :completions, which also auto-updates, but is ephemeral.")
 
+;; TODO Do we need this variable?
+;; Also add a list of marked candidates, or a list of overlays?
+;; When switching the display modes we have to move the overlays too.
+;; We could also turn this variable into an alist, with value=marked.
 (defvar-local embark-collect-candidates nil
   "List of candidates in current collect buffer.")
 
--- a/embark.el
+++ b/embark.el
@@ -2576,7 +2576,17 @@ all buffers."
   "Return candidates in Embark Collect buffer.
 This makes `embark-export' work in Embark Collect buffers."
   (when (derived-mode-p 'embark-collect-mode)
-    (cons embark--type embark-collect-candidates)))
+    (cons embark--type
+          (or (save-excursion
+                (mapcar
+                 (lambda (ov)
+                   (goto-char (overlay-start ov))
+                   (cadr (embark-target-collect-candidate)))
+                 (nreverse
+                  (seq-filter
+                   (lambda (ov) (overlay-get ov 'embark--collect-mark))
+                   (overlays-in (point-min) (point-max))))))
+           embark-collect-candidates))))
 
 (defun embark-completions-buffer-candidates ()
   "Return all candidates in a completions buffer."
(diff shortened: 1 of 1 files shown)


minad:
@oantolin Can we merge #466 before proceeding on this one?

minad:
> Sure, but it'll take me a couple of days to get to this probably.

There is no hurry here! I just think it is better to do a bit of cleanup first before we add new features, see also #468.

oantolin:
I've implemented the simplified embark-collect & embark-live now.

oantolin:
I'm going to take this code and update it for the recent changes.

oantolin:
Done. 6a8e6ded2da3e4bc9672aefe956c0f13e625cab2

I added toggle-marks and unmark-all commands as well.

minad:
Great, thanks! How did you address this comments about moving the overlays when toggling the view modes?

> ;; TODO Do we need this variable?
;; Also add a list of marked candidates, or a list of overlays?
;; When switching the display modes we have to move the overlays too.
;; We could also turn this variable into an alist, with value=marked.

oantolin:
I did not address it all. 🙄 The marks just vanish. 🤣

oantolin:
I created issue #473 so I don't forget to fix this.

bdarcus:
Where would you like feedback, or possibly dumb questions, @oantolin? Here?

oantolin:
Here would be fine, @bdarcus.

bdarcus:
OK, so I tested this on my obvious use case: inserting multiple citations.

So I added `org-cite-insert` to `embark-multitarget-actions`, and then:

1. narrowed completions
2. ran embark-collect
3. marked two candidates
4. ran embart-act-all

... and here's where I got stuck.

I think ideally embark would just run the command.

But it doesn't; it prompts if I want to run it. If I do, it still doesn't actually do so, and prompts again.

In the end, I can't get a result, and find it all confusing at this point.

Am I doing something wrong?

And did you add an option to not prompt? I recall we discussed this earlier.

bdarcus:
Also, how would you two imagine keybindings for mark/unmark?

I had been wondering about `SPC` for both, but I guess, notwithstanding the question of whether that's a good idea, that's currently not possible since there's no `embark-collect-toggle-mark` (there is a "toggle-marks" command, but that applies to all candidates in the collect buffer)?

oantolin:
Oh, I try to not invent key bindings if I can avoid it: I just copied copied dired and ibuffer, mark is `m`, unmark is `u`, toggle everything is `t`, unmark everything is `U`. That's what I imagine Emacs user already have muscle memory for.

bdarcus:
> Oh, I try to not invent key bindings if I can avoid it: I just copied copied dired and ibuffer, mark is `m`, unmark is `u`, toggle everything is `t`, unmark everything is `U`. That's what I imagine Emacs user already have muscle memory for.

Ah right; makes total sense!

bdarcus:
> > So I added `org-cite-insert` to `embark-multitarget-actions`
> 
> I think this shouldn't work, because that tells Embark to call the functions with a list of candidates, and `org-cite-insert` does not take a list as an argument.

OK. I need to think about that some more.

But if I do `citar-insert-citation` instead, I do get a result closer to expectation, but it's not the key that gets inserted, but the whole candidates. 

I don't recall seeing anything like that with `embark-act-all` earlier.

oantolin:
This seems to work if you have `org-cite-insert-processor` set to `basic`:

```elisp
(defun cite-all (refs)
  (interactive)
  (cl-letf (((symbol-function #'org-cite-basic--complete-key)
             (lambda (&optional _)
               (let ((tbl (org-cite-basic--key-completion-table)))
                 (mapcar (lambda (ref) (gethash ref tbl)) refs)))))
    (org-cite-insert nil)))

(add-to-list 'embark-multitarget-actions 'cite-all)
```

PR #468, opened 2022-03-03
Mark embark-collect variables as private

minad:
Without deprecation since I think these variables should have been considered private. I am not aware of any users outside of Embark. Interestingly the embark-collect-* variables are the last variables of Embark which are public besides the customization variables and the keymaps. What do you think?


oantolin:
I wound up redoing this while working on the embark-collect/embark-live simplification.

PR #475, opened 2022-03-22
elp-restore-package, avoid some package lint warnings

minad:


embark.el
```diff
@@ -2868,7 +2867,7 @@ candidate."
       (delete-overlay ov))))
 
 (defun embark-collect-toggle-marks ()
-  "Toggle marks: marked candidates become unmarked, and vice versa."
+  "Toggle markings: Marked candidates become unmarked, and vice versa."
```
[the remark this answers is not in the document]
minad: I just wanted these warnings to go away :shrug:

embark.el
```diff
@@ -3522,15 +3521,15 @@ With a prefix argument EDEBUG, instrument the code for debugging."
...
-;; TODO Report Emacs bug, this function should be provided by Emacs itself.
-(defun embark-elp-restore-package (prefix)
-  "Remove instrumentation from functions with names starting with PREFIX."
-  (interactive "SPrefix: ")
-  (when (fboundp 'elp-restore-list)
-    (elp-restore-list
-     (mapcar #'intern
-             (all-completions (symbol-name prefix)
-                              obarray 'elp-profilable-p)))))
+(when (< emacs-version 29)
```
oantolin, line 3524: I think it would be better to use `(unless (fboundp 'elp-restore-package) ...`) as it more directly expresses intent (and also because I just know I'll wind up using some weird build of Emacs 29 that doesn't have the function 😛).

minad: > I think it would be better to use (unless (fboundp 'elp-restore-package) ...) as it more directly expresses intent (and also because I just know I'll wind up using some weird build of Emacs 29 that doesn't have the function stuck_out_tongue).

Sure. I considered this of course. Problem is only that this doesn't work since `elp-restore-package` is not autoloaded.

minad: See https://github.com/emacs-mirror/emacs/commit/d9d38059d193699c2ae6cdf69341f5aa62e87781 for reference.

embark.el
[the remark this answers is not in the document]
minad: It is this stupid verb warning - you should write "mark" instead of "marks".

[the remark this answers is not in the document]
minad: Maybe the new formulation is more acceptable?

embark.el
minad: > and also because I just know I'll wind up using some weird build of Emacs 29 that doesn't have the function

I don't think this will happen. elp-restore-package got added very early in the Emacs 29 cycle. You have already missed the opportunity as long as you don't intentionally pull an outdated version...

PR #488, opened 2022-04-16
Add align-regexp to embark-target-injection-hooks

detvdl:
The `align-regexp` action did not prompt the user for a regular expression unless the `embark--ignore-target` hook was explicitly added.

Since the align-regexp action was added as a default to the region keymap, I assume this addition to the default target-injection-hooks makes sense.

Tested on `GNU Emacs 29.0.50 (build 1, aarch64-apple-darwin21.1.0, NS appkit-2113.00 Version 12.0.1 (Build 21A559))
of 2021-12-07`

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -415,6 +415,7 @@ entry of `embark-target-injection-hooks' whose key is the action."
     (eval-last-sexp embark--ignore-target)
     (embark-eval-replace embark--ignore-target)
     ;; commands which prompt for something that is *not* the target
+    (align-regexp embark--ignore-target)
     (write-region embark--ignore-target)
     (append-to-file embark--ignore-target)
     (shell-command-on-region embark--ignore-target)


oantolin:
Have you signed the FSF copyright assignment? (I know this is under the limit, but that limit is for cumulative contributions, I believe, so it's best to worry about it from the start.)

oantolin:
I made the change myself, just to get this done quicker, before finding out about your copyright assignment status. (I even did not look at your code, though of course that's ridiculous: it's obvious what change you made. 🙄)

detvdl:
> Have you signed the FSF copyright assignment? (I know this is under the limit, but that limit is for cumulative contributions, I believe, so it's best to worry about it from the start.)

I don't recall, so I'm pretty sure I haven't done that yet. 
No worries, though. It was the tiniest, most obvious change, which might as well have been just an issue.
I simply prefer tackling the issue myself on a branch first and test stuff out, to learn, so I figured a PR would be quicker 😄

Thanks for making the changes! 🎉

PR #498, opened 2022-05-02
Outline review

minad:
I looked into the issues I mentioned in https://github.com/oantolin/embark/commit/5d5def1cb5616ba17987d6c5814183a02afc8981 and https://github.com/oantolin/embark/commit/7a2487f5bb8a6ea5a9b5048386d6dcbbc1b6cc85. I made the following changes to fix the issues mentioned there:

- Lazy load outline, outline-minor-mode is autoloaded
- Use a more localized advice to `embark-collect--metadata`
- Bump embark-consult version dependency to 27.1, since Consult also requires 27.1
- Missing xref defvar in embark-consult.el
- Docstring fixes

This seems to work all nicely. Maybe we can find a better solution regarding the advice and maybe we could reuse the group-function from consult-imenu if I extract it there?


minad:
I added another commit which uses remapped keys in https://github.com/oantolin/embark/pull/498/commits/2d017739a04e59bae4f4a717ffc92ecb312ddf02. Not sure if you like this?

oantolin:
I think advising `embark-collect--metadatum` is fine (and certainly better than doing it to `completion-get-metadatum`). Why do you want to avoid advice? I guess we could always add a hook variable...

oantolin:
I'm not too sure about the paragraph key bindings, mainly because they are harder to type than `M-n` and `M-p`. Embark tends to choose the easiest key bindings available. In fact, how about capital `N` and `P`, instead?

minad:
> I think advising embark-collect--metadatum is fine (and certainly better than doing it to completion-get-metadatum). Why do you want to avoid advice? I guess we could always add a hook variable...

We can keep the advice. It is not too bad. I am only against the more global advice to `completion-get-metadata` due to its bad consequences.

> I'm not too sure about the paragraph key bindings, mainly because they are harder to type than M-n and M-p. Embark tends to choose the easiest key bindings available. In fact, how about capital N and P, instead?

Well, I think most people already rebind M-n and M-p (or some other more convenient key) to forward and backward-paragraph. I like when packages use rebinds since then I don't have to make fine adjustments. My main argument is that fundamental-mode comes with a lot of "useless" keybindings, by remapping them we give them purpose in the special buffers. Therefore I prefer this style over explicit binding.

In Vertico I use the same remappings and so far nobody complained, in this sense it would help consistency if people also use Vertico. But I don't really feel strongly about this, if you prefer to bind M-p/M-n explicitly, that's also okay.

oantolin:
This looks great. I'm merging with two minor changes:

- I'm removing the comment about hooking into `embark-collect--metadatum` since I have no qualms about advising my own function (which frankly seems better than adding yet another hook variable).
- I'm changing the key bindings in `embark-collect-mode-map` a bit. The bindings for {forward,backward}-button there were chosen back when there was a grid mode! It makes more sense to use `n` and `p`. I think I'll go with `{` and `}` for the outline commands, following Embark's usual approach of choosing a key binding that reminds you of the one used in another context but without modifiers.

oantolin:
Mmh, we can *also* remap the paragraph commands. I initially thought of keeping them for navigating paragraphs in multiline candidates, but honestly that's not a common case at all.

oantolin:
How does that sound, using both `}` and `[remap forward-paragraph]` for `outline-next-heading`?

minad:
> How does that sound, using both } and [remap forward-paragraph] for outline-next-heading?

Sounds good!

oantolin:
Thanks a lot for doing this! I'm particularly grateful for adding the `(eq prop 'group-function)` clause to the advice function, which was a total oversight on my part. 😛

minad:
> Thanks a lot for doing this! I'm particularly grateful for adding the (eq prop 'group-function) clause to the advice function, which was a total oversight on my part.

Oh, I didn't even see this. I didn't test your code before making my adjustments :laughing:

oantolin:
It was not really _that_ dangerous that I omitted the clause since this is `:after-until` advice, but still I totally meant to include it.

minad:
I will make the changes to consult-imenu later such that you can reuse the functions.

oantolin:
Good, thanks! It's not a big deal because it is a quite small amount of duplication, but it would certainly be a little nicer.

minad:
Okay I added the extra group function, this should simplify things a bit in embark-consult.

oantolin:
Great, I'll rewrite the corresponding bit in embark-consult later today.

By the way, I just realized that loading of outline-minor-mode could be further delayed until the point where you first use embark-collect with a (non-nil) group-function.

minad:
> By the way, I just realized that loading of outline-minor-mode could be further delayed until the point where you first use embark-collect with a (non-nil) group-function.

Oh right, that sounds good.

oantolin:
Done. 5bdf21c

PR #507, opened 2022-05-15
Silence compiler

tarsius:
Though maybe instead of declaring functions and variables you just want to require the relevant org libraries upfront.

### DIFF
--- a/embark-org.el
+++ b/embark-org.el
@@ -99,6 +99,8 @@
     )
   "Supported Org object and element types")
 
+(declare-function org-element-property "org-element" (property element))
+
 (defun embark-org-target-element-context ()
   "Target the smallest Org element or object around point."
   (when-let (((derived-mode-p 'org-mode))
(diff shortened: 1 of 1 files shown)


embark-org.el
```diff
@@ -273,6 +278,7 @@ what part or in what format the link is copied."
 (embark-org-define-link-copier description description "'s description")
 (embark-org-define-link-copier target target "'s target")
 
+(declare-function embark-org-copy-link-inner-target "embark-org")
```
oantolin, line 281: Oh, this one surprises me, I didn't think I'd need to declare this function since it's definition is one line below!

oantolin:
How does this sound: requiring `org-element` immediately, but trying to delay the loading of `ox` as much as possible?

embark-org.el
tarsius: Me too :stuck_out_tongue_closed_eyes:

PR #527, opened 2022-07-26
Prevent dired from reusing embark-export-dired buffer

OkamiW:
Assume you narrow files inside directory ~/.emacs.d, and export the
completion candidates to embark, embark will create a dired
buffer(using dired-noselect) to display them.

The problem is, dired will reuse buffer whenever possible.
If you go to ~/.emacs.d again later, the narrowed buffer will show up,
instead of the full content under ~/.emacs.d...

A quick work around is to unset dired-directory, then
dired-find-buffer-nocreate won't reuse this buffer.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -3223,6 +3223,7 @@ PRED is a predicate function used to filter the items."
                      (mapcar (lambda (file) (string-remove-prefix dir file))
                              files)))))
     (with-current-buffer buf
+      (setq-local dired-directory nil)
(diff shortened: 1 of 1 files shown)


oantolin:
Hi, sorry for not getting to this earlier. There is something I don't understand about this code: according to the source code of `dired-internal-noselect` the buffer only gets reused if the `dir-or-list` argument is a string, since embark calls it with a list argument, it should not be reusing the buffer anyway.

PR #534, opened 2022-08-28
Avoid using 'buffer-local-value' as a generalised variable

phikal:
Related to https://github.com/minad/consult/issues/636.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -602,8 +602,9 @@ Meant to be be added to `completion-setup-hook'."
   ;; available in the variable standard-output
   (embark--cache-info standard-output)
   (when (minibufferp completion-reference-buffer)
-    (setf (buffer-local-value 'embark--type standard-output)
-          (completion-metadata-get (embark--metadata) 'category))))
+    (with-current-buffer standard-output
+      (setq embark--type
+            (completion-metadata-get (embark--metadata) 'category)))))
 
 ;; We have to add this *after* completion-setup-function because that's
 ;; when the buffer is put in completion-list-mode and turning the mode


oantolin:
Why would they deprecate (setf buffer-local-value)? I don't really like this. Let me read up on the rationale before doing anything about it.

minad:
I don't think the deprecation of these variables is a good idea or justified, but whatever. These setters have side effects. cl-letf is not their only use case, but they are only deprecated because of the side effects there. Furthermore the (alist-get) gv also has a bad side effect but is used far too widely to be deprecated. All in all, a bad development...

For now I've refused merging this in packages. :-P

minad:
See https://github.com/minad/consult/issues/636#issuecomment-1229481271 where I also complained about this. But I am too lazy to debate with the maintainers. They are stubborn anyway. :-P

minad:
> I really have no opinion on this matter. All I know is that nothing changes upstream, all users will encounter at least one warning whenever an embark (or vertico, etc.) command is invoked.

No, the warning will only occur during package installation/compilation.

oantolin:
Reading the [mailing list thread](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=26624) linked to by @tarsius in #546 has made me change my mind.

oantolin:
Thanks @phikal and @tarsius!

minad:
> Reading the [mailing list thread](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=26624) linked to by @tarsius in https://github.com/oantolin/embark/pull/546 has made me change my mind.

@oantolin May I ask which argument in the thread convinced you? For `(setf (alist-get ...) ...)`, the corresponding `cl-letf` will also have side effects if you don't set the REMOVE argument.

~~~ elisp
(let ((list nil))
  (cl-letf (((alist-get 'key list) 123))
    nil)
  (message "LIST %S" list))
~~~

On the other hand if the key/value pair exists, there is no side effect, as in this case if the variable is already buffer-local before the `cl-letf`. I agree that the semantics are not nice but one could have also solved this differently:

1. Check that the pre-condition holds which makes `cl-letf` side-effect free.
2. Extend gv and cl-letf to store more state which allows fully recovery.

oantolin:
Oh, it's not really the case that some argument convinced me that `(setf buffer-local-value)` should be removed. What I meant when I said reading the thread made me change my mind is just that I previously did not know there was _any_ reason for deprecating `(setf buffer-local-value)`, and thought it was just arbitrary. Now I know there is at least some problem with it so the deprecation is not arbitrary; and thus it is unlikely to be reversed, which in turn means this change will need to be made eventually.

Personally I'd be in favor of fixing the issue of localness not being restored. I'd be quite happy with your option 2: extend gv and cl-letf to store more state.

PR #535, opened 2022-08-28
Add a .dir-locals.el file to avoid white space issues

phikal:
When contributing #534, I had first pushed a commit that used tabs, even though the entire file is indented using spaces.  I'd suggest adding a file like the one in this patch to avoid issues like these in the future.


oantolin:
Sure, why not? Thank you.

PR #541, opened 2022-09-14
Register marginalia--cache-reset (Fix #540)

### LINKED ISSUE #540: Update annotations, refresh UI after action (not the same as embark--restart)
Regarding the restart, there is a general problem when annotations must be updated (e.g. for the set-variable action). Maybe we could come up with some mechanism which ensures that the completion UI is updated after each action (not restarted, only updated). Also the Marginalia cache should be cleared.

_Originally posted by @minad in https://github.com/oantolin/embark/issues/538#issuecomment-1245582230_

minad:


minad:
Hmm, I just realized that we already have a little bit of Embark-specific code in Marginalia. See https://github.com/minad/marginalia/blob/db12d80307ea9d6de44b59d18533c9cbf0c51724/marginalia.el#L76. So do you prefer to add this to Marginalia instead?

~~~ elisp
(defvar embark-post-action-hooks)
(with-eval-after-load 'embark
  (push #'marginalia--cache-reset (alist-get :always embark-post-action-hooks)))
~~~

oantolin:
Are there any actions for which we were using `embark--reset` just for the annotations? We should reserve it for actions likely to change the candidate list.

oantolin:
Nope, I think the current `embark--restart` uses are all warranted.

oantolin:
Merged, thanks!

PR #542, opened 2022-09-16
Check `category` is not nil in `embark-target-top-minibuffer-completion`

mattbeshara:
I encounter a bug using `embark-act` from the minibuffer.  I do `eval-expression`, type `(cons` and then invoke `embark-act`.  For me, this causes `Act on nil` to appear at the top of the `*Embark Actions*` buffer.  With this change, the `*Embark Actions*` buffer shows `Act on function` and allows me to use actions for function as I’d expect.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -856,7 +856,8 @@ their own target finder.  See for example
                              (substring contents
                                         0 (or (cdr (last completions)) 0))
                              (car completions)))))))
-      (cons category (or (car (member top candidates)) top)))))
+      (when category
+        (cons category (or (car (member top candidates)) top))))))
 
 (defun embark-target-collect-candidate ()
   "Target the collect candidate at point."


minad:
This doesn't look like a proper fix since many target finders may return nil as category, see:

https://github.com/oantolin/embark/blob/1c2ee48ec5e091b49ebf7e55d570a2bc061bca5c/embark.el#L3269

Instead Embark should somewhere use `(or category 'general)`, probably here:

https://github.com/oantolin/embark/blob/02c4b88f25fb4f4f17a553344875366117ccd2e1/embark.el#L1974

mattbeshara:
Alright, well if my suggested fix would just end up breaking more things, then for now I can just remove `embark-target-top-minibuffer-completion` from my `embark-target-finders`.  I use Vertico as well anyway, so if this target finder is only intended for the default completion UI then it probably won't impact me negatively.

oantolin:
I agree with @minad that this fix isn't quite correct since many calls to `completing-read` don't specify a completion category (and that embark should treat those as type `general`, not `nil`). But I do think that the behavior you found should be considered a bug. In my opinion the real bug is that `embark-target-top-minibuffer-completion` is just using `(minibufferp)` as a test for determining whether one is in a completion session.

So I'll make two small changes: change that test to `(and (minibufferp) minibuffer-completion-table)`, and take up @minad's suggestion of using `general` by default instead of `nil`. 476f2ae8cd9ff2bb1f819352ef32d0aace77be48

mattbeshara:
That seems to fix the problem, thank you!

PR #538, opened 2022-09-13
Add toggle variable action

minad:

PR #543, opened 2022-09-17
embark-toggle-variable: Toggle local variable

minad:
Use set instead of set-default.


minad:
@oantolin This is a follow up to #538. I think we want to make sure that local variables are toggled.

oantolin:
Funny: I was about to do this myself when I realized I really wanted to toggle `sentence-end-double-space` only locally.

PR #546, opened 2022-10-10
Silence byte-compiler

tarsius:
Since Emacs 29 `buffer-local-value` is deprecated as a generalized variable, because it has issues.  https://debbugs.gnu.org/cgi/bugreport.cgi?bug=26624.


oantolin:
I'm not sure why GitHub doesn't notice this is exactly the same change that I just merged in #534, but it is. So I'm closing without merging, but because the change has already been made. Thanks!

PR #550, opened 2022-10-20
Add a warning in case embark-consult cannot be loaded

minad:
A missing embark-consult package seems to be a common issue. Warn about it.

### DIFF
--- a/README.org
+++ b/README.org
@@ -189,10 +189,10 @@ sets:
   package menu mode.
 
   If you use the grepping commands from the [[https://github.com/minad/consult/][Consult]] package,
-  =consult-grep=, =consult-git-grep= or =consult-ripgrep=, then you'll
-  probably want to install and load the =embark-consult= package, which
-  adds support for exporting a list of grep results to an honest
-  grep-mode buffer, on which you can even use [[https://github.com/mhayashi1120/Emacs-wgrep][wgrep]] if you wish.
+  =consult-grep=, =consult-git-grep= or =consult-ripgrep=, then you should
+  install the =embark-consult= package, which adds support for exporting a
+  list of grep results to an honest grep-mode buffer, on which you can
+  even use [[https://github.com/mhayashi1120/Emacs-wgrep][wgrep]] if you wish.
 
 When in doubt choosing between exporting and collecting, a good rule
 of thumb is to always prefer =embark-export= since when an exporter to a
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -24,12 +24,8 @@
 
 ;;; Commentary:
 
-;; This package provides integration between Embark and Consult. To
-;; use it, arrange for it to be loaded once both of those are loaded:
-
-;; (with-eval-after-load 'consult
-;;   (with-eval-after-load 'embark
-;;     (require 'embark-consult)))
+;; This package provides integration between Embark and Consult. The package
+;; will be loaded automatically by Embark.
 
 ;; Some of the functionality here was previously contained in Embark
 ;; itself:
--- a/embark.el
+++ b/embark.el
@@ -3259,6 +3259,14 @@ PRED is a predicate function used to filter the items."
 
 ;;; Integration with external completion UIs
 
+;; consult
+
+(with-eval-after-load 'consult
+  (unless (require 'embark-consult nil 'noerror)
+    (warn "The package embark-consult should be installed if you use both Embark and Consult")))
+
+;; marginalia
+
 ;; Ensure that the Marginalia cache is reset, such that
 ;; `embark-toggle-variable-value' updates the display (See #540).
 (with-eval-after-load 'marginalia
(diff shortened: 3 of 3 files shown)


minad:
It is a bit ugly to special case embark-consult among the possible integration packages. However this issue has come up multiple times (in contrast to integrations of other packages) and embark-consult is maintained as part of the Embark repository, so it may make sense to be a bit pragmatic here. What do you think?

oantolin:
Yes, let's do this. I think both of us have had to remind people to load embark-consult many times.

oantolin:
Oh, no! 🤣 Ironically this will trigger tons of bogus bug reports: now using `(use-package embark-consult :after (embark consult))` as recommended in the README causes a recursive loading error!

I'll fix the README and brace for impact.

minad:
Oh right. Please try the fix in https://github.com/oantolin/embark/pull/551.

oantolin:
Thanks for #551!

archer-65:
> Oh, no! rofl Ironically this will trigger tons of bogus bug reports: now using `(use-package embark-consult :after (embark consult))` as recommended in the README causes a recursive loading error!
> 
> I'll fix the README and brace for impact.

I'm using setup.el:

```elisp
(setup (:straight embark-consult)
  (:hooks embark-collect-mode consult-preview-at-point-mode))
```

The warning is here :D
Suggestions?

oantolin:
I'm not familiar with `setup.el`, @archer-65, and in particular I don't know if those lines you quote  merely install `embark-consult` or also load it. With the latest versions of `embark` and `embark-consult` you don't need to explicitly load `embark-consult`, just have it installed: `embark` will arrange for `embark-consult` to be loaded after `consult` if `consult` ever gets loaded.

Could you give a little more detail on the problem you are experiencing? Then we can try to figure out how to translate the recommended setup to `setup.el`.

archer-65:
> I'm not familiar with `setup.el`, @archer-65, and in particular I don't know if those lines you quote  merely install `embark-consult` or also load it. With the latest versions of `embark` and `embark-consult` you don't need to explicitly load `embark-consult`, just have it installed: `embark` will arrange for `embark-consult` to be loaded after `consult` if `consult` ever gets loaded.
> 
> Could you give a little more detail on the problem you are experiencing? Then we can try to figure out how to translate the recommended setup to `setup.el`.

The lines are expanded like:
```elisp
(straight-use-package 'embark-consult)
(with-eval-after-load 'embark
  (with-eval-after-load 'consult 
    (require 'embark-consult)
    ;; configuration
    ))
```

The package is working without any issue. Regardless, when I start Emacs, the warning you were talking about appears.

oantolin:
Do you mean you get the warning added in this PR? "The package embark-consult should be installed if you use both Embark and Consult"? Or do you mean the recursive loading error I mentioned?

archer-65:
> Do you mean you get the warning added in this PR? "The package embark-consult should be installed if you use both Embark and Consult"? Or do you mean the recursive loading error I mentioned?

I solved my issue "installing" (`straight-use-package`) `embark-consult` BEFORE `consult`.
I have to be honest: I don't like `with-eval-after-load` in external dependencies.

oantolin:
I wonder why you were getting the warning though. Presumably you are not actually installing all packages every time you start Emacs, right? Maybe `straight` doesn't put the directory where it puts the package on the `load-path`? That sounds a little crazy. I don't understand exactly why you got the warning if you did have the package installed.

archer-65:
> I wonder why you were getting the warning though. Presumably you are not actually installing all packages every time you start Emacs, right? Maybe `straight` doesn't put the directory where it puts the package on the `load-path`? That sounds a little crazy. I don't understand exactly why you got the warning if you did have the package installed.

I read `embark`'s code: loading `consult` before of `(straight-use-package 'embark-consult)` causes the error.

Your `(with-eval-after-load 'consult)` block is triggered after `consult`'s loading, so I moved my `embark` related configuration before `consult` and solved the issue. 

A warning is not an error, but in this case it is annoying.

minad:
@archer-65 This doesn't sound correct. Embark checks if embark-consult is available. If it is not available, then you get the warning. Straight or your configuration is at fault here.

archer-65:
> @archer-65 This doesn't sound correct. Embark checks if embark-consult is available and loads it. If it is not available, then you get the warning. Straight or your configuration is at fault here.

I have edited my previous message, it is `straight` fault indeed (related to the recent changes in `embark`'s code).
The problem started after the update of `embark` I made a few days ago, and I didn't understand the the reason until I read the code :D.

hrehfeld:
I still run into this with `straight` and `:after`. 

What's the recommended way to load `consult`, `embark` and `embark-consult` with `straight`?

oantolin:
I'm sorry, I'm not familiar with straight and don't know how to answer, @hrehfeld. Did you try putting `(straight-use-package 'embark-consult)` before loading consult as @archer-65 said worked for him?

PR #551, opened 2022-10-23
Move warning

minad:


oantolin:
Ooh, let me test this and if it works I'll rollback 4ce430c04c667bba322f14d2119890ddaa04dfdd

oantolin:
Ah, OK, this works. And I should keep 4ce430c too, right? It is automatically loaded by Embark, but with this move, loading it in your config still works.

PR #552, opened 2022-10-29
embark-org: transform attachment link to org-file-link

ilupin:
Hi, attachment links are basically the same as file links, so it would be convenient to support attachment links in `embark-org'.

> ‘attachment’
>      Same as file links but for files and folders attached to the
>      current node (see *note Attachments::).  Attachment links are
>      intended to behave exactly as file links but for files relative to
>      the attachment directory.

I've signed the FSF copyright assignment.


oantolin:
Good idea, thank you!

PR #558, opened 2022-11-11
Distribute embark-org, Linting

minad:


oantolin:
Thanks for the strong hint it was time to publish embark-org. I guess I need to do a PR on MELPA, too.

.elpaignore
```diff
@@ -1 +1 @@
-embark-org.el
\ No newline at end of file
+LICENSE
```
oantolin, line 1: Wait, is ignoring the LICENSE correct?

oantolin: Or am I going to get a distressed email from Stefan about removing the LICENSE file? :)

minad: Yes, it is fine. Stefan added many LICENSE ignores in elpa-packages himself.

minad:
> I guess I need to do a PR on MELPA, too.

Indeed, MELPA only distributes embark.el. 

https://github.com/melpa/melpa/blob/master/recipes/embark#L3

.elpaignore
oantolin: OK, good. I thought you probably knew for a fact this was OK, but wanted to confirm just in case.

oantolin:
I made a PR on MELPA: melpa/melpa#8294

Issue #554, opened 2022-11-01
wgrep-mode doesn't  work with consult-line export


agzam:
It seems that wgrep-mode is no longer works with buffers exported from `consult-line`. Embark export of `consult--grep` (multiple files) works fine. But `consult-line` (single file) doesn't.

## Repro steps

- Run consult-line with multiple occurrences in any normal, writable file
- Export the results with `embark-export`
- Run `wgrep-change-to-wgrep-mode` 

### Observed

```
Debugger entered--Lisp error: (wrong-type-argument markerp nil)
  occur-after-change-function(1 285 284)
  put-text-property(1 285 read-only t)
  wgrep-prepare-header/footer()
  wgrep-prepare-to-edit()
  wgrep-change-to-wgrep-mode()
  funcall-interactively(wgrep-change-to-wgrep-mode)
  command-execute(wgrep-change-to-wgrep-mode record)
  execute-extended-command(nil "wgrep-change-to-wgrep-mode" nil)
  funcall-interactively(execute-extended-command nil "wgrep-change-to-wgrep-mode" nil)
  command-execute(execute-extended-command)
```

If you run `wgrep-change-to-wgrep-mode` again, it appears to be changing to wgrep-mode, but the export buffer remains read-only.

### Versions

Emacs: 
GNU Emacs 29.0.50 (build 1, x86_64-apple-darwin21.6.0, NS appkit-2113.60 Version 12.6 (Build 21G115)) of 2022-10-21

Consult: 3226c1a01e0a2a4352b2c885b8f4eb22610d99af
Embark: 3a86e4ba70e2e35c2d833016cd1e5b1fd8d8416d
Emacs-wgrep: f9687c28bbc2e84f87a479b6ce04407bb97cfb23

oantolin:
Export from consult-line does not produce a grep-mode buffer but rather an occur-mode buffer. To edit those use the built-in occur-edit-mode not wgrep. It is bound to `e` in any occur-mode buffer.

artemkovalyov:
You can benefit from this alternative exporter if you still want to use wgrep with consult-line

https://github.com/oantolin/embark/pull/560/files#diff-f584aec8c356d120612b2236de2a31a953f48f69b5ddbfc28d89ab0e4857c615

Issue #47, opened 2020-12-12
[feature request] Make  "*Embark Occur*" supporting wgrep for some special command 


jixiuf:
Xref provides helpful commands for code navigation and discovery, such as `xref-find-definitions` and `project-find-regexp` ,and   `project-find-regexp` support `ripgrep` now( on emacs master) .

When a xref query hits multiple candidates Emacs shows a Xref
buffer with the various candidates and their location.

On emacs master, it support using `completing-read` showing the multiple candidates  now with `xref--show-defs-minibuffer`

https://lists.gnu.org/archive/html/emacs-devel/2020-11/msg00707.html

```
(with-eval-after-load 'xref
  (setq xref-search-program 'ripgrep)     ;default `grep`,  for project-find-regexp
  (when (functionp 'xref--show-defs-minibuffer)
    (setq xref-show-definitions-function 'xref--show-defs-minibuffer) ; using completing-read
    (setq xref-show-xrefs-function 'xref--show-defs-minibuffer)))

```
this is a screenshot of `project-find-regexp` in minibuffer
<img width="654" alt="image" src="https://user-images.githubusercontent.com/461743/101976673-a2392080-3c82-11eb-8278-f8768001e0be.png">

And I can export the result to   "*Embark Occur*" with `embark-occur` 

<img width="714" alt="image" src="https://user-images.githubusercontent.com/461743/101976721-f5ab6e80-3c82-11eb-9105-c9025e88b6ca.png">

So I think it is great for "*Embark Occur*" support `wgrep` like ivy-occur/grep,
I think https://github.com/minad/consult/issues/46 can benefit from it too, once consult-rg/ag is implemented.

oantolin:
I wouldn't want to reinvent `wgrep` here. So I think the best option is to not use `embark-occur`, but rather make it so `embark-export` works with these functions you mentioned, and produces an actual grep buffer you can use with the standard wgrep.

jixiuf:
it’s ok for me

manuel-uberti:
I second @jixiuf enthusiasm on this. :)

jixiuf:
```
(require 'wgrep)

;;;###autoload
(defun embark-occur-wgrep-setup ()
  (set (make-local-variable 'wgrep-header/footer-parser)
       'embark-occur-wgrep-prepare-header/footer)
  (wgrep-setup-internal))

(defun embark-occur-wgrep-prepare-header/footer ()
  (let ((beg (point-min))
        (end (point-min)))
[...]

PR #560, opened 2022-11-21
Export location based consult results to grep buffer.

artemkovalyov:
I  prefer `wgrep` over `occur-edit-mode`.  I created an alternative exporter for `consult-lines`  to use grep mode buffer.
The benefits are:

- using wgrep and its multi-line results editing capabilities
- uniformity of results editing user experience for grep-like buffers.

I prefer a grep buffer for `consult-lines` because I use it a lot for narrowing down searches in an open buffer as opposed to grep-based searching tools from consult that constantly search over a directory. In-place editing is richer with wgrep and using one tool makes you ponder less about key combinations and how you enable in-place editing.

Related issues:

- #47 
-  #554

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -103,7 +103,7 @@
 (setf (alist-get 'consult-location embark-default-action-overrides)
       #'embark-consult-goto-location)
 
-(defun embark-consult-export-occur (lines)
+(defun embark-consult-export-location-occur (lines)
   "Create an occur mode buffer listing LINES.
 The elements of LINES are assumed to be values of category `consult-line'."
   (let ((buf (generate-new-buffer "*Embark Export Occur*"))
@@ -148,6 +148,64 @@ The elements of LINES are assumed to be values of category `consult-line'."
       (occur-mode))
     (pop-to-buffer buf)))
 
+(defun embark-consult-export-location-grep (lines)
+  "Create an occur or grep mode buffer listing LINES.
+The elements of LINES are assumed to be values of category `consult-location'."
+  (let ((buf (generate-new-buffer "*Embark Export Grep*"))
+        (count 0)
+        prop)
+    (with-current-buffer buf
+      (insert (propertize "Exported line search results (file-backed buffers only):\n" 'wgrep-header t 'font-lock-face '(:weight bold)))
+      (save-excursion (let (last-buf
+                            filename
+                            (not-saved-buffers '()))
+                        (dolist (line lines)
+                          (pcase-let*
+                              ((`(,loc . ,num) (consult--get-location line))
+                               (lineno (format "%d" num))
+                               (contents (embark-consult--strip line))
+                               (this-buf (marker-buffer loc))
+                               (file-path (buffer-file-name this-buf)))
+                            (if file-path
+                                ;; Handle `consult-line-multi' scenario where many buffers contain matches
+                                (progn (unless (eq this-buf last-buf)
+                                         (setq last-buf this-buf)
+                                         (setq filename (file-name-nondirectory file-path))
+                                         ;; Start listing matches for a new file
+                                         (insert "\n" (propertize (concat "file: " filename) 'wgrep-ignore t 'font-lock-face '(:inherit compilation-info :underline t)) "\n"))
+                                       (insert (propertize
+                                                (concat file-path ":")
+                                                'invisible t)
+                                               lineno
+                                               ":"
+                                               contents "\n" ))
+                              (cl-pushnew 'not-saved-buffers this-buf))))
+                        ;; Show a footer warning listing not saved buffers
+                        (when not-saved-buffers
+                          (save-excursion
+                            (insert "\n\nSome " (propertize "buffers" 'font-lock-face '(:weight bold)) " are not visiting (saved to) a file and are missing from exported results:\n")
+                            (dolist (nsbuf not-saved-buffers)
+                              (insert "- " (buffer-name nsbuf) "\n"))
+                            (insert "Either save the buffers or use `embark-consult-export-location-occur' as your export adapter`")
+                            ;; Send simplified information about unsaved buffers to Messages
+                            (message "This exporter requires the following buffers to be saved first %s" not-saved-buffers))
+                          ;; Add footer properties to a warning so that it is skipped by `wgrep'
+                          (add-text-properties (point) (point-max) '(read-only t wgrep-footer t front-sticky t)))))
+      (grep-mode)
+      (when (> count 0)
+        (setq-local grep-num-matches-found count
+                    mode-line-process grep-mode-line-matches))
+      ;; Make this buffer current for next/previous-error
+      (setq next-error-last-buffer buf)
+      ;; Set up keymap before possible wgrep-setup, so that wgrep
+      ;; restores our binding too when the user finishes editing.
+      (use-local-map (make-composed-keymap
+                      embark-consult-revert-map
+                      (current-local-map)))
+      (setq-local wgrep-header&footer-parser #'embark-consult--wgrep-prepare)
+      (when (fboundp 'wgrep-setup) (wgrep-setup)))
+    (pop-to-buffer buf)))
+
 (defun embark-consult--upgrade-markers ()
   "Upgrade consult-location cheap markers to real markers.
 This function is meant to be added to `embark-collect-mode-hook'."
@@ -156,8 +214,10 @@ This function is meant to be added to `embark-collect-mode-hook'."
       (when (car entry)
         (consult--get-location (car entry))))))
 
+;; Set default `occur-mode' based exporter for consult-line, consult-line-multi, consult-outline and alike
+;; Another option is using grep-mode by using `embark-consult-export-location-grep'
 (setf (alist-get 'consult-location embark-exporters-alist)
-      #'embark-consult-export-occur)
+      #'embark-consult-export-location-occur)
 (cl-pushnew #'embark-consult--upgrade-markers embark-collect-mode-hook)
 
 ;;; Support for consult-grep
@@ -253,12 +313,12 @@ This function is meant to be added to `embark-collect-mode-hook'."
                        'minibuffer-exit-hook
                        (lambda ()
                          (throw 'xref-items
-                           (xref-items
-                            (or
-                             (plist-get
-                              (embark--maybe-transform-candidates)
-                              :candidates)
-                             (user-error "No candidates for export")))))
+                                (xref-items
+                                 (or
+                                  (plist-get
+                                   (embark--maybe-transform-candidates)
+                                   :candidates)
+                                  (user-error "No candidates for export")))))
                        nil t))
                   (consult-xref fetcher))))))
         `((fetched-xrefs . ,(xref-items items))


artemkovalyov:
@oantolin , what would be your recommendation for a more granular configuration to keep other tools like consult-outline, etc. , still using `occur`.  With the current setup, other consult-location based tools will also use grep buffer for exporting. I don't mind that at all but would like to consult about your vision for this.

minad:
Thank you for this. It is probably a good idea to provide alternative exporters.

However I have two questions or counter points I'd like to see discussed or sorted out first:

1. Are the technical reasons strong enough to justify the addition?

Isearch also exports to occur, we cannot argue the grep/occur distinction away. Embark/Consult and the other packages are somewhat opionated. We try to avoid needless duplications and stay minimal. Therefore I wonder what specific aspects of grep are preferrable for you. Are there strong technical reasons? You can probably configure both occur and grep to behave similarly, if this is only about key bindings. I am kind of sceptical regarding arguments based on taste in contrast to technical reasons. Also note that the grep buffer is weaker out of the box than the occur buffer, if wgrep doesn't exist. Otherwise my suggestion would be to add this to the wiki.
 
2. Is it reliable given that buffers and files are not the same?

Using grep/wgrep on open buffers worries me a bit and may even lead to inconsistencies between the file and buffer. Are you sure that this behaves correctly? What happens if you use consult-line on a modified bbuffe and export? What happens then if you wgrep?

oantolin:
I'll read through this conversation carefully later, but my first reaction is that it doesn't really make sense to have `consult-line` export to a grep-mode buffer because `consult-line` is meant to search through a buffer whether or not that buffer is visiting a file (and I, for one, do use it very often on non-file buffers!); `grep-mode` on the other hand is meant to list results only in files. So my first instinct is to reject this PR on semantic grounds: it seems actually incorrect to me to do this.

Of course, if you do not want the capability of exporting `consult-line` results from non-file buffers, you are free to reject this capability in your personal configuration, and since other may also want to forgo this feature in exchange for `grep-mode` it would be nice to have this documented in the wiki.

oantolin:
On second thought, if we keep occur as the default and document the limitations of exporting to grep, I don't see why we couldn't keep both exporters.

oantolin:
OK, I'm reading the conversation here now (which is not as long as I thought it was, there is a large screenshot which made me overestimate the length :)).

> what would be your recommendation for a more granular configuration to keep other tools like consult-outline, etc. , still using occur

I'd suggest using marginalia to have `consult-line` report a different category, say `consult-line-location`, and configure the exporter only for that.

I see @minad brought up the same concern I have.

I'm also curious about what is better about `wgrep` compared to `occur-edit-mode`. I've always thought of `wgrep` as implementing `occur-edit-mode` but for grep. You mention "multiline edits", but I'm not sure what you mean by that, I can think of two meanings but both work fine here:

1. Maybe you meant editing an occur buffer containing multiline results (as obtained by calling `occur` with a numeric prefix parameter specifying the number of context lines). That works here just fine, and it is probably not what you meant since there is no way to get embark export to give you context lines.

2. Maybe you meant editing several different single-line results in an occur-mode buffer. That also works perfectly well here and I would say is probably the main use case for `occur-edit-mode`. It would be weird if that didn't work.

So either something is a little off in your configuration (and we can certainly try to figure out what), or you meant a third thing by "multiline edit".

Continuing, you say "I had to dig into google search and docs to figure out inline editing with occur", which is a perfectly validway of finding out about it, but I want to point out that: (1) that would be the same for `wgrep`, people need to find out about it somehow in order to use it; (2) for `wgrep` finding out *necessarily* involves the internet, but `occur-edit-mode` is built-in and well-documented inside Emacs, so no googling is *required* to find it. 

I'm glad you looked into @minad's questions, and your answers would form the backbone of documenting the pitfalls of using `grep-mode` export for `consult-line`. I do have one minor correction, you say "grep buffer is weaker - for search-replace it's not". But it definitely is weaker at least in the sense that editing requires the third-party `wgrep` package, while `occur-edit-mode` is built-in.

I think that on balance, the first thing to figure out is why whatever you meant by "multiline edits" are not working for you with `occur-edit-mode`. If we fix that you may not even feel a need for grep export anymore.

minad:
@oantolin Iirc there was a feature request for export with context lines to occur.

oantolin:
> Do you know cases when people rely on the export to navigate the file? 

Forgot to answer this: I do that fairly often. I use it when I need to make a bunch of changes everywhere some string appears, for example, if I'm programming and changed a function signature and need to consider how best to update each call-site, or if I'm writing some math and decided to change the name of a concept I'm defining. What I like about using occur for these situations is that the occur buffer works like a temporary todo list containing all the spots where I need to change something; and I can navigate among the spots with `next-error` and `previous-error`, I don't even need to switch to the occur mode buffer.

Of course, in these use cases I'm editing an actual file so grep would do just as well. But I do sometimes use `consult-line` in the `list-packages` buffer and export to an occur-mode buffer which I treat as a todo list of packages I wish to check out (I do this when I want to search both package names and one-line descriptions, for just packages names I prefer to use `describe-package`). But this use case could easily be replaced by `consult-focus` (which I also sometimes use for the same purpose, I'm inconsistent that way).

oantolin:
@minad said:

> Not sure if we had this discussion before but maybe we also need a way to select between multiple exporters.

You mean selecting at export time? That could be a nice to have. A first idea on the UI would be that if `(alist-get embark-exporters-alist type)` is a list (whose `car` is not `lamba` or `closure`... 🙄), then you get prompted for which exporter you want. But it may be premature to do that, we seem to have managed with a single "best" exporter for each type of targets so far. In this particular case, I'm still not sure I understand why `occur-edit-mode` isn't good enough for @artemkovalyov.

> Iirc there was a feature request for export with context lines to occur.

Yes, there was, and I do think that's a good idea, I just forgot about it. I guess a first step would be to pass the prefix argument to the exporter function somehow, probably just as the value of `prefix-arg` (which I think currently gets reset to `nil` before the exporter even runs).

minad:
@oantolin Yes, I'd also like to understand the real value of this addition better and why @artemkovalyov isn't happy with occur. I don't think you can "run away" from occur given that it is also baked into isearch. Just adding it because we can and because of some consistency details doesn't seem justified (we have the wiki for that), considering that we introduce new problems given non-file backed buffers and given that wgrep is needed to reach parity with occur.

artemkovalyov:
@oantolin, I think you didn't get what multi-line edit means. It's about the ability to add lines in an exported buffer that are committed to the original buffer (file). 

Here are examples with Wgrep and occure-edit

- In Wgrep new lines are handled perfectly well and that's a very useful feature (at least to me)
image

- In occur-edit I can only edit the exported line, I can't remove it or add new lines
image

minad:
@artemkovalyov That's indeed an interesting feature! I wasn't aware of that. I think occur-edit-mode should be improved such that this is also supported.

oantolin:
You're absolutely right that I hadn't understood what you meant, @artemkovalyov! (I suspected as much.) That is a cool feature I did not know wgrep had! Like @minad I think the right approach is to enhance occur-edit-mode to support this, but in the meantime, I do think that for that feature it is worth having the grep exporter even if it only works properly for file-backed buffers.

README.org
```diff
@@ -990,6 +993,14 @@ The =embark-consult= package provides the following exporters:
   to navigate among matches, and, if you install the [[http://github.com/mhayashi1120/Emacs-wgrep/raw/master/wgrep.el ][wgrep]] package,
   you can use it to edit the matches in place.
 
+- An alternate exporter for =consult-line= allows to use it with grep buffer
+  and benefit from wgrep and its multi-line results editing capabilities.
```
minad, line 997: @artemkovalyov Can you please clarify what you meant by multi-line here. I suspect @oantolin and I are not the only ones who are not aware of this feature of wgrep.

embark-consult.el
```diff
@@ -181,6 +181,37 @@ This function is meant to be added to `embark-collect-mode-hook'."
...
+      (goto-char (point-min))
+      (grep-mode)
+      (setq next-error-last-buffer buf)
+      ;; Set up keymap before possible wgrep-setup, so that wgrep
+      ;; restores our binding too when the user finishes editing.
+      (use-local-map (make-composed-keymap
+                      embark-consult-revert-map
+                      (current-local-map)))
+      (setq-local wgrep-header/footer-parser #'ignore)
+      (when (fboundp 'wgrep-setup) (wgrep-setup)))
```
minad, line 212: Is there some code here which could be shared with `embark-consult-export-grep`? Maybe one could introduce a helper function `embark-consult--export-grep` which sets up the grep buffer is passed a function as argument. The function argument should insert the actual content into the buffer.

~~~ elisp
(defun embark-consult--export-grep (insert)
  (let ((buf (generate-new-buffer "*Embark Export Grep*")))
    (with-current-buffer buf
      (insert (propertize "Exported grep results:\n\n" 'wgrep-header t))
      (funcall insert)
      (goto-char (point-min))
      (grep-mode)
      ;; Make this buffer current for next/previous-error
      (setq next-error-last-buffer buf)
      ;; Set up keymap before possible wgrep-setup, so that wgrep
      ;; restores our binding too when the user finishes editing.
      (use-local-map (make-composed-keymap
                      embark-consult-revert-map
                      (current-local-map)))
      (setq-local wgrep-header/footer-parser #'ignore)
      (when (fboundp 'wgrep-setup) (wgrep-setup)))
    (pop-to-buffer buf)))

(defun embark-consult-export-grep (lines)
  ...)
~~~

embark-consult.el
```diff
@@ -181,6 +181,37 @@ This function is meant to be added to `embark-collect-mode-hook'."
       (when (fboundp 'wgrep-setup) (wgrep-setup)))
     (pop-to-buffer buf)))
 
+(defun embark-consult-export-lines-to-grep (lines)
```
minad, line 184: Any ideas for a better name?

embark-consult.el
oantolin: I think that instead of the function you propose @minad, that this new export function should just build a grep-formatted list of lines and then call the existing `embark-consult-export-grep`.

minad: I thought a callback would be more efficient, since we can write directly to the buffer. Building an intermediate list requires allocations, which can be costly if we have many candidates. It would be even worse to build many intermediate strings.

oantolin: You're right, of course. I just thought it shouldn't be a problem since I'd expect few matching lines compared to the number of lines in the buffer, so the slow step would actually be the initial gathering of candidates in consult-line. I mostly wanted to reuse more code, in particular to reuse the line insertion loop, so maybe the best option is a callback but one that is called per line.

minad:
On 11/22/22 18:10, Omar Antolín Camarena wrote:
> ***@***.**** commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In embark-consult.el
> <https://github.com/oantolin/embark/pull/560#discussion_r1029619030>:
> 
>> +             (contents (propertize (embark-consult--strip line)))
> +             (this-buf (marker-buffer loc)))
> +          (unless (eq this-buf last-buf)
> +            (insert (propertize "Exported grep results:\n\n" 'wgrep-header t))
> +            (setq last-buf this-buf))
> +          (insert (concat (file-name-nondirectory (buffer-file-name this-buf)) ":" lineno contents "\n"))))
> +      (goto-char (point-min))
> +      (grep-mode)
> +      (setq next-error-last-buffer buf)
> +      ;; Set up keymap before possible wgrep-setup, so that wgrep
> +      ;; restores our binding too when the user finishes editing.
> +      (use-local-map (make-composed-keymap
> +                      embark-consult-revert-map
> +                      (current-local-map)))
> +      (setq-local wgrep-header/footer-parser #'ignore)
> +      (when (fboundp 'wgrep-setup) (wgrep-setup)))
> 
> You're right, of course. I just thought it shouldn't be a problem since
> I'd expect few matching lines compared to the number of lines in the
> buffer, so the slow step would actually be the initial gathering of
> candidates in consult-line. I mostly wanted to reuse more code, in
> particular to reuse the line insertion loop, so maybe the best option is
> a callback but one that is called per line.

Hmm, I would probably go with the most efficient solution, a single
call.  I often export many candidates from consult-grep/ripgrep. As an
alternative we could also use a macro with a BODY.

embark-consult.el
artemkovalyov: I'll update the name after refactoring. This was my "code name" because I was looking for a solution specifically for lines. I didn't draw much difference from `consult-location` group back when I started my initial solution discovery.

artemkovalyov (commented):
Thanks for the productive discussions. I'll work on updating the PR.

oantolin:
Oh, @artemkovalyov have you signed the FSF copyright assignment papers?

README.org
artemkovalyov: @minad, I'll update the docs and add some wiki after we agree on the implementation. It's a pre-requisite.

artemkovalyov:
> Oh, @artemkovalyov have you signed the FSF copyright assignment papers?
I assume there's no GH plugin to do that? Do you have a simple instruction or link? Googling confuse me :)

artemkovalyov:
@minad, @oantolin, please, take another look at the implementation. I'm not an experienced LISPer but I tried to follow your recommendations. I had to create wrappers for actual exporters to preserve their APIs but split the implementation into a buffer setup and processing of results to make them reusable.

If you have better ideas, please, share. I also renamed an exporter to `occur` into a `location`, as the destination buffer can now be different. It seems to play nicely with the naming strategy and everything before that also uses `location` for the range of `location`-enabled consult functions.

artemkovalyov:
@minad, I made sure the buffers exported to grep buffer are visiting a file. Here are a couple of screenshots:
image
image
All the temporary buffers like *scratch* and others will be excluded from `consult-line-multi` for example. 
After we agree on the implementation and details I'll add some WIKI about this exporter.
The switch of the behavior is done via `defcustom` var for now.

embark-consult.el
```diff
@@ -66,6 +66,14 @@
 (eval-when-compile
   (require 'cl-lib))
 
+(defcustom embark-consult-export-location-candidates-to-grep-mode nil
+  "If set to true, `grep-mode' is used to export results of
+`consult-line', `consult-line-multi' and other `consult-location'
+enabled commands.
+
+If set to nil `occur-mode' is used."
+  :type 'boolean )
```
minad, line 75: This variable is not necessary. Users can simply adjust the exporter alist directly.

~~~ elisp
(setf (alist-get 'consult-location embark-exporters-alist) ...)
~~~

embark-consult.el
```diff
@@ -103,44 +111,89 @@
...
+  "Create an occur or grep mode buffer listing LINES.
+The elements of LINES are assumed to be values of category `consult-location'."
+  (if embark-consult-export-location-candidates-to-grep-mode
+      (funcall #'embark-consult--setup-grep-mode
+               "Exported Consult results\n\n"
+               lines
+               #'embark-consult--location-to-grep-results-writer)
+    (funcall #'embark-consult--setup-occur-mode
+             lines
+             #'embark-consult--location-to-occur-results-writer)))
```
minad, line 195: Instead of the customization variable, we should offer two different exporter functions:

embark-consult-export-location-grep
embark-consult-export-location-occur

embark-consult.el
```diff
@@ -163,11 +216,25 @@ This function is meant to be added to `embark-collect-mode-hook'."
...
+   "Exported grep results:\n\n"
+   lines
+   #'embark-consult--grep-results-writer))
+
+(defun embark-consult--grep-results-writer (lines)
+  "Simplistic writer to insert results produced byte
+`consult-grep' range of commands into a grep-mode buffer."
+  (dolist (line lines) (insert line "\n")))
+
+(defun embark-consult--setup-grep-mode (message lines results-writer)
```
minad, line 231: Do we use LINES somewhere? Otherwise we shouldn't pass it as an argument. Instead the RESULTS-WRITER can be a closure.

embark-consult.el
```diff
@@ -103,44 +111,89 @@
...
+  "Format the results from `consult-location' range of commands for the `grep-mode' as alternative to occur.
+You can edit results with `wgrep' in grep buffer including line deletion and insertion of new lines.
+The elements of LINES are assumed to be values of category `consult-location'."
+  (let (last-buf
+        filename)
+    (insert (propertize "Caution:" 'wgrep-header t 'font-lock-face '(:inherit 'compilation-warning :underline t))
+            (propertize " only buffers backed by a file are exported to grep-mode.\n" 'wgrep-header t 'font-lock-face '(:inherit 'compilation-warning)))
+    (insert (propertize "Hint:" 'font-lock-face '(:inherit 'compilation-info :underline t))
+            (propertize " use wgrep for editing of the results in place and saving them to respective files\n" 'font-lock-face '(:inherit 'compilation-info)))
+    (dolist (line lines)
```
minad, line 158: The hint should be removed. We didn't have that before and we also don't have it for the normal grep exporter.

embark-consult.el
```diff
@@ -103,44 +111,89 @@
...
+        (insert (concat lineno contents nl))))))
+
+(defun embark-consult--location-to-grep-results-writer (lines)
+  "Format the results from `consult-location' range of commands for the `grep-mode' as alternative to occur.
+You can edit results with `wgrep' in grep buffer including line deletion and insertion of new lines.
+The elements of LINES are assumed to be values of category `consult-location'."
+  (let (last-buf
+        filename)
+    (insert (propertize "Caution:" 'wgrep-header t 'font-lock-face '(:inherit 'compilation-warning :underline t))
+            (propertize " only buffers backed by a file are exported to grep-mode.\n" 'wgrep-header t 'font-lock-face '(:inherit 'compilation-warning)))
```
minad, line 155: The caution note isn't a bad idea, but I would prefer something less verbose, 
instead of "Exported Consult results\n\n" what about "Exported line results (file-backed buffers only):\n\n"?

embark-consult.el
```diff
@@ -103,44 +111,89 @@
...
+            (setq filename (file-name-nondirectory (buffer-file-name this-buf)))
+            (insert "\n" (propertize (concat "file: " filename) 'wgrep-ignore t 'font-lock-face '(:inherit compilation-info :underline t)) "\n"))
+          (insert (propertize
+                   (concat filename ":")
+                   'invisible t)
+                  lineno
+                  ":"
+                  contents "\n" ))))))
+
+(defun embark-consult--setup-occur-mode (lines resutls-writer)
```
minad, line 176: resutls typo

minad:
@artemkovalyov Thanks, this looks quite good. I've added a few comments, mostly details.

minad:
@artemkovalyov What's going on in the second screenshot of https://github.com/oantolin/embark/pull/560#issuecomment-1328360294? Why do we see the results of the embark export buffer? It shouldn't be backed by a file.

artemkovalyov:
> @artemkovalyov Thanks, this looks quite good. I've added a few comments, mostly details.
> 
> More importantly, please try to get the FSF assignment, see the description at https://protesilaos.com/emacs/modus-themes#h:111773e2-f26f-4b68-8c4f-9794ca6b9633.

On it. I'll update you when it's ready. Take another look at the code changes when you have time.

embark-consult.el
```diff
@@ -220,21 +210,18 @@ This function is meant to be added to `embark-collect-mode-hook'."
...
    "Exported grep results:\n\n"
-   lines
-   #'embark-consult--grep-results-writer))
-
-(defun embark-consult--grep-results-writer (lines)
+   (lambda ()
   "Simplistic writer to insert results produced byte
 `consult-grep' range of commands into a grep-mode buffer."
-  (dolist (line lines) (insert line "\n")))
+     ((dolist (line lines) (insert line "\n"))))))
```
oantolin, line 216: The two consecutive opening parentheses look wrong to me.

embark-consult.el
```diff
@@ -111,88 +103,83 @@ If set to nil `occur-mode' is used."
...
-               lines
-               #'embark-consult--location-to-grep-results-writer)
-    (funcall #'embark-consult--setup-occur-mode
-             lines
-             #'embark-consult--location-to-occur-results-writer)))
+  (funcall
+   #'embark-consult--setup-grep-mode
+    "Exported line search results (file-backed buffers only):.\n\n"
+   (lambda ()
+      "Format the results from `consult-location' range of commands for the `grep-mode' as alternative to occur.
```
oantolin, line 161: Is there any way the user can see these docstring for inner lambdas? I think there isn't, and if I'm right then these docstrings are only for readers of the source code, in which case they would be better off as comments.

oantolin:
This looks pretty good, @artemkovalyov, but I unfortunately I won't be able to merge it until you've signed the FSF copyright assignment, since Embark is now on GNU ELPA.

artemkovalyov:
> This looks pretty good, @artemkovalyov, but I unfortunately I won't be able to merge it until you've signed the FSF copyright assignment, since Embark is now on GNU ELPA.

Roger that, @oantolin, I've sent the form already and hope to get the agreement soon. In the meanwhile I'll install it from my fork to catch UX issues if any.

artemkovalyov:
@oantolin, I have my SFS assignment done, the number is RT: 1896958. I see the branch is in conflict now. I'll see what's changed and hopefully resolve them.

embark-consult.el
oantolin: Don't forget to fix this.

oantolin:
I'm glad you got the copyright assignment worked out! I don't think too much has changed so the code shouldn't be hard to update.

embark-consult.el
artemkovalyov:
```suggestion
     (dolist (line lines) (insert line "\n")))))
```

artemkovalyov:
@oantolin, I've brushed up my PR and tested both `grep` and `occur` based export to work properly for viewing and editing. While resolving merge conflicts I've seen you added highlighting, which is cool. I slightly updated the implementation to use `(save-excursion ...)` instead of `(goto-char ...)`. I hope you don't mind as it doesn't change the logic and puts the point right after the header when the buffer is ready.

embark-consult.el
```diff
@@ -103,44 +103,84 @@
...
+      (funcall results-writer)
       (goto-char (point-min))
       (occur-mode))
     (pop-to-buffer buf)))
 
+(defun embark-consult-export-location-occur (lines)
+  "Create an occur or grep mode buffer listing LINES.
+The elements of LINES are assumed to be values of category `consult-location'."
+  (funcall
+   #'embark-consult--setup-occur-mode
```
minad, line 119: Unnecessary funcall here.

embark-consult.el
```diff
@@ -103,44 +103,84 @@
...
+                       (format "lines from buffer: %s\n" this-buf)
+                       'face list-matching-lines-buffer-name-face))
+              (setq last-buf this-buf))
+            (insert (concat lineno contents nl))))))))
+
+(defun embark-consult-export-location-grep (lines)
+  "Create an occur or grep mode buffer listing LINES.
+The elements of LINES are assumed to be values of category `consult-location'."
+  (funcall
+   #'embark-consult--setup-grep-mode
```
minad, line 158: Same here.

minad:
@artemkovalyov  Thanks for the refactoring. From my side this looks good. However I haven't tested this yet. @oantolin can surely give you some additional comments.

embark-consult.el
artemkovalyov: You're right. I remember you suggested closures and I overlooked that there's an even simpler solution. The fix is on its way.

artemkovalyov:
@oantolin, here are some test results with both `consult-line` and `consule-multi-line`
image
image
image

artemkovalyov:
@minad, I addressed your suggestions. Take another look.

minad:
@oantolin Did you find time to look at this PR again?

embark-consult.el
```diff
@@ -103,42 +103,100 @@
...
+        (count 0)
+        prop)
+    (with-current-buffer buf
+      (insert (propertize "Exported line search results (file-backed buffers only):\n" 'wgrep-header t 'font-lock-face '(:weight bold)))
+      (save-excursion (let (last-buf
+                            filename
+                            (not-saved-buffers '()))
+                        (dolist (line lines)
+                          (pcase-let*
+                              ((`(,loc . ,num) (consult--get-location line))
```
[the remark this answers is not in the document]
artemkovalyov: @amosbird, I appreciate your feedback. I'm using this functionality frequently myself. It would be great meaningfully extend embark further if @oantolin can look into it again. I hope he's well and sound and we'll hear from him soon.

artemkovalyov:
Hey, @oantolin, do you think we can merge this? The FSF process is long finished and the feature seems to see a bit of traction. I keep using it frequently myself.

oantolin:
Sure, if you don't have time to fix the issues in my code review, I can do that for you, I just wanted to give you a chance to do that first, @artemkovalyov. Or did you fix them and just not mark them as done? I'll check.

oantolin:
Oh, boy, I think I had not submitted the review so that means that all these months you hadn't even seen it! I'm very sorry, it's the first time I ever do a code review (I'd always used the "add a single comment" feature before), and I did not know I had to submit it or where to click to submit.

artemkovalyov:
@oantolin, it happened to the best of us :) I'll take a look at your comments and suggestions. It's so sad I'm not coding LISP daily although using Emacs extensively.

minad:
Any news on this one?

oantolin:
Not on my end. I'm stil hoping @artemkovalyov fixes those issues I brought up.

minad:
I suspect that @artemkovalyov followed your example and forgot to click the "submit commits" button. ;)

oantolin:
Well, I didn't forget to submit: I didn't know you had too! And also didn't notice the submit button. 😬 I don't know if this is better or worse than forgetting to submit when you know you have to.

artemkovalyov:
Hey @oantolin & @minad, I resolved the conflicts and fixed minor issues that were pointed out by @oantolin.
I like your humor in the comments. The actual reason is my two three year old twins. They are making coding for fun possible only when they are asleep :)

artemkovalyov:
@oantolin, if you'll point me at the place where I can learn how to provide a quick menu to disambiguate between two location exporters, I might do another PR to add that convenience. Otherwise it's done via config and can be kept stupid simple.
`(setf  (alist-get 'consult-location  embark-exporters-alist) '(#'embark-consult-export-location-grep))`

artemkovalyov:
Tested for both single and multi-buffer case
image
image

artemkovalyov:
I'm feeling super dull, but I didn't manage to figure out how to make a function that would create a prompt based on a defined keymap. All those targets and indicators are getting me lost...

oantolin:
Thanks for fixing the things I pointed out! And congratulations on the twins, I can certainly understand that they make hobby coding a very low priority, since I went through the same thing (with one half of the number of babies).

>  I took a quick look at how to provide a choice of a desired exporting format for consult-location but failed to find a nice pattern. My idea was to have a list of exporter functions on mapping alist and prompt about which one to use. This seems to interfere with your design quite a bit and needs quite some refactoring.

I don't think that's a good idea or even necessary, I think most people will always want occur export or always want grep, so it's much better as a setting in the user's configuration. The code you gave is exactly what I had in mind:

```elisp
(setf  (alist-get 'consult-location  embark-exporters-alist) #'embark-consult-export-location-grep)
```

I'll test the new code later today and merge. Thanks again!

oantolin:
There is a *whole lot* of duplication between `embark-consult-export-grep` and `embark-consult-export-location-grep`. I will refactor as we suggested above. (I'll fix a couple of bugs I spotted in the new function along the way: one is that it doesn't correctly gather the list of non-file-backed buffers, another that it doesn't handle newer versions of wgrep which changed a variable name.)

oantolin:
Another small issue: the new function does not count the matches so it does not set the `grep-num-matches-found` variable. I don't know if that is actually important or not, but we might as well set it to the correct value.

oantolin:
I like that the new function prints the file names invisibly and has a header for each file that is visible but marked as ignored by wgrep.

oantolin:
OK, I'm done with my changes. It looks ready to merge on my end. I'll maybe test a little bit more, then merge.

artemkovalyov:
Thank you for the collaboration, @oantolin, and for bothering to create embark :)

PR #561, opened 2022-11-27
Auto load Org integration

minad:
I propose to make embark-org available automatically. The situation is the same with embark-consult. This way we make sure that the actions are available automatically.

We should probably wait for https://github.com/melpa/melpa/pull/8294, such that we don't trigger errors.

### DIFF
--- a/embark-org.el
+++ b/embark-org.el
@@ -31,6 +31,7 @@
 
 (require 'embark)
 (require 'org)
+(require 'org-element)
 
 ;;; Basic target finder for Org
 
@@ -98,8 +99,6 @@
     ;; verse-block
     )
   "Supported Org object and element types.")
-
-(declare-function org-element-property "org-element" (property element))
 
 (defun embark-org-target-element-context ()
   "Target the smallest Org element or object around point."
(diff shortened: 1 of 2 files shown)


oantolin:
I was already planning to do this once the MELPA PR is merged, but thanks for making it a single button-push for me!

minad:
> I was already planning to do this once the MELPA PR is merged, but
> thanks for making it a single button-push for me!

Haha, okay. I was too impatient. On the other hand, we all waited for a
long time to finally get official access to embark-org. ;)

embark-org.el
```diff
@@ -289,10 +288,8 @@ what part or in what format the link is copied."
 (embark-org-define-link-copier description description "'s description")
 (embark-org-define-link-copier target target "'s target")
 
-(declare-function embark-org-copy-link-inner-target "embark-org")
-(fset 'embark-org-copy-link-inner-target 'kill-new)
-(put 'embark-org-copy-link-inner-target 'function-documentation
-      "Copy 'inner part' of the Org link at point's target.
+(defalias 'embark-org-copy-link-inner-target #'kill-new
+  "Copy 'inner part' of the Org link at point's target.
```
oantolin, line 292: Ha! I didn't know `defalias`  took a docstring too! I would have used it had I known.

oantolin:
Thanks!

PR #564, opened 2022-12-01
embark-export-apropos: Ensure that symbol order is preserved

minad:
It would be great if we can improve the apropos export a little bit. I would like to deprecate the `consult-apropos` command which has always been somewhat ill-defined. See https://github.com/minad/consult/commit/eacc143a4439a394b5da2124e2550816c66c6d41.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -3165,9 +3165,11 @@ PRED is a predicate function used to filter the items."
   "Create apropos buffer listing SYMBOLS."
   (embark--export-rename "*Apropos*" "Apropos"
     (apropos-parse-pattern "") ;; Initialize apropos pattern
-    (apropos-symbols-internal
-     (delq nil (mapcar #'intern-soft symbols))
-     (bound-and-true-p apropos-do-all))))
+    ;; HACK: Ensure that order of exported symbols is kept.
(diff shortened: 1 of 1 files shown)


oantolin:
I guess it would be nice if instead of `apropos-sort-by-scores`, there were a user option where you can specify the comparison function. But with things as they stand, I guess this is basically the only way to control the sort.

minad:
I think it is best (and simplest) to just preserve the order as is since the order comes from completion. In the completion UIs there are usually flexible mechanisms to control the sorting, vertico-sort-(override-)function, completion-sort, etc.

oantolin:
I agree that it is best to keep the order when exporting. I'm not sure, but I have the impression that maybe my previous remark was misunderstood. You described this dynamic binding of `sort` as a "hack", and I was trying to say that I don't there is any better way of achieving the result.

minad:
On 12/2/22 14:41, Omar Antolín Camarena wrote:
> I agree that it is best to keep the order when exporting. I'm not sure,
> but I have the impression that maybe my previous remark was
> misunderstood. You described this dynamic binding of |sort| as a "hack",
> and I was trying to say that I don't there is any better way of
> achieving the result.

Okay, then I've misunderstood.

PR #566, opened 2022-12-02
Fix #565

### LINKED ISSUE #565: embark-collect fails for consult-line-multi
Due to the group headers get a nil error since `(car entry)` in `embark-consult--upgrade-markers` can be nil.

~~~ elisp
Debugger entered--Lisp error: (args-out-of-range 0 0)
  (get-text-property 0 consult-location nil)
  (let ((loc (get-text-property 0 'consult-location cand))) (if (consp (car loc)) (progn (setcar loc (set-marker (make-marker) (cdr (car loc)) (car (car loc)))))) loc)
  (consult--get-location nil)
  ...
  (embark-consult--upgrade-markers)
~~~

~~~ elisp
(defun embark-consult--upgrade-markers ()
  "Upgrade consult-location cheap markers to real markers.
This function is meant to be added to `embark-collect-mode-hook'."
  (when (eq embark--type 'consult-location)
    (mapc (lambda (entry) (consult--get-location (car entry))) ;; Should be guarded
          tabulated-list-entries)))
~~~

minad:


oantolin:
Thanks for doing this! I guess no consult commands that produced `consult-location` targets ever had group headers before.

minad:
On 12/2/22 14:42, Omar Antolín Camarena wrote:
> Thanks for doing this! I guess no consult commands that produced
> |consult-location| targets ever had group headers before.

consult-line-multi always did. But I guess it wasn't used often. The new
version is more powerful.

oantolin:
`consult-line-multi` is not a new command!? 🤣 I've never used it before this reimplementation and didn't know it existed.

oantolin:
Now that I read the docstring I remember the command. I still think I've never used it before today.

PR #568, opened 2022-12-10
Add match highlight and count to grep export

astoff:
What do you think? The match counting part is a bit problematic (and we could get rid of it) because:
 1. Consult currently only highlights the first match in a line, so the counting, effectively, is of matched lines and not actual matches as in `rgrep` etc.
 2. Even if all matches were highlighted, we wouldn't be able to count adjacent matches correctly (say "abc" matches "abcabc" twice), because we are actually counting the number of contiguous regions of text that match.

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -155,6 +155,8 @@ This function is meant to be added to `embark-collect-mode-hook'."
 
 ;;; Support for consult-grep
 
+(defvar grep-mode-line-matches)
+(defvar grep-num-matches-found)
 (defvar wgrep-header/footer-parser)
 (declare-function wgrep-setup "ext:wgrep")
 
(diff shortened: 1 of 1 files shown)


minad:
Great, the highlights are a good idea! But I would change the match count. The match count should simply be the number of exported lines, since grep is line based, consult-grep is line-based and the Vertico indicator shows the number of candidates/lines.

astoff:
> The match count should simply be the number of exported lines

This is not what `M-x rgrep` does, though. Grep is line-based, but if you ask it to highlight matches it highlights all matches in each line.

By the way, I think it would be useful to highlight all matches in a line in `consult--highlight-regexps`.

minad:
On 12/11/22 09:43, Augusto Stoffel wrote:
>     The match count should simply be the number of exported lines
> 
> This is not what |M-x rgrep| does, though. Grep is line-based, but if
> you ask it to highlight matches it highlights all matches in each line.

Okay, we can leave this then as is. But is notion of a match count well
defined given the abcabc problem you mentioned before? What about
matching "aaa" against "a.*"? Is this one match or three matches?

> By the way, I think it would be useful to highlight all matches in a
> line in |consult--highlight-regexps|.

Sure, we can add this if you send me a patch. :)

astoff:
> But is notion of a match count well defined given the abcabc problem you mentioned before?

Grep is correct in this regard (as is the Emacs rgrep):
```
$ echo abaa | grep -o -e a
a
a
a
```

> What about matching "aaa" against "a.*"?

That would be one match, since a regexp is supposed to find the longest match. Three matches for "a.*?".

oantolin:
Great, thanks!

PR #575, opened 2022-12-25
BROKEN -- Add embark--unfold -- BROKEN

### LINKED ISSUE #571: Wish: respect `isearch-open-invisible`on n/p
My favorite way to find symbols is to invoke `embark-act` followed by `embark-next|previous-symbol`.  It would be lovely if these commands respected `isearch-open-invisible` in the same manner isearch does: opening and re-hiding invisible text blocks (ala `outline-mode`) using the function named in this text property.  Could possibly just re-use isearch's capabilities here since the two shouldn't be possible to run at the same time.

minad:
Do not merge!

Add an embark--unfold post action to open closed overlays. Unfortunately it doesn't work yet as expected in Org files. I don't know if this is the fault of Embark or of the new Org fold API. Unfortunately the new Org fold API is problematic. While the idea isn't bad, I only had issues with it.

Fix #571

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -416,7 +416,9 @@ arguments and more details."
     (embark-rename-buffer embark--restart)
     (,'tab-bar-rename-tab-by-name embark--restart) ;; Avoid package-lint warning
     (,'tab-bar-close-tab-by-name embark--restart)
-    (package-delete embark--restart))
+    (package-delete embark--restart)
+    (embark-next-symbol embark--unfold)
+    (embark-previous-symbol embark--unfold))
   "Alist associating commands with post-action hooks.
 The hooks are run after an embarked upon action concludes.  See
 `embark-target-injection-hooks' for information about the hook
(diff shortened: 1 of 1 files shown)


jdtsmith:
Any thoughts on adapting and merging something like this?  I find `reveal-mode` works well for things like xref, and of course isearch already reveals by default.   So it's really embark that is the worst offender, leaving me inside the `...` (`n/`s` on symbols is the typical case).  Between outli and org, I tend to fold documents aggressively these days.

oantolin:
I'd like to have something like this in Embark, but don't currently know enough to implement it myself. I'd definitely welcome help with the implementation, or patience while i find the time to read about this stuff.

oantolin:
But preferably help. 😉

minad:
@jdtsmith You can take a look at `consult--invisible-open-permanently` and add this to Embark. In this attempt I tried exactly this. The imported code could be outdated by now and for some reason it did not work.

jdtsmith:
I've found `reveal-mode` actually works quite well with embark via its post-command-hook.  The only problem is that the PCH is run only when  embark-act finally completes, so not as useful for actions which don't return immediately (i.e. most, for me).  My crude workaround so far is:

```elisp
(let ((rpc (lambda (&rest _)
             (if (bound-and-true-p reveal-mode) (reveal-post-command)))))
  (dolist (cmd '(embark-previous-symbol embark-next-symbol))
    (cl-pushnew rpc (alist-get cmd embark-post-action-hooks))))
```

@minad did you investigate and rule out the use of `reveal-mode` in consult for some reason?

minad:
On 5/3/23 22:26, JD Smith wrote:
> @minad <https://github.com/minad> did you investigate and rule out the
> use of |reveal-mode| in consult for some reason?

Consult use the same approach which is also used by Isearch and
reveal-mode to unfold hidden overlays
(consult--invisible-open-permanently). I consider reveal-mode as too
intrusive and I wouldn't want it enabled by Embark by default. If you
can make it work, that's great, but probably belongs to the wiki. But I
am pretty sure we can fix the lower-level approach used in
consult--invisible-open-permanently and integrate it into Embark as I
attempted in this PR.

jdtsmith:
Right.  If you want xref and others to work, you still need something like reveal-mode though (or for each tool to be patched).  BTW, I only now noticed `consult-line`'s position preview does not unfold in org, though it works in other types of folded buffers.

minad:
> BTW, I only now noticed consult-line's position preview does not unfold in org, though it works in other types of folded buffers.

That's an issue of your configuration then. consult-line unfolding works in Org buffers.

jdtsmith:
Strange, it was intermittently not unfolding and has now started working again (same Emacs session, same document).  It actually seemed to unfold and refold in quick succession on preview, causing a bit of flashing.  I wonder if overlay properties reveal-mode leaves behind are interfering (I had disabled it).  I'll keep an eye out and open an issue if I uncover anything.

PR #591, opened 2023-01-30
Switch to the new key convention

minad:
This change affects:

- embark-help-key
- embark-cycle-key
- embark-keymap-prompter-key

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -121,7 +121,11 @@
 
 (defgroup embark nil
   "Emacs Mini-Buffer Actions Rooted in Keymaps."
-  :group 'minibuffer)
+  :link '(info-link :tag "Info Manual" "(embark)")
+  :link '(url-link :tag "Homepage" "https://github.com/oantolin/embark")
+  :link '(emacs-library-link :tag "Library Source" "embark.el")
+  :group 'minibuffer
+  :prefix "embark-")
 
 (defcustom embark-keymap-alist
   '((file embark-file-map)
--- a/embark.el
+++ b/embark.el
@@ -227,27 +227,27 @@ prompts for an action with completion."
 (defcustom embark-keymap-prompter-key "@"
   "Key to switch to the keymap prompter from `embark-completing-read-prompter'.
 
-The key must be either a string or a vector.
-This is the key representation accepted by `define-key'."
-  :type '(choice key-sequence (const :tag "None" nil)))
+The key must be either nil or a string.  The
+string must be accepted by `key-valid-p'."
+  :type '(choice string (const :tag "None" nil)))
 
 (defcustom embark-cycle-key nil
   "Key used for `embark-cycle'.
 
 If the key is set to nil it defaults to the global binding of
-`embark-act'.  The key must be either a string or a vector.  This
-is the key representation accepted by `define-key'."
-  :type '(choice key-sequence (const :tag "Use embark-act key" nil)))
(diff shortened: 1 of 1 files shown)


oantolin:
I like the links in the customize buffer and of course, fixing my broken call to `make-obsolete`, but isn't switching from `define-key` to `keymap-set` just a loss of functionality? In particular I don't want to change the custom type from `key-sequence` to `string`.

oantolin:
I merged the last two commits, but am waiting a bit on switching to the new key convention because it doesn't seem to have any advantages to me. If for example you told me that the plan is to eventually deprecate `define-key`, then yes, I guess we should switch.

oantolin:
And thanks again for the fix to the make-obsolete business, I feel really bad about not testing that in fresh Emacs session where `embark-define-keymap` wasn't already defined.

minad:
> I merged the last two commits, but am waiting a bit on switching to the new key convention because it doesn't seem to have any advantages to me.

The advantage is that the user can use kbd strings for configuration, which are clearly more user friendly and uniform than the vector or string syntax

define-key won't get deprecated since it is used internally, but from what I've seen they don't mention it anymore in the manuals.

oantolin:
That soft deprecation is reason enough to switch for me, I guess. The custom type is a little unfortunate, but not too bad.

oantolin:
> The advantage is that the user can use kbd strings for configuration

They already could, by using `kbd`. But it is nice to not have to type `(kbd )`. If Emacs is gradually moving in the direction of kbd strings by default that is an improvement.

minad:
> The custom type is a little unfortunate, but not too bad.

Indeed. I guess one could specify a validation function. Maybe it is also possible to define a custom widget type or maybe Emacs already comes with such a type. I don't know.

> They already could, by using kbd. 

Yes, but that's quite ugly and I remember that key specifications were a significant source of confusion or me with Emacs for a long time.

Btw, I also moved to the new convention for the Consult key customization variables.

embark.el
```diff
@@ -949,7 +949,11 @@ their own target finder.  See for example
 
 (defun embark--cycle-key ()
   "Return the key to use for `embark-cycle'."
-  (or embark-cycle-key (car (where-is-internal #'embark-act))))
+  (if embark-cycle-key
+      (if (key-valid-p embark-cycle-key)
+          (key-parse embark-cycle-key)
+        (error "`embark-cycle-key' is invalid"))
+    (car (where-is-internal #'embark-act))))
```
oantolin, line 960: I think this is wrong: you later call `keymap-set` with this key and `where-is-internal` does not return strings that are `key-valid-p`, which is what `keymap-set` takes.

minad:
On 2/1/23 19:19, Omar Antolín Camarena wrote:
> I think this is wrong: you later call |keymap-set| with this key and
> |where-is-internal| does not return strings that are |key-valid-p|,
> which is what |keymap-set| takes.

No, I don't think I use keymap-set with embark--cycle-key but define-key?

oantolin:
That's right you do use `define-key`, sorry!

oantolin:
Yes. The only really compelling argument not to merge is the custom-type and I can live with that slight loss of functionality.

oantolin:
But I did also take the advantage to learn how to merge some but not all the commits in a PR!

oantolin:
Fully merged now.

PR #593, opened 2023-02-02
Do not install a global advice on tabulated-list-revert

minad:
Instead define our own revert function for embark-collect-mode buffers. It is good practice to avoid advices, and global advices even more so.

(I did some advice cleaning in my Vertico and Corfu packages and then I also found this advice in Embark.)

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -2528,14 +2528,6 @@ This is only used for annotation that are not already fontified.")
 (defvar-local embark--rerun-function nil
   "Function to rerun the collect or export that made the current buffer.")
 
-(defun embark-collect--post-revert (&rest _)
-  "Run `embark-collect-post-revert-hook'.
-This function is used as :after advice for `tabulated-list-revert'."
-  (when (derived-mode-p 'embark-collect-mode)
-    (run-hooks 'embark-collect-post-revert-hook)))
-
-(advice-add 'tabulated-list-revert :after #'embark-collect--post-revert)
-
 (autoload 'package-delete "package")
 (declare-function package--from-builtin "package")
 (declare-function package-desc-extras "package")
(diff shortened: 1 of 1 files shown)


oantolin:
That does look a little better, thanks!

oantolin:
You know, that entire hook and revert function could probably be removed now. I doubt anyone still uses them. It was useful back in the dark days when embark also had a completions UI.

minad:
On 2/3/23 00:28, Omar Antolín Camarena wrote:
> You know, that entire hook and revert function could probably be removed
> now. I doubt anyone still uses them. It was useful back in the dark days
> when embark also had a completions UI.

Okay, please do that then. All simplifications are welcome in my eyes!
When I applied the change I didn't dig deeper, it was just a mechanic
change.

oantolin:
In fact, if I recall, correctly I added it so Prot could have an easy way of resize the embark completions window to the minimum height needed for the candidates. :)

minad:
On 2/3/23 00:37, Omar Antolín Camarena wrote:
> In fact, if I recall, correctly I added it so Prot could have an easy
> way of resize the embark completions window to the minimum height needed
> for the candidates. :)

There seems to be still a usage in the embark-collect-zebra-minor-mode.
But there you could use as well the tabulated-list-revert-hook? Why do
we need an extra hook?

oantolin:
It's an extra hook so that it is not shared with non-embark tabulated lists, i.e., that way users can set it to a hook run whenever you revert an embark collect buffer but not when you revert some other tabulated list. For the purposes of the zebra-mode that is not important because the hook it adds is added locally any way.

minad:
Sure, I understand. But it is still almost useless since you can add a hook to `embark-collect-mode-hook` which then installs a local `tabulated-list-revert-hook` hook. I believe we can safely remove it. If there were many users, which would then avoid the aforementioned two hook indirection, it would be justified to keep it, but that does not seem to be the case?

oantolin:
Yeah, I should remove it. And also fix zebra-mode! It uses the revert hook, because embark-live _used to_ use revert, but since the new lighter embark-live came about it doesn't use reverting anymore. This means that zebra-mode doesn't actually update in embark-live buffers.

minad:
Btw, zebra mode should be rewritten anyway to not use overlays, which are too expensive for this use case. It would be better to use text properties.

oantolin:
It is still the only sensible way to browse the kill-ring. Not that I actually browse the kill-ring that much.

oantolin:
I personally wouldn't miss zebra mode if it were gone.

oantolin:
I can live with 0 sensible ways to browse the kill-ring, I guess. 🤣

oantolin:
I'll carefully clean these things up in the next few days.

minad:
And then release immediately! :)

Issue #164, opened 2021-02-23
Zebra mode overlay takes precendence over candidate background


minad:
In `consult-mark` I am using a background face to visualize the point position. The background is overwritten by the zebra overlay since overlays win over text properties. Any ideas how we can work around this? `consult-register-window` has the same problem, since I am also using overlays there.

minad:
What about using thin lines? I think I had proposed that before, but I don't recall if there was a reason to prefer the zebra.

~~~ elisp
(setq-local inhibit-point-motion-hooks nil)
(insert (concat
         "line1"
         (propertize (concat "\n" (propertize " " 'display '(space :align-to right)) "\n")
         'face '(:height 0.01 :underline t :foreground "#ccc")
         'cursor-intangible t 'intangible t)
         "line2"
         (propertize (concat "\n" (propertize " " 'display '(space :align-to right)) "\n")
         'face '(:height 0.01 :underline t :foreground "#ccc")
         'cursor-intangible t 'intangible t)
         "line3"))
~~~

oantolin:
I guess one reason is that the "thin" lines still take up a normal height line in the terminal version of Emacs.

oantolin:
Also, I think it was I that proposed the lines, for `consult-yank`, and you vetoed it. 😉

minad:
And I thought I voted for lines. The snippet above creates lines of height zero by scaling the face. This works only in graphical mode obviously. I thought about removing zebra mode from the consult register view and replace it by that. Or offer it as an alternative to zebra. But since the zebras hide the marks it is really not good for registers.

oantolin:
Another option is to reimplement zebra stripes with text properties.

oantolin:
My old completing-read yank function that I used before Consult added lines to the candidates.

minad:
Do you prefer lines or zebras? I think about removing zebras in favor of lines. Alternatively provide both variants as options.

oantolin:
Well, it does bother me a bit that lines take up an entire row in `emacs -nw`. (Well, I'm assuming, I haven't tested, is that really what happens?) But that's me just looking out for terminal Emacs users, I personally never use `emacs -nw`.

That's the only thing, except for that, I don't have any reason to prefer lines over zebra stripes or viceversa.

oantolin:
[...]

PR #596, opened 2023-02-10
Unzebra

minad:
Just opening this such that the discussion in #593 does not get lost. Remove the zebra and the (almost useless) embark-collect-revert-hook. Given how much code space the Zebra needs, I don't think it is justified to keep it in captivity in our zoo. As alternative I suggest a meerkat mode. Also "fixes" #164 .

### DIFF
--- a/README.org
+++ b/README.org
@@ -1024,13 +1024,7 @@ default.
 
 Besides those exporters and candidate collectors, the =embark-consult=
 package provides many subtle tweaks and small integrations between
-Embark and Consult. For example, if you run =embark-collect= from any of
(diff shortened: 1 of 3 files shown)


minad:
No sympathy for meerkats? :cat2:

oantolin:
I wonder if an outright removal is best. Maybe I should go with obsoletion warnings first.

minad:
Obsoletions are generally more user friendly, however I always wonder about their visibility. If you enable obsolete zebra you won't notice anything in most cases.

oantolin:
I decided to bite the bullet and just remove it. If I had to bet, I'd bet nobody even notices.

PR #600, opened 2023-02-14
Add embark-eldoc-function (Fix #599)

### LINKED ISSUE #599: Embark Eldoc?
https://www.masteringemacs.org/article/seamlessly-merge-multiple-documentation-sources-eldoc

We could show the first target at point.

minad:
I think it is worth adding this. It is such a cheap addition given that Eldoc handles everything, timers etc. Needs this configuration:

~~~ elisp
(add-hook 'eldoc-documentation-functions #'embark-eldoc-function)
(setq eldoc-documentation-strategy #'eldoc-documentation-compose-eagerly)
~~~

I haven't added documentation to the README.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -992,6 +992,15 @@ If CYCLE is non-nil bind `embark-cycle'."
       (concat (car (split-string target "\n" 'omit-nulls "\\s-*")) "…")
     target))
 
+(defun embark-eldoc-function (report &rest _)
+  "Eldoc function reporting the target at point via the REPORT callback.
+This function can be added to `eldoc-documentation-functions'."
+  (when-let ((target (car (embark--targets))))
+    (funcall report
+             (format "Embark on %s ‘%s’"
+                     (plist-get target :type)
+                     (embark--truncate-target (plist-get target :target))))))
+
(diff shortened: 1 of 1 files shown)


minad:
Thoughts about this and #596?

oantolin:
So using this for a while confirmed my intial impression: I always know what the first target is! So for me personally that eldoc function isn't useful. I recognize I'm probably hyperaware of Embark and that other people might find your function useful, but for me, I renamed it more informatively and added a second function _I_ find useful: one that reports the types of _all_ targets at point. The great thing about knowing the order of the targets at point is that you can use `embark-dwim` with a numeric prefix to select the default action for each type.

oantolin:
Thanks a lot for doing this. It would probably never have occurred to me.

oantolin:
I think that the default eldoc strategy seems a better fit for these embark functions: they notify you of embark stuff only if there isn't actual documentation to show you.

minad:
Thanks! Very good. I think the first function is mainly useful for beginners, the second more for experts. I will try the first one for a while, also with the merge strategy.

PR #603, opened 2023-02-18
Drop Selectrum support

minad:
Selectrum has been deprecated in favor of Vertico. Selectrum support has already been dropped from Marginalia and Consult.

We could keep support for longer, since the additional code doesn't hurt. But I think it is better to nudge the community to move over to completion UIs which are more compliant with the Emacs completion mechanism (Vertico, Icomplete, etc.).

### DIFF
--- a/README.org
+++ b/README.org
@@ -223,7 +223,7 @@ Finally, there is also an =embark-live= variant of the =embark-collect=
 command which automatically updates the collection after each change
 in the source buffer. Users of a completion UI that automatically
 updates and displays the candidate list (such as Vertico, Icomplete,
-Selectrum, Fido-mode, or MCT) will probably not want to use
+Fido-mode, or MCT) will probably not want to use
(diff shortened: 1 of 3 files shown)


oantolin:
Sure, let's remove it. People that insist on using Selectrum can add this code to their configurations.

oantolin:
So no deprecation warning, just straight up removal? I'm in favor of just removing it, but I want to make sure you think that's OK too.

minad:
Yes, I think that's okay by now, since most users already have migrated. I had some deprecation warning in Consult, but I have also removed this now. I suspect that the only users left using Selectrum are the ones who use Selectrum only or Selectrum+Prescient. These users can be notified if we add some deprecation message to Selectrum itself, but that's not relevant here.

oantolin:
OK, let's just remove it, then. Thanks for the pull request!

PR #604, opened 2023-02-18
Fix wgrep read-only header (Fix #602)

### LINKED ISSUE #602: Embark grep export - Wgrep is broken
I observed that the first two matching lines in the grep export are not editable with wgrep (error: Text is read-only). I think this breakage is recent, maybe caused by changes to wgrep itself. It seems to work in normal `rgrep` buffers, but not when using the Embark export.

minad:


oantolin:
Do you think it's fine to only update the variable name or should embark work with both variable names to support both the current and older version of wgrep? I'm inclined to only support the new version (as done in this PR).

minad:
I think supporting only the new name is sufficient.

yalfg:
@oantolin : the version of wgrep in NonGNU ELPA and MELPA stable are still 3.0.0 and using the old name (with "/"), and because of this they break with a recent embark. Ideally a new stable version of wgrep would be the best solution but it's seems there not much activity there (there's already an issue asking for this).

In the meantime I leave this note here, and maybe a quick note in the embark readme when wgrep is mentioned could help newcomers who may stick to the stable repos?

Thanks

minad:
@yalfg Thanks, this should be fixed here in the meantime. See #686.

PR #605, opened 2023-02-18
Autoload eldoc functions

minad:
Follow-up of #600 plus some minor improvements


minad:
I missed that we should autoload the eldoc functions such that they can be registered before Embark is loaded.

PR #606, opened 2023-02-19
Disable Embark Eldoc in minibuffer

minad:
In the minibuffer the Embark target is obvious. The problem is that Eldoc abuses the modeline to display the Eldoc information at point, which is noisy and unnecessary in this case.

Follow-up of #600 and #605

### DIFF
--- a/README.org
+++ b/README.org
@@ -306,6 +306,10 @@ starting configuration:
     ;; Optionally replace the key help with a completing-read interface
     (setq prefix-help-command #'embark-prefix-help-command)
 
+    ;; Show the Embark target at point via Eldoc
+    (setq eldoc-documentation-strategy #'eldoc-documentation-compose-eagerly)
+    (add-hook 'eldoc-documentation-functions #'embark-eldoc-first-target)
+
     :config
 
     ;; Hide the mode line of the Embark live/completions buffers
--- a/README.org
+++ b/README.org
@@ -306,9 +306,10 @@ starting configuration:
     ;; Optionally replace the key help with a completing-read interface
     (setq prefix-help-command #'embark-prefix-help-command)
 
-    ;; Show the Embark target at point via Eldoc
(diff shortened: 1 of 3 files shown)


minad:
There are probably ways to configure Eldoc to not abuse the mode line, but then the info may pop up somewhere else. I think it is just easier to check if we are not in the minibuffer in the embark-eldoc-* functions.

minad:
I added another commit which updates the example configuration. You may also want to document the Eldoc functionality somewhere. I am not sure about the best place.

oantolin:
Ha! You can tell I didn't just leave these functions on for a while, but rather did some focused testing, moving around various buffers and the turned them off.

minad:
I will keep them on in normal buffers. There I like them, since I am already accustomed to Eldoc information popping up for function arguments. Also I use a large value for the delay. In the minibuffer the information bothers me, I think it is the combination of useless information plus mode-line abuse.

minad:
The situation is comparable to Marginalia. I think it is good to have more information ready, since it makes Emacs more accessible. If you are experienced you may want to turn off the noise again. But then you can also just selectively not perceive the noise, like Lisp parentheses. ;)

oantolin:
I hesitate to recommend the eager composition eldoc strategy. Maybe we should just mention the eldoc embark functions and tell people to check out the eldoc strategies?

minad:
> Maybe we should just mention the eldoc embark functions and tell people to check out the eldoc strategies?

Sure, sounds good. To explain why I use the eager composition strategy - multiple eldocs are not uncommon. There can be variable and function argument eldocs at the same time, there can be flymake errors and now Embark. It may get noisy but I don't see why one is necessarily preferred over another.

minad:
@oantolin Is this okay to merge for now?

oantolin:
Perfect, thanks!

minad:
Thanks!

PR #610, opened 2023-03-08
Docs: (README.org) Minor fix in code example

alphapapa:
The wrong variable name was used.


oantolin:
Thanks for catching that!

PR #615, opened 2023-03-22
Make `embark-export` abide to `display-buffer-alist`

mpereira:
This change makes it so that `buffer` is named before it is
`pop-to-buffer`'d to, so that the rules specified `display-buffer-alist`
can be matched by the Embark export buffer name patterns.

E.g., the following addition makes Embark export buffers show in a
side window.
```elisp
(add-to-list 'display-buffer-alist '("\\*Embark Export: .*"
                                     (display-buffer-in-side-window)))
```

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -3118,17 +3118,17 @@ buffer."
                         (alist-get t embark-exporters-alist))))
       (if (eq exporter 'embark-collect)
           (embark-collect)
-        (let ((after embark-after-export-hook)
-              (cmd embark--command)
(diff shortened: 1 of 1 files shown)


oantolin:
Thanks for this! I hadn't even noticed this caused problems with `display-buffer-alist`.

mpereira:
> Have you signed the FSF copyright assignment, @mpereira?

Not yet. I'll do it if I come up with more PRs.

> Thanks for this!

You're welcome. Thank _you_ for your packages.

PR #626, opened 2023-04-26
Add Flymake diagnostic target

minad:


oantolin:
Good idea.

oantolin:
This reminds me that someone made a target finder for that mode for merging conflicts in version controlled files, smerge-mode, I think it's called (I use it all the time, I'm just not sure of the name). I've been using repeat-mode for that, but think that maybe Embark should just ship with that. The mode's keymap is already good to go as the action keymap.

oantolin:
Just for completeness here's what I use in my configuration right now:

```elisp
(map-keymap
  (lambda (_key cmd)
    (when (symbolp cmd)
      (put cmd 'repeat-map 'smerge-basic-map)))
  smerge-basic-map)
```

It would be nice to use `C-.` instead of smerge-mode's ugly `C-c ^` (which I could of course configure), but I think I do prefer one feature of this repeat-mode approach: I made all commands repeat. Whenever I choose a version to keep I always want to go the next conflict.

If I do add the target finder to Embark, I'd make more of the actions repeatable.

oantolin:
Anyway, thanks for this PR!

minad:
Thanks! I am glad you like the repeat feature, which you initially resisted iirc. ;)

minad:
Embark is really nice to cheaply create such simple "transients". Similar to Flymake I can use this for Jinx:

~~~ elisp
(defvar-keymap embark-jinx-map
  :doc "Keymap for Embark actions on misspelled words."
  :parent embark-general-map
  "RET" #'jinx-correct
  "$" #'jinx-correct
  "n" #'jinx-next
  "p" #'jinx-previous)

(defun embark-jinx-target ()
  "Target a misspelled at point."
  (when-let ((ov (seq-find (lambda (ov) (eq 'jinx (overlay-get ov 'category)))
                           (overlays-in (max (point-min) (1- (point))) (point)))))
    `(jinx ,(buffer-substring-no-properties
                (overlay-start ov) (overlay-end ov))
              ,(overlay-start ov) . ,(overlay-end ov))))

(add-to-list 'embark-target-finders #'embark-jinx-target)
(add-to-list 'embark-keymap-alist '(jinx embark-jinx-map))
(add-to-list 'embark-repeat-actions 'jinx-next)
(add-to-list 'embark-repeat-actions 'jinx-previous)
(add-to-list 'embark-target-injection-hooks '(jinx-correct embark--ignore-target))
~~~

oantolin:
I was thinking of adding a macro to define overlay-based target finders, similar to the macro for defining thingatpt-based target finders.

minad:
Yes, but there are not always exactly the same. For example the bug-reference finder returns a url. The jinx finder looks or overlays between 1-point and point.

oantolin:
Right, I'd have to build a little flexibility into the macro. Probably just leave a couple of slots for an expression. Like my [define-thing-marker macro](https://github.com/oantolin/emacs-config/blob/1eba050840eb72988b7a592248ce67af59578be4/my-lisp/text-extras.el#L86--L123).

minad:
Something like this could work:

~~~ elisp
(cl-defmacro embark-define-overlay-target (name cond &key string type)
  `(defun ,(intern (format "embark-target-%s-at-point" name)) ()
     ,(format "Target %s at point." name)
     (when-let ((ov (seq-find (lambda (ov) ,cond)
                              (overlays-in (max (point-min) (1- (point))) (point)))))
       (cons ',(or type name) (cons ,(or string '(buffer-substring-no-properties
                                                  (overlay-start ov) (overlay-end ov)))
                                    (cons (overlay-start ov) (overlay-end ov)))))))

(embark-define-overlay-target jinx (eq 'jinx (overlay-get ov 'category)))
(embark-define-overlay-target flymake (overlay-get ov 'flymake-diagnostic))
(embark-define-overlay-target bug-reference (overlay-get ov 'bug-reference-url)
                              :type url :string (overlay-get ov 'bug-reference-url))
~~~

Not really worth it if you ask me...

PR #627, opened 2023-04-27
Update docstring for `embark-bindings-at-point' in embark.el

sympodius:
Escape single quotes in docstring for `Stash'

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -1752,7 +1752,7 @@ text properties `keymap' or `local-map', from either buffer text
 or an overlay.  These are not widely used in Emacs, and when they
 are used can be somewhat hard to discover.  Examples of locations
 that have such a keymap are links and images in `eww' buffers,
-attachment links in `gnus' article buffers, and the 'Stash' line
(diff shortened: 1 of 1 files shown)


oantolin:
Why did you want to make this change?

minad:
Because the bytecode compiler started to complain about single quotes for no good reason.

oantolin:
Oh, I'm still safe from that new bytecompiler warning here in 28.2-land. :) I guess the semantically correct fix is to either drop the quotes (and lowercase stash) or use double quotes, then.  I'll drop the quotes. cf5f25a22f91d7ed33f582feec7a2e71d766fc61

minad:
I moved to Emacs 29 not too long ago. It works quite well since the branch has been mostly frozen. :)

sympodius:
> Because the bytecode compiler started to complain about single quotes for no good reason.

Yeah, this was the reason. And you are quite right, @oantolin ; my suggestion to use the backtick-single-quote combination would have been semantically incorrect. You and @minad continue to do great work :)

PR #628, opened 2023-04-28
Add embark-define-overlay-target macro

minad:
See my comment https://github.com/oantolin/embark/pull/626#issuecomment-1527702873.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -735,21 +735,6 @@ different priorities in `embark-target-finders'."
               (lib (ffap-el-mode name)))
     `(file ,lib . ,(bounds-of-thing-at-point 'filename))))
 
-(defun embark-target-bug-reference-at-point ()
-  "Target a bug reference at point."
-  (when-let ((ov (seq-find (lambda (ov) (overlay-get ov 'bug-reference-url))
-                           (overlays-at (point)))))
-    `(url ,(overlay-get ov 'bug-reference-url)
-          ,(overlay-start ov) . ,(overlay-end ov))))
-
-(defun embark-target-flymake-at-point ()
-  "Target a Flymake diagnostic at point."
-  (when-let ((ov (seq-find (lambda (ov) (overlay-get ov 'flymake-diagnostic))
-                           (overlays-at (point)))))
-    `(flymake ,(buffer-substring-no-properties
-                (overlay-start ov) (overlay-end ov))
-              ,(overlay-start ov) . ,(overlay-end ov))))
-
 (defun embark-target-package-at-point ()
   "Target the package on the current line in a packages buffer."
   (when (derived-mode-p 'package-menu-mode)
@@ -822,6 +807,33 @@ different priorities in `embark-target-finders'."
          (end (ignore-errors (scan-sexps start 1))))
       (unless (eq start (car (bounds-of-thing-at-point 'defun)))
       `(expression ,(buffer-substring start end) ,start . ,end)))))
+
+(defmacro embark-define-overlay-target (name prop &optional pred type target)
+  "Define a target finder for NAME based on overlays with property PROP.
+The overlay is let-bound to the variable NAME and the overlay
+property is bound to the variable PROP.  PRED is an optional
+predicate expression.  TYPE is an optional type and defaults to
+the symbol NAME.  TARGET is an expression evaluating to the
+target string and defaults to the buffer substring."
+  `(defun ,(intern (format "embark-target-%s-at-point" name)) ()
+     ,(format "Target %s at point." name)
+     (when-let ((,name (seq-find
+                        (lambda (,name)
+                          (when-let ((,prop (overlay-get ,name ',prop)))
+                            (ignore ,prop)
+                            ,(or pred t)))
+                        (overlays-in (max (point-min) (1- (point)))
+                                     (min (point-max) (1+ (point))))))
+                (,prop (overlay-get ,name ',prop)))
+       (ignore ,prop)
+       (cons ',(or type name)
+             (cons ,(or target `(buffer-substring-no-properties
+                                 (overlay-start ,name) (overlay-end ,name)))
+                   (cons (overlay-start ,name) (overlay-end ,name)))))))
+
+(embark-define-overlay-target flymake flymake-diagnostic)
+(embark-define-overlay-target bug-reference bug-reference-url nil
+                              url bug-reference-url)
 
 (defmacro embark-define-thingatpt-target (thing &rest modes)
   "Define a target finder for THING using the thingatpt library.


oantolin:
With the macro, this pull request one only have to make the change once. 😛

minad:
Yes, but for me the magic threshold is 3. :-P

minad:
Maybe we will find some more overlay targets in the future which would justify the macro.

oantolin:
Well, smerge conflicts would make 3.

minad:
Luckily this has not been smerged yet :sweat_smile:

minad:
I give in and admit that an overlay target macro is generally useful ;) I added an `embark-define-overlay-target` macro. Please give it a try.

oantolin:
The role of `%` should be mentioned in the docstring. Also, maybe it would be more user-friendly if the predicate already included an `overlay-get`, no? We'd need a bit of extra syntax if you want a fancier predicate. But maybe enabling something like this would be nice:

```elisp
(embark-define-overlay-target flymake flymake-diagnostic)
(embark-define-overlay-target bug-reference bug-reference-url url %%)
(embark-define-overlay-target jinx (category (eq 'jinx %%)))
```

That is, we keep `%` for the overlay, but change the PRED argument to a PROP argument that specifies which property to get from the overlay. The property gets bound to `%%` (or we can think of a better symbol); and if you don't want the test to be just `(overlay-get % prop)`, instead of PROP being a symbol, you can give a two element list or a symbol and an expression, like in the jinx example.

oantolin:
Something like:

```elisp
(defmacro embark-define-overlay-target (name prop &optional type target)
  "Define a target finder for NAME using overlays.
PROP should be either a symbol specifying the overlay property
name to search for, or a two element list of the form (PROPNAME
CONDITION) where CONDITION is an s-expression specifying a
condition the overlay and the property value must satisfy.  The
s-expression can use `%' to refer to the overlay and it can use
`%%' to refer to the property of the overlay named PROPNAME.  If
given, TYPE is the target type reported by the target finder
being defined; if TYPE is nil, NAME is used as the target type.
TARGET is an s-expression evaluated to obtain the target string,
it can also use `%' and `%%' and defaults to the buffer substring
covered by the overlay."
  (let ((propname (if (consp prop) (car prop) prop))
        (pred (if (consp prop) (cadr prop) '%%)))
    `(defun ,(intern (format "embark-target-%s-at-point" name)) ()
       ,(format "Target %s at point." name)
       (when-let ((% (seq-find
                      (lambda (%) (let ((%% (overlay-get % ',propname))) ,pred))
                      (overlays-in (max (point-min) (1- (point)))
                                   (min (point-max) (1+ (point))))))
                  (%% (overlay-get % ',propname)))
         (cons ',(or type name)
               (cons ,(or target '(buffer-substring-no-properties
                                   (overlay-start %) (overlay-end %)))
                     (cons (overlay-start %) (overlay-end %))))))))
```

I think something like this would be more comfortable for users, but I am sure about the specific syntax: the two-element list case of PROP and the names `%` and `%%`.

minad:
I have a better idea - bind NAME to the overlay and PROP to the property value. See https://github.com/oantolin/embark/pull/628/commits/7f8708e39a2d855d586a7bab78d90ba61bdeafbe.

oantolin:
I love it, thanks!

oantolin:
Do you think it's worth adding a MODE argument to the macro, so that the target finder first checks if the mode is on, and if not, it skips going through the overlays? All of our examples can only find those overlays if a specific minor mode is on. Whether it is worth doing this or not depends on how many overlays are likely to be present and how slow it is to go through them. My guess is that almost always you have 0 overlays at point (or right befoer it), so it's better not to complicate the code by checking the mode first. What do you think?

minad:
> My guess is that almost always you have 0 overlays at point (or right befoer it), so it's better not to complicate the code by checking the mode first. What do you think?

Yes, this is my expectation too. I think we don't need an additional mode check.

oantolin:
I found a funny bug: if the name and prop arguments are the same, the property shadows the overlay and then default value for the target errors, as do the bounds. :)

minad:
> I found a funny bug: if the name and prop arguments are the same, the property shadows the overlay and then default value for the target errors, as do the bounds. :)

Right, that's a problem for the smerge overlays. Probably it is easier to just use fixed names like `prop` and `ov`? It is not pretty to bind fixed names, but at least we avoid complications.

oantolin:
I was thinking that maybe if NAME and PROP are the same, binding PROP-1 instead of PROP. But maybe `%o` and `%p` are better.

oantolin:
I guess another option is to bind two different variables to the overlay, NAME and a secret symbol, that way the bounds and default value of TARGET can be correct even if NAME and PROP are the same. And we let users worry about shadowing if it affects them.

minad:
Yes, I just considered theses ideas too, a derived name like prop-1 and the additional bindings. I think going with fixed symbols is just easier. I would use P and O since these symbols are uncommon and stand out.

oantolin:
I think I prefer `%o` and `%p`, because they're like `format` specifiers. I'll make the change.

minad:
Good point. I had forgotten why I used % in the first place, likely because of format. I also have a macro in Vertico iirc which binds %. :)

oantolin:
OK, done. ae544d68c29af19495194ddb34f53f31d59cf034

hmelman:
> No, it is not because of the overhead or at least that would be a minor reason in this case here. The actual reason is that using plain expression leads to a more concise macro call.

That was my other guess (after seeing the smerge example).  So in my docstring, s/for performance reasons/for convenience/.

minad:
For convenience or conciseness.

hmelman:
Conciseness works too.  I'd find a function name passed to the macro easier to read, and I think having it passed parameters instead of binding `%o` and `%s` is more conventional, but that would necessitate another defun (since these functions don't exist from elsewhere).  So using an expression makes the macro call itself (perhaps) less concise but the overall code more concise.  Not having to create (and name) a defun is convenient to the programmer.

minad:
Yes, note that the whole point of the macro is to avoid having to write a custom target finder for overlays. This is only worth it if the macro can be used concisely. My initial implementation had the following syntax and this wasn't concise enough for @oantolin. :)

~~~ elisp
(embark-define-overlay-target jinx (eq 'jinx (overlay-get ov 'category)))
(embark-define-overlay-target flymake (overlay-get ov 'flymake-diagnostic))
(embark-define-overlay-target bug-reference (overlay-get ov 'bug-reference-url)
                              :type url :string (overlay-get ov 'bug-reference-url))
~~~

hmelman:
I will say I do like the explicitness of the keywords `:type` and `:string` particularly since they're optional args.  Removing the `overlay-get` from the call is a good thing.

minad:
I think keyword arguments are not really needed for the three optional arguments, since they would also reduce conciseness. The macro is also rather special and hardly part of the commonly used public Embark API. There is not much profit if we fine tune the API for user friendliness and convenience. A much more interesting debate would be if there are more such target finder macros we should define. Currently we have the macro for thingatpt and overlays. Are there other useful patterns? One idea could be a regexp-based finder. One can maybe utilize `thing-at-point-looking-at` for that purpose?

minad:
@oantolin Would you like to add such a macro?

~~~ elisp
(defmacro embark-define-regexp-target (name regexp &optional type target)
  `(defun ,(intern (format "embark-target-%s-at-point" name)) ()
     ,(format "Target %s at point." name)
     (save-match-data
       (when (thing-at-point-looking-at ,regexp)
         (cons ',(or type name)
               (cons ,(or target '(match-string 0))
                     (cons (match-beginning 0) (match-end 0))))))))

(embark-define-regexp-target
 wikipedia "wikipedia:\\([[:alnum:]_]+\\)"
 url (format "https://en.wikipedia.org/wiki/%s" (match-string 1)))
~~~

oantolin:
There _was_ some problem with `thing-at-point-looking-at`, but unfortunately I don't remember what the problem was. I do remember thinking that `org-in-regexp` worked better. Maybe comparing those two I can quickly figure out what the issue was.

oantolin:
Well, one difference between those two functions is that  `thing-at-point-looking-at` will search the entire buffer by default, whereas `org-in-regexp` is limited by default to the current line. In both cases you can change this default by passing an optional parameter: for `thing-at-point-looking-at` the optional `distance` parameter is a number of characters before or after point to search; for `org-in-regexp` it is a number of lines.

Maybe that was my only problem? That line-based limits feel more natural for new target syntaxes? Certainly the default of the current line seems like a better default for a target finder than the entire buffer (having several target finders search the entire buffer could potentially slowdown `embark-act` noticeably).

For defining a new macro however, we can ensure we always pass a limit to `thing-at-point-looking-at` so this particular problem is moot. I'm not sure what limit I'd pass by default, maybe `(max (- (line-end-position) (point)) (- (point) (line-beginning-position)))` which is the smallest radius containing the current line.

hmelman:
I haven't looked at the new treesitter stuff, but maybe something from there?

oantolin:
How's this for the docstring of `embark-define-overlay-target`?

Define a target finder for NAME that targets overlays with property PROP.
The function defined is named embark-target-NAME-at-point and it
returns Embark targets based on the overlays around point. An overlay
provides a target if it has a non-nil property named PROP.

If PRED is non-nil, it should be an expression and it further
restricts the targets to only those overlays for which PRED evaluates
to a non-nil value.

The target finder returns target type NAME or optional symbol TYPE if
it is non-nil.

The target finder returns the substring of the buffer covered by the
overlay as the target string or the result of evaluating the optional
expression TARGET if it is non-nil.

PRED and TARGET are expressions (not functions) and when evaluated the
symbols `%o' and `%p' are bound to the overlay and the overlay's
property respectively.

oantolin:
> I haven't looked at the new treesitter stuff, but maybe something from there?

Eventually yes, but I'm stuck on supporting Emacs 27.1 for a while.

hmelman:
> How's this for the docstring of `embark-define-overlay-target`?

Looks good.  My only comment is to stick to only one way to describe if an argument is optional ("optional symbol" or "if it is non-nil").  I didn't see a preference in the elisp-ref Documentation Tips.  I prefer "optional argument" but I agree "if non-nil" is more technically specific and I think both are understandable.  

And two nits: in "for which PRED evaluates to a non-nil value." I think you can just say "for which PRED evaluates to non-nil."  Also be careful with the quoting around %o and %p in the actual lisp code.

oantolin:
Thanks, @hmelman! I agree "optional" reads better. 0e9ad9f43b5e599c1ec499bafa7a0d6f27ff338b

oantolin:
Most of the improvement is just cosmetic, but there is something that was missing from the original docstring: that `PRED` doesn't _replace_ `%p` being non-nil, instead _both_ things must be true.

oantolin:
I decided to add @minad's `embark-define-regexp-target` but with a default limit of the current line, and the possibility of specifying a different limit. e4fa12a11c387e7004af009d2cef69cef56bcdd1

minad:
Great! Can we use this anywhere? At least in the README?

oantolin:
Yes, I still need to update the manual.

PR #630, opened 2023-04-30
Fix: Rename category embark-selected to embark-selected-overlay

minad:
Otherwise the face property added by `defface embark-selected` conflicts with the overlay category properties. This leads to very weird errors in the messages buffer like `Invalid face reference: 1234` and the overlays are not displayed properly.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -555,10 +555,10 @@ action."
 
 ;; high priority to override both bug reference and the lazy
 ;; isearch highlights in embark-isearch-highlight-indicator
-(put 'embark-target 'face 'embark-target)
-(put 'embark-target 'priority 1001)
-(put 'embark-selected 'face 'embark-selected)
-(put 'embark-selected 'priority 1001)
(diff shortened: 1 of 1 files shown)


oantolin:
Oh, thank goodness you figured this out. It was driving me  little crazy.

oantolin:
Oh, thank goodness you figured this out. It was driving me  little crazy.

minad:
Yes, this pattern is actually more dangerous than I initially thought. For example Flymake uses the `flymake` category but the `defgroup flymake` also adds to the `(symbol-plist 'flymake)`. In Jinx I also renamed to `jinx-overlay` to ensure that no clashes can happen at all.

minad:
Thanks for the quick merge!

oantolin:
No, thank you for the fix! Sometimes the selection overlay wasn't displaying at all despite `C-u C-x =` saying the overlay was there and it had me totally baffled.

PR #632, opened 2023-05-03
Silence byte-compiler

tarsius:
```
lib/embark/embark-consult.el:263:20: Warning: ‘assq’ called with literal list that may never match (arg 1)
lib/embark/embark-consult.el:266:20: Warning: ‘assq’ called with literal list that may never match (arg 1)
```


oantolin:
Thanks!

tarsius:
You are welcome!

PR #635, opened 2023-05-07
Better embark--display-string

minad:
Taken from vertico--display-string. Inherit face properties. To observe the improvement run consult-line in an Org buffer with TODOs and with org-modern-mode enabled and collect a snapshot. The TODO properties added by org-modern-mode are not displayed nicely otherwise.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -2935,22 +2935,24 @@ For non-minibuffers, assume candidates are of given TYPE."
                     (if-let (a (funcall annotator c)) (list c "" a) c))
                   candidates)))))
 
-(defun embark--for-display (string)
-  "Return visibly equivalent STRING without display and invisible properties."
-  (let ((len (length string)) (pos 0) chunks)
-    (while (/= pos len)
-      (let ((dis (next-single-property-change pos 'display string len))
-            (display (get-text-property pos 'display string)))
-        (if (stringp display)
-            (progn (push display chunks) (setq pos dis))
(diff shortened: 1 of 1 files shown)


oantolin:
So this is something that hasn't hit Vertico yet, right? Because I see the same slightly incorrect display of TODOs in Vertico right now.

oantolin:
Ah, I see you fixed this just 6 hours ago, it'll land pretty soon.

oantolin:
Ooh, this is much nicer, thanks!

minad:
Thanks for merging. This is most likely still not strictly correct, but the function is supposed to return a display string approximation only, which works in most common scenarios.

oantolin:
The slightly wrong display of TODOs is really weird, I see the OD in the right font and the T and final O without any styling. Why is that?

oantolin:
image

minad:
The reason is that org-modern attaches a display property "T" -> " T" to the first character. This way org-modern fakes additional padding for the labels to make them more pretty. The face property of the displayed string however is supposed to be inherited from the original text.

oantolin:
Ah, I see. And the same goes for the final "O".

minad:
I could have also fixed this in org-modern to some extent, but I decided that is preferable to make `vertico--display-string` better. The`display` string property also had the issue that completion style faces were lost. This is not the case anymore with this change.

Issue #629, opened 2023-04-30
Unselect candidates after they've been acted on?


localauthor:
I'm playing with the new `embark-select` feature, for the purpose of integrating it with some of my packages. When acting on selected candidates in a buffer, the candidates stay selected after they've been acted on. What do you think about adding an option to unselect candidates after they have been acted on---something like `embark-quit-after-action`?

oantolin:
This seems like a reasonable option to add, I only hesitate because Embark already has so many options.

For what it's worth, dired doesn't unmark files after you run a command on them either.

localauthor:
Two fair points. No worries then. ;)

oantolin:
Well, I'm certainly open to further discussion on this. As I said, I think it is a reasonable thing to want.

localauthor:
On the issue of unselecting candidates, and perhaps a way forward with this: there might be a reason to have a post-action-hook that clears `embark--selection` for certain actions. For example, when using `delete-region` on candidates --- the regions are gone but the candidates are still in `embark--selection`, and even the overlays seem to remain.

oantolin:
I've actually made us of the fact that the selection remains after `delete-region`: after `delete-region` you can still `embark-insert` the deleted text to bring it together, for example. But I agree it feels a little weird to have the selection preserved when it is no longer present in the buffer.

localauthor:
This is a bit of a problem from a UX perspective, since there is no visual indicator that there are still selections in the buffer. So if you later select other candidates in the same buffer, you end up with more candidates than you might expect.

oantolin:
Yeah, I've been thinking that maybe I should add an indicator (this is a technical term: the functions that go in `embark-indicators`) saying how many items are selected.

localauthor:
That could certainly be helpful. How best to go about it, another question entirely :)

minad:
I think it is okay to keep the candidates selected, in order to act again for example. It is also easy to unmark all via `C-. A SPC`. What about adding an Embark selected indicator to the modeline?

minad:
Better variant:

~~~ elisp
(advice-add #'embark-select :after (lambda (&rest _) (force-mode-line-update t)))

(defun embark--selection-mode-line ()
  (when-let ((sel
              (buffer-local-value
               'embark--selection
               (or (when-let ((win (active-minibuffer-window)))
                     (window-buffer win))
                   (current-buffer)))))
[...]

PR #636, opened 2023-05-17
Add embark-selection-indicator (See #629)

minad:
The mode line indicator proposed in https://github.com/oantolin/embark/issues/629#issuecomment-1529483532.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,4 +1,8 @@
 #+title: Embark changelog
+
+* Development version
+- Added a mode line indicator showing the number of selected targets in
+  the current buffer (contributed by @minad, thanks!)
 * Version 0.22.1 (2023-04-20)
 ** New feature: selections
 Now users can select several targets to make an ad hoc collection. The
--- a/README.org
+++ b/README.org
@@ -245,7 +245,12 @@ member of the current selection. Similarly if no targets are selected
 and you are in a minibuffer completion session, running =embark-select=
 from =embark-act-all= will select all the current completion candidates.
 
-This functionality is supported everywhere:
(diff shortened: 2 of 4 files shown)


oantolin:
Thanks for doing this!

I haven't tested the code yet, but here are my initial reactions:

1. Didn't you run into some problem by using literal fontified strings once? Something to do with byte compilation, I think.

2. I like using `embark-selection-indicator` as both a boolean and a format string!

3. This updates the mode line after each individual item is deselected. When doing `C-. A SPC` to deselect all candidates, I wonder if maybe that's too much updating. Have you tried? Do you see a visible count down?

4. I don't have much experience with the mode line, but I do remember you aren't supposed to do anything expensive in an `:eval`. I think this is probably fine, but I think best practice is supposed to be to just have a global variable you write to, so that `:eval` only does a variable lookup.

minad:
1. Yes, in native compilation in an Emacs 28 development version. This has been fixed a long time ago. I use propertied strings at many places in my packages.

3. force-mode-line-update just sets a variable. It doesn't redisplay. See the C source.

4. Yes. I think this should be fine. We could save the string formatting, but formatting a single string should be cheap enough. We need the buffer local lookup in any case such that the active minibuffer is taken into account (We could change the message to EmbarkMinibuffer in that case. Do we want that?).

oantolin:
Great! I think this looks pretty good then. I'll test drive it a bit to see if it feels confusing going from a buffer with a selection to a minibuffer with a selection, but my initial feeling is that this is fine as is, that we don't need to change the message to EmbarkMinibuffer.

oantolin:
Thank you!

PR #638, opened 2023-05-28
Add embark-isearch-backward

alternateved:
Would it be possible to add `isearch-backward` action to embark? I often find that I'm searching for a region backwards and although there is default embark action to search forwards, there is no to search backwards.


oantolin:
Ha! Thanks, but I just added an embark-isearch-backward and renamed the existing embark-isearch to embark-isearch-forward.

alternateved:
That's great! I was missing it quite a bit.

PR #643, opened 2023-06-16
Fix projectless search of Consult async commands

minad:
Fixes two things:

1. Always call the action, even if associated directory is nil.
2. If target is a directory itself, use it as associated directory.

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -387,9 +387,10 @@ for any action that is a Consult async command."
 The values of TYPE which are considered to have an associated
 directory are: file, buffer, bookmark and library.  The REST of
 the arguments are also passed to RUN."
-  (when (embark--associated-directory target type)
-    (let (consult-project-function)
-      (apply run :target target :type type rest))))
+  (if (embark--associated-directory target type)
+      (let (consult-project-function)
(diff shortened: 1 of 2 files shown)


oantolin:
I'll definitely merge the change to the projectless function. As for the other, I'm not at my computer and can't test, but I thought that already worked. Is the problem file paths that are directories but don't end in a slash?

minad:
> Is the problem file paths that are directories but don't end in a slash?

Yes, for example in Dired when acting on a directory with `C-. C g`, the file name doesn't end with a slash. However I don't know if `embark--associated-directory` is the right place to fix the issue. Maybe the file target finders itself should better be fixed to return the paths in a canonical form.

oantolin:
Well, let's fix it here for now, but consider whether the file target finder should insist on slashes for directories.

oantolin:
Thanks!

oantolin:
I think it's worth fixing the slashes at the end of directories in the file target finder. It seems to me that directories in the minibuffer always already end in a slash, and the only issue is slashless-directory names found in regular buffers. The file target finder already accesses the file system, so we might as well do the directory check there.

PR #654, opened 2023-08-17
Only pop-to-buffer once when exporting with different variants

amosbird:
This PR eliminates unnecessary `pop-to-buffer` calls so that window management will be easier (for instance in doomemacs). It also adds a variant `embark-export-current-window` which allows to export into current window. This doesn't work for `embark-collect` because it uses different mechanism somehow.

I've already assigned FSF copyright.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -3165,6 +3165,9 @@ The parameter KIND should be either `embark-export' or `embark-collect'."
       (funcall embark--rerun-function)
     (user-error "No function to rerun collect or export found")))
 
+(defvar embark--buffer-display #'pop-to-buffer
+  "Buffer display function.")
+
 ;;;###autoload
 (defun embark-export ()
   "Create a type-specific buffer to manage current candidates.
@@ -3194,13 +3197,13 @@ buffer."
                (cmd embark--command)
                (name (embark--descriptive-buffer-name 'export))
                (rerun (embark--rerun-function #'embark-export))
-               (buffer (save-excursion
-                         (funcall exporter candidates)
+               (buffer-display embark--buffer-display)
+               (buffer (with-current-buffer (funcall exporter candidates)
                          (rename-buffer name t)
                          (current-buffer))))
           (embark--quit-and-run
            (lambda ()
-             (pop-to-buffer buffer)
+             (funcall buffer-display buffer)
              (setq embark--rerun-function rerun)
              (use-local-map
               (make-composed-keymap
@@ -3211,6 +3214,13 @@ buffer."
              (let ((embark-after-export-hook after)
                    (embark--command cmd))
                (run-hooks 'embark-after-export-hook)))))))))
+
+;;;###autoload
+(defun embark-export-current-window ()
+  "Variant of `embark-export' which exports to current window."
+  (interactive)
+  (let ((embark--buffer-display #'switch-to-buffer))
+    (embark-export)))
 
 (defmacro embark--export-rename (buffer title &rest body)
   "Run BODY and rename BUFFER to Embark export buffer with TITLE."
@@ -3296,7 +3306,7 @@ PRED is a predicate function used to filter the items."
     (with-current-buffer buf
       (dired-unadvertise (car dired-directory)) ; avoid reuse of this buffer
       (rename-buffer (format "*Embark Export Dired %s*" default-directory)))
-    (pop-to-buffer buf)))
+     buf))
 
 (autoload 'package-menu-mode "package")
 (autoload 'package-menu--generate "package")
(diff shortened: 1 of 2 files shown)


minad:
If I recall correctly we had discussed if it would be possible to only create a buffer in an exporter and decided against this, such that the exporter has more flexibility. For example `embark-export-customize-variable` creates a buffer and displays it directly. You did not update all the exporters accordingly. If this PR is to be merged, all the exporters must be updated for consistency.

But then, I think this PR is unnecessary since you can use the intended mechanism for display action customization, namely the `display-buffer-alist`. Embark makes this easy since Embark export buffers follow a consistent naming convention. Also the `embark-export-current-window` command is quite ad-hoc and may belong to the wiki instead. You can implement the same command right now, by let-binding `display-buffer-alist` or `display-buffer-overriding-action` around the `embark-export` call.

embark.el
```diff
@@ -3194,13 +3197,13 @@ buffer."
...
                (rerun (embark--rerun-function #'embark-export))
-               (buffer (save-excursion
-                         (funcall exporter candidates)
+               (buffer-display embark--buffer-display)
+               (buffer (with-current-buffer (funcall exporter candidates)
                          (rename-buffer name t)
                          (current-buffer))))
           (embark--quit-and-run
            (lambda ()
-             (pop-to-buffer buffer)
```
minad, line 3203: Looking at this code, I wonder why this call is here, since it somehow contradicts the original implementation where the exporter also creates windows. The reason is probably that some exporter created windows and some do not.

embark.el
```diff
@@ -3165,6 +3165,9 @@ The parameter KIND should be either `embark-export' or `embark-collect'."
       (funcall embark--rerun-function)
     (user-error "No function to rerun collect or export found")))
 
+(defvar embark--buffer-display #'pop-to-buffer
+  "Buffer display function.")
```
minad, line 3169: I think this variable should go away, instead one should use the `display-buffer-alist` mechanism. 

I suspect that you copied this variable over from Consult which has a similar variable to implement `consult-buffer-other-window`. Having written that code in Consult, I can only say that the design is not great and should probably be changed there too. The reason for the Consult design was mostly a difficulty of getting Consult preview to work properly, which is really hard with display actions. Anyway, we should not carry over that variable from Consult.

embark.el
amosbird: Yes. The original motivation of this PR is to achieve consistency here. For now we did renaming the buffer twice.

embark.el
amosbird: >  instead one should use the display-buffer-alist mechanism.

Could you post an example of how to let-binding `display-buffer-alist` in a way that it will use `switch-to-buffer` and will live long enough for `embark--run-after-command`, which is run out of the scope of `embark-export`.

minad: You are right, let-binding won't work because of the delay. This means one has to temporarily modify the `display-buffer-alist` and undo the modification in an `embark--run-after-command` function.

embark.el
minad: > Yes. The original motivation of this PR is to achieve consistency here.

That's good. 

> For now we did renaming the buffer twice.

Yes, seems to make sense to clean that up, but it seems there are more exporters which may need similar changes.

oantolin:
I don't think I want the `embark--buffer-display` variable, it feels unnecessary given `display-buffer-alist`. I also think the `embark-export-current-window` is probably better for the wiki, since it feels too specialized. To make implementing `embark-export-current-window` easier, I would welcome a change to `embark-export` that captures the value of `display-buffer-alist` and let-binds it again in the call to `embark--run-after-command`.

minad:
> To make implementing embark-export-current-window easier, I would welcome a change to embark-export that captures the value of display-buffer-alist and let-binds it again in the call to embark--run-after-command.

I am not sure if this is a good idea. First, there are multiple `display-buffer-alist` action variables (override, fallback, ...), which would potentially all require binding. Second, I don't think it is worth prematurely adding a complication to simplify the implementation of a `embark-export-current-window` command. At first I'd try the approach mentioned in https://github.com/oantolin/embark/pull/654#discussion_r1297217311, but I haven't tested this so far.

oantolin:
Mmh, I hadn't taken into account the other `display-buffer-*` variables, @minad. You're right that one should at least try the manual set-restore method to implement `embark-export-current-window`.

@amosbird I still wanted the removal of the duplicate pop-to-buffers! You needn't close the pull request. Although I guess now that you've spotted the duplication I can remove those myself. Thanks for the PR, even if I didn't agree with adding a current window variant to embark. I definitely want it to be possible to implement that variant, and would like it documented on the wiki.

amosbird:
> I can remove those myself. Thanks for the PR

No problem! Embark is quite good to play with. Thanks for all the efforts.

PR #662, opened 2023-09-15
Meta: Add makem.sh, et al for linting and CI

alphapapa:
Hi Omar,

I saw https://github.com/oantolin/embark/commit/0d89add290f9176b77a2d7155a9935e30351d90f and thought that this might help prevent minor oversights like that in the future.  Feel free to reject this PR if it's not your style, I'm just presenting it for your consideration.  Embark is so useful, that if makem.sh can be helpful to the project, I'd be glad.

Anyway, even if this isn't merged, you might find some of these lint warnings helpful, especially on older Emacs/Org versions, e.g. https://github.com/alphapapa/embark/actions/runs/6203814262/job/16844997873#step:6:157

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+makem.sh linguist-vendored
+Makefile linguist-vendored
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,81 @@
+# * test.yml --- Test Emacs packages using makem.sh on GitHub Actions
+
+# URL: https://github.com/alphapapa/makem.sh
+# Version: 0.6-pre
+
+# * Commentary:
+
+# Based on Steve Purcell's examples at
+# <https://github.com/purcell/setup-emacs/blob/master/.github/workflows/test.yml>,
+# <https://github.com/purcell/package-lint/blob/master/.github/workflows/test.yml>.
+
+# * License:
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+# * Code:
+
+name: "CI"
+on:
+  pull_request:
+  push:
+    # Comment out this section to enable testing of all branches.
+    # branches:
+    #   - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        emacs_version:
+          - 26.3
+          - 27.2
+          - 28.2
+          - 29.1
+          - snapshot
+    steps:
+    - uses: purcell/setup-emacs@master
+      with:
+        version: ${{ matrix.emacs_version }}
+
+    - uses: actions/checkout@v2
+
+    - name: Install Ispell
+      run: |
+        sudo apt-get install ispell
+
+    - name: Initialize sandbox
+      run: |
+        SANDBOX_DIR=$(mktemp -d) || exit 1
+        echo "SANDBOX_DIR=$SANDBOX_DIR" >> $GITHUB_ENV
+        ./makem.sh -vv --sandbox=$SANDBOX_DIR --install-deps --install-linters
+
+    # The "all" rule is not used, because it treats compilation warnings
+    # as failures, so linting and testing are run as separate steps.
+
+    - name: Lint
+      # NOTE: Uncomment this line to treat lint failures as passing
+      #       so the job doesn't show failure.
+      # continue-on-error: true
+      run: ./makem.sh -vv --sandbox=$SANDBOX_DIR lint
+
+    - name: Test
+      if: always()  # Run test even if linting fails.
+      run: ./makem.sh -vv --sandbox=$SANDBOX_DIR test
+
+# Local Variables:
+# eval: (outline-minor-mode)
+# End:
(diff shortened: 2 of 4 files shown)


oantolin:
Thank you, I really appreciate this. I know I can use help catching silly mistakes like that one you saw, but I've never really had patience for linters. In my very limited experience it seems like it's very hard to write one that doesn't mostly give false positives. For example, here I think that only the thing about  `org-fold-reveal` and maybe `package-desc-name` are actual problems, in all 400 lines of output. And if I missed something else important, that just proves my point about how hard it is to find the real problems among the false positives.

I guess if you commit to a particular linter you gradually learn how to warp your code so that is both correct and also satisfies the linter, so that you don't have to wade through so much output, right?  (Although here I'm not sure how I'd get rid of the incorrect complaints about `defvar-keymap`. Maybe declaring it as coming from `compat`?)

Please note that I don't mean to criticize your linter specifically, as this has always been my experience with them and basically the only linter-like things I find myself sticking with are compiler warnings, like `gcc -Wall` or the byte-compiler in Emacs (which I forgot to run after 0d89add290f9176b77a2d7155a9935e30351d90f, so I could definitely use some automation!). But maybe this is just because as a non-professional, I've never had to satisfy anyone's code guidelines, and thus haven't seen the benefits of doing so.

I'll definitely think about merging this, after maybe trying it a bit myself. I do use the byte-compiler, `checkdoc` and `package-line-current-buffer` (of the 3 the byte-compiler is by far the best, having very close to 0 false positives), but I do not run them automatically and sometimes forget. So maybe I should automate that, at least, byte-compile on buffer save sounds like a good idea.

alphapapa:
Oh, you're definitely right about linters being noisy and sometimes impractical to satisfy.  This is certainly the case here, especially when testing on multiple Emacs versions.  It's often impossible to get it to lint cleanly on several versions at once.

I will point out, though, that `makem.sh` is highly modular and configurable, so if, e.g. you only wanted to run the byte-compile linter, you would just change this line:

https://github.com/oantolin/embark/pull/662/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R73

Instead of `lint`, call `lint-compile` (or any combination of the rules, like `lint-compile lint-declare`, etc).

You can also edit the `test.yml` file so that the linter will run but not count as a failure in the GitHub Action if it returns non-zero, and then you can just check the job results when you're interested.

Of course, you might also try adding `flymake-mode` to your `emacs-lisp-mode-hook`, as it will flag most of the same issues within the buffer.  For me, it's easy to overlook those, and I like seeing a project-wide lint/test run in the compilation buffer, so I still use `makem` before pushing significant changes on my projects.

Regarding the `defvar-keymap` warnings: `makem.sh` doesn't implement most of the linters itself; it just wraps them.  That warning comes from the `package-lint` linter (https://github.com/purcell/package-lint); so I guess the the problem is that `package-lint` isn't accounting for the dependency on `compat`.

Anyway, the bottom line for me is that these linters help me catch certain kinds of bugs before they reach users (especially bugs regarding supporting multiple versions of Emacs and Org), and therefore before they file a bug report about it.  :)  But like any automated tool, the results have to be interpreted with a grain of salt.

minad:
Oh this adds so much code. @alphapapa Can you create a reusable GitHub action instead, similar to purcell/setup-emacs@master? I think tarsius also maintains such a reusable action which does linting and which does not duplicate large scripts across repositories. This makes it easier to keep the linter maintained and up to date.

alphapapa:
> Oh this adds so much code. @alphapapa Could you please turn this into a reusable GitHub action, similar to purcell/setup-emacs@master? I think tarsius also maintains such a reusable action which does linting and which does not duplicate large scripts across repositories. This makes it easier to keep the linter maintained and up to date. I would find a linter setup useful, which I could easily add to many projects by only adding the CI workflows yaml.

I don't know, maybe.  It's always been my intention for `makem.sh` and its GitHub Actions configuration to be customizeable by each project as-needed, and for it to be upgraded intentionally by the package maintainer.  It's always annoying to finally have achieved a clean linting, only for the linting tool to automatically change and not lint cleanly anymore.

The idea is for the package maintainer to be in control, not upstream package maintainers, nor the maintainer of some YAML file (though I greatly appreciate Steve's maintenance of the Emacs setup action, which `makem.sh`'s YAML file uses).

However, when used on GitHub Actions, `makem.sh` already uses the latest versions of the linting tools each time it runs, and `makem.sh` rarely changes, so I don't think being outdated is much of a problem.

Of course it would reduce the number of lines added to package repos, but another purpose of `makem.sh` is to work locally the same way it works on remote CI; it seems like a severe anti-pattern to have to push commits to a remote clone in order to run linters.  I'm constantly running it locally, which only takes seconds or less, whereas pushing to GitHub and waiting minutes for an Action to run would be tedious and impractical.  It is useful in the way that it makes it easy to lint and test on multiple Emacs versions, rather than having to maintain all of those locally, but for general development, doing so on my local Emacs version or two is enough for me.

So, I'm not opposed to offering a way to run `makem.sh` as a GitHub Action (and someone already [suggested](https://github.com/alphapapa/makem.sh/issues/33) a way to do so with Docker), but I don't know if I will invest the time to do that myself, because I wouldn't use it that way, and I generally encourage package authors to add the script to the repo and customize it if needed (perhaps paradoxically, as I also intend it to be usable without configuration...).

The bottom line, to me, is that it Just Works, so it's not as if it needs to be "maintained," so there's no reason not to add it.  (Or if having it in a project root directory seems messy, that will be addressed by https://github.com/alphapapa/makem.sh/pull/40.)

minad:
Sure. I understand your points. It would just be my preference to not include any scripts, while I would use some action prepared by the upstream maintainer, in this case by you or Steve. If there is interest I may try to prepare such an action and contribute it to your repository. Regarding breakage I don't worry much. I think GitHub actions have some versioning mechanism too.

minad:
> it seems like a severe anti-pattern to have to push commits to a remote clone in order to run linters.

I have my local package-lint-flymake installed in Emacs for direct feedback and I don't need makem.sh for that. Also I don't keep multiple Emacs versions around locally, usually just the current one. Therefore CI for such extended linting adds quite some value, even if it runs only after pushing as a secondary assurance.

josephmturner:
FWIW, `makem.sh` can now be used as a git submodule.

oantolin:
I really should have closed this a long time ago, since I already knew back when we had this discussion I was not going to merge. As I said back then, I do really appreciate the PR, though.

PR #665, opened 2023-09-20
Fix custom type specifications

tarsius:
The byte-compiler learned about some more common mistakes, which this fixes for this package.


oantolin:
Thank you!

tarsius:
A typical case of "as soon as you release, you get a new pull-request". :grin: 

By the way, I just noticed that some tags seem to be missing, e.g., 0.22 and 0.20.

PR #676, opened 2023-11-04
Fix dired export

minad:


oantolin:
Thanks for fixing this! I was busy moving this week and couldn't get to it.

(By the way, checkdoc tricked you into making a grammatical error in the docstring!)

minad:
> (By the way, checkdoc tricked you into making a grammatical error in the docstring!)

I wouldn't care as long as the warning goes away. :-P

But "target the guess at point" is grammatically correct, isn't it?

oantolin:
Yes, but "file ffap guess" sounds off to me. Changing it to either "ffap file-guess" or "file guessed by ffap" sounds fine.

PR #678, opened 2023-11-10
Change key for org-sort to not conflict with outline-move-subtree-up

ashton314:
embark-org binds `^` to `org-sort` which overrides the binding for `outline-move-subtree-up`. This means that, when invoking `embark-act` on an org-mode heading, it's easy to move a subtree down, but you can't move it back up.

This little PR binds `org-sort` to `~`, which doesn't seem to have any conflicts on org headings. No other motivation for picking the `~` key.

This certainly doesn't fit the bill of a significant contribution, but I *have* signed a copyright assignment to the FSF, so we're doubly-covered in that regard.

### DIFF
--- a/embark-org.el
+++ b/embark-org.el
@@ -398,7 +398,7 @@ bound to i."
   "T" #'org-tree-to-indirect-buffer
   "<left>" #'org-do-promote
   "<right>" #'org-do-demote
-  "^" #'org-sort
+  "o" #'org-sort
   "r" #'org-refile
   "R" #'embark-org-refile-here
   "I" #'org-clock-in


oantolin:
Hahaha! I hadn't noticed the shadowing. I do want to change the keybinding, but am not convinced `~` is the best choice for `org-sort`. Unfortunately `s` is taken by `org-schedule`. How about `|` or `/`? Don't those look more like sorting than `~`?

oantolin:
Also available: `o` for "order".

hmelman:
FWIW, I like `|`, though it should never shadow it's binding as `shell-command-on-region`.  Unrelated to `embark` or `org` I do this (though I realize now maybe swapping these is better):

```elisp
(global-set-key (kbd "s-_") #'sort-lines)
(global-set-key (kbd "s-|") #'sort-columns)
```

ashton314:
Personally I like the `o`, as it's the second letter in "sort" as well as the first in "order".

I imagined the `~` as the path items take as they swap places during sorting, but that might not be a very transferable image.

My vote is `o`. Shall I change the PR to that?

oantolin:
Yes, please. I do think `o` is probably the best one considered so far.

ashton314:
@oantolin It is done

oantolin:
Thanks!

PR #685, opened 2023-11-25
Add embark-open-externally for urls and bookmarks

minad:
- For urls it makes a difference if `browse-url-browser-function` is configured to use eww by default. Then the default action and "b" will use eww and the "x" action will use the external browser.
- For bookmarks this is a new addition and allows you to open locations externally (files or urls).
- I like the uniformity of pressing "x" to open externally.
- Do not use `bookmark-location` since it does not return nil, but a string if the location is missing.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -3798,8 +3798,11 @@ with command output.  For replacement behavior see
                              (and replace (current-buffer)))))
 
 (defun embark-open-externally (file)
-  "Open FILE using system's default application."
-  (interactive "fOpen: ")
+  "Open FILE or url using system's default application."
+  (interactive "sOpen externally: ")
+  (unless (string-match-p "\\`[a-z]+://" file)
+    (setq file (expand-file-name file)))
+  (message "Opening `%s' externally..." file)
   (if (and (eq system-type 'windows-nt)
            (fboundp 'w32-shell-execute))
       (w32-shell-execute "open" file)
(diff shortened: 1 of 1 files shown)


oantolin:
It definitely does seem reasonable to me: I have equivalent functionality (only for URLs, not bookmarks) in my personal configuration. Soon this will quite down at work and I'll have a chance to test this, but I can safely say it is something that would be good to have in Embark.

oantolin:
Now that we're doing this, I'm tempted to make it work on WSL too. On WSL `system-type` is `gnu/linux`, but instead of using `xdg-open` you may want to use the windowsy version, namely `wslview`. Now, some people on WSL may prefer `xdg-open` and some may prefer `wslview`, so it seems to me a user option is warranted. I'd propose adding an `embark-open-externally-program` so people can configure this. Even on other OSes with a single canonical thing-opener, people might have their own pets view commands they've written as shell scripts. We could even allow a function that picks a program based on the file name, in case people want more control inside Emacs.

What do you think?

oantolin:
What I'm doing right now for WSL is just symlinking `wslview` to `xdg-open` (since I don't use `xdg-open`). Maybe symlinking is good enough and we don't need configurability on the Emacs side.

minad:
This recent patch from the Emacs bug tracker may be useful if you want to support more platforms: https://lists.gnu.org/archive/html/bug-gnu-emacs/2023-11/msg01599.html

hmelman:
> This recent patch from the Emacs bug tracker may be useful if you want to support more platforms: https://lists.gnu.org/archive/html/bug-gnu-emacs/2023-11/msg01599.html

Oh that's good to see, I remember many debates about adding such a thing.  I do really wish that emacs had a command that takes a filename as an argument and opens it externally.  Bonus for taking several files and opening them efficiently (via one command if the system supports it or multiple if it doesn't).  While the obvious use is in dired (and it's been countlessly recreated), I'd want it for other file types I might open in emacs and want to see in other apps (.html, .md, .pdf, images, etc.).  If you're writing such a thing for embark, please consider proposing it for emacs.

oantolin:
@hmelman Embark has had `embark-open-externally` for a long time now, is it not what you want? (It doesn't attempt your bonus points of being efficient for multiple files).

hmelman:
@oantolin Yes I know, and I've had my own version for dired use for a long time.  I just think emacs should have a version of this function builtin (and not just as a guess for dired-do-shell-command) since it's recreated so many times.

oantolin:
Oh, I see, I had misread.

PR #686, opened 2023-11-25
Fix wgrep-header/footer-parser problem (See #602 and #604)

minad:


oantolin:
Thanks! Let's keep both variable names until the dust settles.

Issue #707, opened 2024-03-19
embark-which-key-indicator cannot handle paging


richcarl:
I was hoping that I could get the C-h paging in which-key to work with Embark via the `embark-which-key-indicator` from the Wiki (https://github.com/oantolin/embark/wiki/Additional-Configuration#use-which-key-like-a-key-menu-prompt) by doing `(setq embark-help-key "?")` to leave `C-h` free for which-key.

I found that the indicator function in the Wiki passes `t` for the `no-paging` argument in the call to `which-key--show-keymap`, and I tried to change that to `nil` to allow which-key to react to the paging keypress. This works on the which-key side, but once you have pressed e.g. `C-h n` to page forward and the call returns, I got `"Error running timer: (error "embark-next-symbol is not a keymap")`, and I'm not good enough at elisp to see what's going on. Maybe someone else can figure out how to fix this so that the which-key indicator function can do paging; it would be nice to have.

oantolin:
I'd love to have this working, but I never was able to figure it out myself. I don't use which-key so I also didn't spend a ton of time on it, but I did try for a while.

oantolin:
In the meantime, might I suggest trying `embark-minimal-indicator` and if you need a reminder of the key bindings, pressing `embark-help-key` (`C-h` by default)? I find that much more convenient that either the verbose or the which-key indicators. If you use Vertico, you can even make the display very which-key-like with:

```elisp
(cl-pushnew '(embark-keybinding grid) vertico-multiform-categories)
```

You can select a command by completion (and you can complete on either the command name or the string form of the key binding), or, if you press `@`, you can then type the key binding itself.

richcarl:
`embark-minimal-indicator` + `embark-help-key` with Vertico grid layout was a great suggestion! I hadn't discovered `vertico-multiform-mode` until now.

oantolin:
I'm glad you liked that suggestion. It's also related to what I use instead of which-key generally:

```elisp
(setq prefix-help-command #'embark-prefix-help-command)
```

With that configuration pressing `C-h` after a prefix pops embark's completing-read prompter in the minibuffer, listing the command under the prefix and their bindings. (It's just like pressing `embark-help-key` during `embark-act` but for arbitrary prefixes.)

Personally, I like this a lot more than which-key for several reasons:

[...]

PR #708, opened 2024-03-21
Add tips for Vertico users to get which-key like behaviour

richcarl:
Tried to summarize the conclusions from #707 in a form that would be helpful to others.

### DIFF
--- a/README.org
+++ b/README.org
@@ -480,41 +480,10 @@ definition from the wiki into your configuration and customize the
 =embark-indicators= user option to exclude the mixed and verbose
 indicators and to include =embark-which-key-indicator=.
 
-If you use [[https://github.com/minad/vertico][Vertico]], there is an even easier way to get a =which-key= like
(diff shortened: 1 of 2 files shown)


oantolin:
Thank you for this! I probably should have added something like this to the documentation a while ago, but I guess I needed some more external validation that this is a good configuration to recommend. 😛

I'll proof read and check it's in the best location within the documentation before merging, probably this evening.

richcarl:
Glad to be able to contribute a little!

minad:
I also like this setup with grid mode :)

oantolin:
Hi @richcarl. I thought some of the information you added really belonged to different sections so I moved some stuff around (and added some new stuff too, on `embark-bindings`). What do you think?

richcarl:
Looks great!

oantolin:
Merging, then. Thanks again for getting the ball rolling on this!

PR #711, opened 2024-03-26
Add `embark-consult-export-xref-grep'.

amosbird:
Since https://github.com/oantolin/embark/pull/560 has been merged, we can also export xref items into a grep buffer.

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -149,6 +149,23 @@ category `consult-line'."
       (occur-mode))
     (pop-to-buffer buf)))
 
+(defun embark-consult-export-location--insert-footer (non-file-buffers)
+  "Insert the names of `NON-FILE-BUFFERS'.
+Any locations that come from a buffer which is not visiting a file
+will be inserted at the end of the grep buffer."
+  (when non-file-buffers
+    (let ((start (goto-char (point-max))))
+      (insert "\nSome results were in buffers with no associated file"
+              " and are missing\nfrom the exported result:\n")
+      (dolist (buf non-file-buffers)
+        (insert "- " (buffer-name buf) "\n"))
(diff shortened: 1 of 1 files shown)


Kaspazza:
@oantolin Will this make consult-xref to be able to be exported to wgrep?

minad:
@Kaspazza I believe so. I haven't tested the feature though. Did you try it, if it works for your use case?

Kaspazza:
@minad I've just tried it, and it's still not working with message 
"embark category project-file doesn't support writable export"

 I'm using doom emacs
GNU Emacs 28.2 (build 2, aarch64-apple-darwin23.4.0, NS appkit-2487.50 Version 14.4 (Build 23E214)) of 2024-03-22

I've set embark to this commit:
(package! embark
  :pin "86fc40909ba9506676a28d35879f8c091018ac42"
  :recipe (:host github :repo "amosbird/embark"))
  
And made sure to bump the package and run doom sync -u and reload.

vseguip:
I tested this by copy&pasting the functions into my config and setting the exporter and they work (I can create a grep buffer and switch it to wgrep). I would also like to see this merged as it's super useful for my workflows.

oantolin:
This looks good and works fine in my testing. Now that Emacs has a built-in grep-edit-mode this is even slightly more useful than it was before.

minad:
@oantolin I am not sure if we still need this. The newest xref.el will have an xref-edit-mode (just added a few days ago), and we should rather encourage that people use it directly, instead of duplicating functionality unnecessarily. Note that while xref.el is builtin, it can also be installed from ELPA, so there is also no need to wait for the next Emacs release. The last xref.el on ELPA is from 2024, but I am sure there will soon be a new release, when Dmitry considers it sufficiently stable.

PR #727, opened 2024-07-16
Add default action for consult-xref

minad:
I applied a small [simplification](https://github.com/minad/consult/commit/41d382c9da9a84ab5391ee906e24f7126d03bd53) to consult-xref, but now we need an explicit default action override. I believe it is also advantageous and more efficient to use this override here, since we don't have to rerun the search action.

### DIFF
--- a/README.org
+++ b/README.org
@@ -398,8 +398,8 @@ About the suggested key bindings for =embark-act= and =embark-dwim=:
   installations of) GNOME to input emojis, and Emacs doesn't even get
   a chance to respond to the binding. You can select a different key
   binding for =embark-act= or use =ibus-setup= to change the shortcut for
-  emoji insertion (Emacs 29 will likely use =C-x 8 e e=, in case you
-  want to set the same one system-wide).
+  emoji insertion (Emacs 29 uses =C-x 8 e e=, in case you want to set
+  the same one system-wide).
 - The suggested alternative of =M-.= for =embark-dwim= is bound by default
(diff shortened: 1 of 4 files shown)


oantolin:
This looks good to me, but I'm a little confused about why the change is necessary now (and maybe also about why it was not necessary before)! Would you mind explaining a bit why let-binding this-command makes it so rerunning the search doesn't work? And: would this default action override also have worked before (just making things more efficient)?

oantolin:
Thanks for the miscellaneous cleanup (some of which is enabled by compat-30 if I understood correctly).

minad:
> Would you mind explaining a bit why let-binding this-command makes it so rerunning the search doesn't work? 

I think it doesn't work since `consult-xref` is not a command. This means Embark uses its non-command calling convention which is not supported by `consult-xref`.

> And: would this default action override also have worked before (just making things more efficient)?

Yes, definitely.

oantolin:
> I think it doesn't work since `consult-xref` is not a command. This means Embark uses its non-command calling convention which is not supported by `consult-xref`.

Got it. That must be it.

oantolin:
As a side effect of the let-binding, if you use embark-export now the exported buffer gets called consult-xref, but I guess that's not really a problem.

Thanks for the PR!

oantolin:
Wait, I think the exported buffer already recieved that name before and I'm a little confused about why that is. :|

PR #731, opened 2024-09-18
Require Emacs 28.1

minad:
With [Emacs 30](https://lists.gnu.org/archive/html/emacs-devel/2024-09/msg00305.html) on the final track I think we can safely drop support for Emacs 27.1 to reduce the amount of cruft we carry around. We still support the last three stable versions then. I've also made this move for [Marginalia](https://github.com/minad/marginalia/commit/16b7d03088656b597a253418e41a4a62c811993d) and [Consult](https://github.com/minad/consult/commit/eb71b39d282895b193a8002bddac5ca87891ac1e), so I think it won't hurt to also update the dependencies for Embark. I've not gotten any complaints. In any case, it should always be possible to use older versions on Emacs 27. For Embark, the code impact is not large, but in the case of Consult I could get rid of some larger workarounds for Emacs 27 bugs.

### DIFF
--- a/README.org
+++ b/README.org
@@ -797,7 +797,7 @@ similar situations where the easiest option is not available.
 
 *** New minibuffer target example - tab-bar tabs
 
-As an example, take the new [[https://www.gnu.org/software/emacs/manual/html_node/emacs/Tab-Bars.html][tab bars]] from Emacs 27. I'll explain how
+As an example let us take a look at the [[https://www.gnu.org/software/emacs/manual/html_node/emacs/Tab-Bars.html][tab bars]]. I'll explain how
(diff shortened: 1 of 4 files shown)


oantolin:
Sure, if you've received no complaints for Consult, which is probably quite a bit more popular than Embark, then lets do it here too. The current LTS release of Ubuntu has Emacs 29.1, which is also a good sign.

Thanks a lot for preparing the PR!

minad:
Thanks for merging. Regarding popularity you can take a look at the rank here: https://elpa.gnu.org/packages/. Afaik the rank is computed from recent downloads, but I don't know what recent means.

PR #734, opened 2024-10-02
Fix error in face definition

tarsius:
I don't actually know whether you actually meant to make this face italic, or the `italic` symbol should just be removed.


oantolin:
Whoops, thank you! I think it looks better italic and I believe that was the original intent.

PR #742, opened 2024-11-14
Simplify embark-target-completion-list-candidate for Emacs 28

minad:
@oantolin We can simplify since we've raised the Embark dependency to Emacs 28, which always provides `completion--string`. Please test.


minad:
We can wait for the new function: https://github.com/emacs-compat/compat/issues/57

PR #747, opened 2025-01-16
Add support for `grep-use-headings`

### LINKED ISSUE #719: embark-export: Support grep-use-headings
I've just migrated to Emacs 30. There is a new option `grep-use-headings` which we could also support in Embark grep export.

minad:
Fix #719

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -6,7 +6,7 @@
 ;; Maintainer: Omar Antolín Camarena <omar@matem.unam.mx>
 ;; Keywords: convenience
 ;; Version: 1.1
-;; Homepage: https://github.com/oantolin/embark
+;; URL: https://github.com/oantolin/embark
 ;; Package-Requires: ((emacs "28.1") (compat "30") (embark "1.1") (consult "1.8"))
 
 ;; This program is free software; you can redistribute it and/or modify
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -1,6 +1,6 @@
 ;;; embark-consult.el --- Consult integration for Embark -*- lexical-binding: t; -*-
 
-;; Copyright (C) 2021-2023  Free Software Foundation, Inc.
+;; Copyright (C) 2021-2025  Free Software Foundation, Inc.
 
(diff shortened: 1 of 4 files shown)


oantolin:
I don't have Emacs 30 yet, so I can't test that it works as intended. I can at least test that this doesn't break stuff on 29.4.

embark-consult.el
```diff
@@ -225,13 +225,18 @@ count of the matches (there may be more than one match per line).
...
-        (funcall footer)
-        (goto-char (point-min))
-        (grep-mode)
-        (setq-local grep-num-matches-found count
-                    mode-line-process grep-mode-line-matches))
+      (grep-mode)
+      (setq-local mode-line-process grep-mode-line-matches)
+      (let ((inhibit-read-only t))
+        (insert (propertize header 'wgrep-header t 'front-sticky t))
+        (dlet ((compilation-filter-start (point)))
```
oantolin, line 232: Woah! I never knew dlet existed! But, why use it here at all? `compilation-filter-start` is a dynamic variable, isn't it? So that would make this do the same thing with let in place of dlet.

minad:
> I can at least test that this doesn't break stuff on 29.4.

Yes, testing that this doesn't break things would be great. You see, I've reordered the code slightly.

minad:
@oantolin Does it work for you?

oantolin:
Yep, looks good to me (in other words, I see no difference with what it did before 😛 ). Are you done with it, should I merge?

minad:
Thanks for checking. Yes, please merge.

oantolin:
Thanks for making this change! This is actually tempting me a bit to install Emacs 30...

aaronjensen:
@minad at least on my machine, this appears to consistently fail to convert the last entry.

minad:
Fix at #747

oantolin:
Now that Emacs 30.1 was released I update, and this does look quite nice! Thanks again, @minad.

minad:
Don't thank me. I think @astoff added this feature to Emacs!

PR #749, opened 2025-01-31
grep export: Fix last heading

minad:
When grep-use-heading is used point must be at the end of the buffer.


aaronjensen:
This change worked for me. Thank you.

oantolin:
Thanks for fixing this, @minad.

Issue #470, opened 2022-03-11
Keystrokes 'hanging around' in echo area after embark-act


codesuki:
Hi! First off, awesome package! 

When I run `C-x C-f `(find-file), select a file, run `embark-act`, and press `RET` (find-file), the echo-area shows `RET-` waiting for a key-chord. If I then for example press `s` it shows `RET-s`, disappears, and inserts `s` into the buffer.

It doesn't really have a negative impact, but I was thinking maybe it hints at some problem (Why is it waiting for a command?). I was looking at the code of `embark--act` and it does manipulate `this-command-keys` which could be connected (I don't know enough about how it works).

oantolin:
I haven't been able to reproduce this yet. What completion UI do you use?

oantolin:
Ah, got it, if you using a quitting action and have a very small value of `echo-keystrokes` (below 0.2 in my tests, but this probably depends on your computer, completion UI and Emacs version) then this happens. Curious.

codesuki:
Thanks for investigating! And sorry for giving too little context. I am using Vertico, Corfu. Emacs is version 29.0.50, would have to check the actual commit :)

It's fine because it actually doesn't do anything. It doesn't even say `RET-s is undefined` which it normally should say.  I could probably increase `echo-keystrokes`, it's 0.25 right now. 

But, I agree it's curious. I don't know about the implementation that drives keystroke echoing, but I imagine it should react properly on the quit signal / command finishing. Somehow Emacs seems to lose track.

oantolin:
I just want to confirm that as you suspected this *is* caused by using `set--this-command-keys` in `embark--act`. That was done to fix #182. It would be nice to figure out a different fix for #182 that didn't have this lingering keys issue.

codesuki:
Oh, that's kinda good news :) 
I tried to reproduce the issue in #182, but my `browse-url` doesn't work even without embark. 😅

montag451:
Hi @oantolin,

Do you have a plan to fix this issue ?

oantolin:
I don't know _how_ to fix it and am open to suggestions.

montag451:
I can have a look but to be honest if you didn't find a solution to fix this issue I don't think I will be able to find one :). Anyway it's not a big issue so it's not a big deal if it's not fixed

ywwry66:
[...]

Issue #182, opened 2021-03-09
browse-url doesn't work as an embark-act action from the minibuffer


jakanakaevangeli:
To reproduce with emacs -q:

Press M-s M-w, type some random url and then use `embark-act` to open it with
`browse-url` (with `C-. M-x browse-url RET` for example).

This happens to all commands that use `browse-url-interactive-arg` in their
interactive spec because `(this-command-keys)` has zero elements.

oantolin:
I can't reproduce this. Using `M-x browse-url` does open the URL for me when used as an action from the minibuffer (I tried from `M-s M-w` and from `find-file`).

oantolin:
Oh, it does give an error if I try exactly the same thing from `emacs -Q`.

oantolin:
I bet this is one of those things that moving to `set-transient-map` would fix.

jakanakaevangeli:
Perhaps set--this-command-keys could be abused, so before
command-execute, we could do something like

    ;; Pacify commands that access (this-command-keys)
    (when (= 0 (length (this-command-keys)))
      (set--this-command-keys (kbd "RET")))

oantolin:
Sorry I took so long to fix this, @jakanakaevangeli. Thanks a lot for pointing out a solution. The only thing I did a little differently is try to use the key binding used to invoke the action if possible.

PR #750, opened 2025-02-25
fix: limit set--this-command-keys to browse-url

LemonBreezes:
My view after investigating this #470 and #182 is that the underlying cause is a bug in `browse-url-interactive-arg` itself, as it should handle `this-command-keys` being the empty string or vector. However, even if that is fixed, older versions of Emacs will still have the issue. So I propose limiting the use of `set--this-command-keys` to the functions with `browse-url` in their name.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -2070,12 +2070,15 @@ minibuffer before executing the action."
                             (embark--run-action-hooks embark-pre-action-hooks
                                                       action target quit)
                             (minibuffer-with-setup-hook inject
-                              ;; pacify commands that use (this-command-keys)
-                              (when (= (length (this-command-keys)) 0)
+                              ;; HACK work around `browse-url-interactive-arg'
+                              ;; expecting a non-empty `this-command-keys'
+                              ;; output.
+                              (when (and (eq action 'browse-url)
+                                         (= (length (this-command-keys)) 0))
                                 (set--this-command-keys
                                  (if (characterp last-command-event)
                                      (string last-command-event)
-                                  "\r")))
+                                   "\r")))
                               (setq this-command action)
                               (embark--run-around-action-hooks
                                action target quit)))


embark.el
```diff
@@ -2070,12 +2070,16 @@ minibuffer before executing the action."
                             (embark--run-action-hooks embark-pre-action-hooks
                                                       action target quit)
                             (minibuffer-with-setup-hook inject
-                              ;; pacify commands that use (this-command-keys)
-                              (when (= (length (this-command-keys)) 0)
+                              ;; HACK work around `browse-url-interactive-arg'
+                              ;; expecting a non-empty `this-command-keys'
+                              ;; output.
+                              (when (and (string-match-p "browse-url"
```
oantolin, line 2076: Symbols are much easier to compare for equality, I'd use (eq action 'browse-url).

LemonBreezes: Okay. Just changed it to that.

oantolin:
I cannot reproduce the original bug (#182) on Emacs 30.1, `(this-command-keys)` doesn't seem to be empty now, it returns `" "` instead. I think in the long run we can just delete the work around, and maybe for now limiting it to `browse-url` is a good idea.

oantolin:
@LemonBreezes What version of Emacs do you use can you reproduce #182?

LemonBreezes:
> @LemonBreezes What version of Emacs do you use can you reproduce #182?

Yes. It's because `this-command-keys` can return the empty string which gives an error with `aref`>

oantolin:
Right, I remember being able to reproduce the error with `this-command-keys` returning an empty string on Emacs 28.1. Now I cannot on Emacs 30.1. What version of Emacs are you using?

LemonBreezes:
> Right, I remember being able to reproduce the error with `this-command-keys` returning an empty string on Emacs 28.1. Now I cannot on Emacs 30.1. What version of Emacs are you using?

I can reproduce it on Emacs31.

oantolin:
Huh! Interesting! OK, so I guess I should just merge your pull request. I'll do a bit more testing and then merge.

LemonBreezes:
> Huh! Interesting! OK, so I guess I should just merge your pull request. I'll do a bit more testing and then merge.

Ok!

LemonBreezes:
> Huh! Interesting! OK, so I guess I should just merge your pull request. I'll do a bit more testing and then merge.

Any update?

oantolin:
Sorry for the delay, thanks for the PR!

Issue #753, opened 2025-04-27
Bug using embark with tmr package


eflanigan00:
Reference this issue here from tmr:  https://github.com/protesilaos/tmr/issues/5

When I try to run embark in the tmr mini-buffer don't get the tmr keys. Embark is trying to act on the number and not on the timer.

![Image](https://github.com/user-attachments/assets/06eda386-7aae-4af6-9e76-7c343c825abb)

minad:
It is not clear to me how to trigger the issue. Can you describe the exact key presses or commands you invoke? I suspect you are acting with Embark in an `tmr-tabulated-mode` buffer, which is not correct as I wrote in https://github.com/protesilaos/tmr/issues/5. This means you confuse the "mini buffer" with an actual TMR buffer.

eflanigan00:
tmr-tabluated-views the cursor is over the first timer, run embark-act

I'm fairly new to embark. I thought that embark could be a complete replacement for hydra. The benefit is that embark would prioritize the mode-map options in the pop-up.

For now I will use "M-X" which runs the command execute-extended-command-for-buffer and/or hydra.

It would be great if there could be a way to embark-act-mode which would call the popup and order the major mode selections.


You can see what happens when I call embark-act from ibuffer. Embark allows keys "p" and "n" to navigate in the buffer.  I don't know what the blue asterisk mean?

![Image](https://github.com/user-attachments/assets/bde8368a-4a2b-49af-b73c-62d82899fb3b)

minad:
> You can see what happens when I call embark-act from ibuffer. Embark allows keys "p" and "n" to navigate in the buffer. Although I think this is by accident acting on a list and not providing the ibuffer commands. I don't know what the blue asterisk means next to "p" and "n"?

Here Embark acts in Ibuffer on the text objects, e.g., words, functions or paragraphs. The `[ Default ]` title is recognized as Defun. I admit that this is not very useful here, but in other occasions it is, since you can put point on some symbol or url and then act on that item.

Embark usually avoids providing an alternative to the major mode bindings, since otherwise it would compete with them, and lead to layered complexity. The assumption is that going directly via the major mode commands is most efficient and worth for the user to learn. At least to me, avoiding such layering seems like a good design guideline. Now in certain cases there is a point in coherence, so this needs good judgement.

[...]

PR #757, opened 2025-05-27
Add embark-target-buffer-at-point

minad:
Works in ibuffer-mode and Buffer-menu-mode.

As discussed in #753

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -176,6 +176,7 @@ or a list of such symbols."
     embark-target-email-at-point
     embark-target-url-at-point
     embark-target-file-at-point
+    embark-target-buffer-at-point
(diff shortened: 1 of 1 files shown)


oantolin:
Mmh, can you add a test for `Buffer-menu-mode` for the buffer menu case? With the current code if you are in some tabulated list and the tabulated-list-id just randomly happens to be the name of some buffer the target finder triggers!

oantolin:
Oh, I thought `buffer-live-p` also returned true for strings that are names of buffers. You are correct that it only does so for actually buffer objects. In that case I agree that this is fine as is. Thanks, Daniel!

Issue #724, opened 2024-06-20
embark-prefix-help-command hides bindings for menu items


djr7C4:
This occurs for example with some `lsp-mode` bindings such as `lsp-rename`. This is bound by default in `lsp-command-map` to "rr" (`lsp-command-map` is bound to `lsp-keymap-prefix` in `lsp-mode-map`). These show "menu-item" instead of the key binding in `embark-prefix-help-command`. 

In other words, there will be a line when `embark-prefix-help-command` is run that shows

    menu-item    lsp-rename-maybe

when 

    r r    lsp-rename

is what I would prefer.

The reason is that `embark--formatted-bindings` contains the code

    (concat
     (if (eq (car-safe def) 'menu-item)
         "menu-item"
       (key-description key)))

in the loop that populates the `commands` variable.

Part of the issue is with `embark--command-name` which contains 

    ((eq (car-safe cmd) 'menu-item) (eval (cadr cmd)))

in the `cond`. I have this patched locally and could put together a pull request but I'm not sure what behavior is needed. I personally only care about the key bindings (and not menu items) but perhaps you want to preserve the menu items in `embark-prefix-help-command` so these should be split into two separate candidates.

bradhowes:
I think this is also related to an issue I encountered: executing `C-x 8 C-h` does not show the key translations when `prefix-help-command` is set to `embark-prefix-help-command` in my configuration. I am still trying to incorporate Embark into my workflow so this is not a big issue for me, but it did take me some time to figure out where in my configuration I had lost the full output of `C-x 8 C-h`.

PR #759, opened 2025-06-03
show keybinding instead of 'menu-item' in embark-bindings

### LINKED ISSUE #715: embark-prefix-help-command fails to show key bindings under lsp-mode
When using `(setq prefix-help-command #'embark-prefix-help-command)`, normally everything is fine, but for some reason in combination with `lsp-mode` (which uses a right-click popup menu), it doesn't show you the actual key bindings, but only "menu-item" (even though these items do have keybindings as well as being in the menu):

Skärmavbild 2024-04-13 kl  15 51 56

Whereas if you use the standard prefix-help-command, it succeeds in showing you the bindings:

Skärmavbild 2024-04-13 kl  15 53 01

I can't tell if it's `embark-bindings-in-keymap` that needs to be more clever, or if it's lsp-mode's `lsp-define-conditional-key` that does something in a nonstandard way.

dwa:
A "works-for-me" pull request which aims to fix #715, and seems to be in line with what is hinted at in #724.

Seems to work in my limited testing, no harm if others test it as well.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -1309,7 +1309,7 @@ first line of the documentation string; for keyboard macros use
    (cond
     ((or (stringp cmd) (vectorp cmd)) (key-description cmd))
     ((stringp (car-safe cmd)) (car cmd))
-    ((eq (car-safe cmd) 'menu-item) (eval (cadr cmd)))
+    ((eq (car-safe cmd) 'menu-item) (format "%S" (caddr cmd)))
     ((keymapp cmd)
      (propertize (if (symbolp cmd) (format "+%s" cmd) "<keymap>")
                  'face 'embark-keymap))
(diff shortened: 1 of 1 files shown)


minad:
@dwa Do you have a simple use case for testing this, something which doesn't involve large packages like lsp-mode, ideally starting with bare emacs -Q only?

minad:
Thanks. I think this is already helpful.

minad:
Okay, as far as I can tell, this PR does not improve things:

Before:
before

After:
after

minad:
> The main reason for the patch is to show keybindings from lsp-mode, so a no-change/no-worse on other entries should really be a good thing here.

Sure. But it is not a "no-change/no-worse" here. The "after" screenshot is less meaningful than the one from "before". My point is that we should not accept a change to the current logic if it is not universally better than before. Otherwise we are simply replacing the current hack with another hack.

> That said, I'm surprised to see that your "After" screenshot still show "menu-item" entries.

No, idea. But I think we have to dig a bit deeper on how to handle the menu-items properly.

PR #763, opened 2025-06-10
embark-consult--export-grep: Call compilation--ensure-parse

### LINKED ISSUE #762: grep-edit-mode does not work well in Embark-exported buffers
This issue is a bit special, so I want to write down here what I found for now. 

When exporting grep results from `consult-grep`, then invoking the new Emacs 31 `grep-edit-mode`, and then using `query-replace`, the replacement fails for matches outside of the window (as soon the window scrolls).

The stack trace for `occur-after-change-function` compiled:

~~~ elisp
Debugger entered--Lisp error: (wrong-type-argument markerp nil)
  (occur-after-change-function 2524 2527 3)
  (replace-match-maybe-edit "2.3" nil t nil (2524 2527 #<buffer *Embark Export: consult-ripgrep - #2.2*>) nil)
  (perform-replace "2.2" "2.3" t nil nil nil nil nil nil nil nil)
  (query-replace "2.2" "2.3" nil nil nil nil nil)
  (funcall-interactively query-replace "2.2" "2.3" nil nil nil nil nil)
  (command-execute query-replace)
~~~

And for `occur-after-change-function` interpreted:

~~~ elisp
Debugger entered--Lisp error: (wrong-type-argument markerp nil)
  (marker-buffer nil)
  (let* ((line-beg (line-beginning-position)) (targets (get-text-property line-beg 'occur-target)) (m (occur--targets-start targets)) (buf (marker-buffer m)) col) (if (and (get-text-property line-beg 'occur-prefix) (not (get-text-property end 'occur-prefix))) (progn (if (= length 0) (progn (put-text-property beg end 'occur-target targets) (save-excursion (and (search-forward "\n" end t) (delete-region ... end))))) (let* ((line (- (line-number-at-pos) (line-number-at-pos ...))) (readonly (save-current-buffer (set-buffer buf) buffer-read-only)) (win (or (get-buffer-window buf) (display-buffer buf ...))) (line-end (line-end-position)) (text (save-excursion (goto-char ...) (setq col ...) (buffer-substring-no-properties ... line-end)))) (let ((save-selected-window--state (internal--before-with-selected-window win))) (save-current-buffer (unwind-protect (progn ... ... ... ... ...) (internal--after-with-selected-window save-selected-window--state))))))))
  (save-excursion (goto-char beg) (let* ((line-beg (line-beginning-position)) (targets (get-text-property line-beg 'occur-target)) (m (occur--targets-start targets)) (buf (marker-buffer m)) col) (if (and (get-text-property line-beg 'occur-prefix) (not (get-text-property end 'occur-prefix))) (progn (if (= length 0) (progn (put-text-property beg end 'occur-target targets) (save-excursion (and ... ...)))) (let* ((line (- ... ...)) (readonly (save-current-buffer ... buffer-read-only)) (win (or ... ...)) (line-end (line-end-position)) (text (save-excursion ... ... ...))) (let ((save-selected-window--state ...)) (save-current-buffer (unwind-protect ... ...))))))))
  (occur-after-change-function 2241 2244 3)
  (replace-match-maybe-edit "2.3" nil t nil (2241 2244 #<buffer *Embark Export: consult-ripgrep - #2.2*<2>>) nil)
  (perform-replace "2.2" "2.3" t nil nil nil nil nil nil nil nil)
  (query-replace "2.2" "2.3" nil nil nil nil nil)
  (funcall-interactively query-replace "2.2" "2.3" nil nil nil nil nil)
  (command-execute query-replace)
~~~

The problem seems to be that the matches miss `occur-target` properties for matches outside of the displayed windows. These are usually not present in Grep buffers, but are added by `grep-edit--prepare-buffer`. Now I confirmed that `grep-edit--prepare-buffer` fails to add `occur-target` properties to the matches outside of the window.

Now the odd thing is that when I invoke `rgrep`, and start `grep-edit-mode`, then the `occur-target` properties are added correctly. But not for the Embark exported buffer from `consult-grep`.

minad:
Fix #762

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -210,6 +210,7 @@ This function is meant to be added to `embark-collect-mode-hook'."
 
 (defvar grep-mode-line-matches)
 (defvar grep-num-matches-found)
+(declare-function compilation--ensure-parse "compile")
 (declare-function wgrep-setup "ext:wgrep")
 
 (defvar-keymap embark-consult-rerun-map
@@ -232,7 +233,11 @@ The function FOOTER is called to insert a footer."
         (dlet ((compilation-filter-start (point)))
           (setq-local grep-num-matches-found (funcall insert lines))
           (goto-char (point-max))
-          ;; Emacs 30 feature `grep-use-headings'
+          ;; Ensure that all `compilation-message' text properties are added.
+          (compilation--ensure-parse (point))
+          ;; Instead of (run-hooks 'compilation-filter-hook), we only run the
+          ;; Emacs 30 grep heading filter. The other `compilation-filter-hook'
+          ;; functions handle escape sequences, which we do not need here.
           (when (and (bound-and-true-p grep-use-headings)
                      (fboundp 'grep--heading-filter))
             (grep--heading-filter))


minad:
Friendly ping @oantolin :)

oantolin:
I shouldn't have posponed merging this, since I never was really going to install Emacs 31 to test it. 😅  (I've tested wgrep still works on Emacs 30, at least.)

minad:
Thanks. I've switched to 31 a while ago. Since I compile myself anyway it does not make a big difference. Also it feels more attractive to maintain a bunch of my own patches and to try to upstream a few of them. It is only important for me that Emacs is stable enough. These days no big changes are in the pipeline, except of the new igc. Therefore master overall feels stable.

PR #764, opened 2025-06-20
Introduce `embark--as-file` to inject targets as file paths

minad:
- Introduce `embark--as-file` and `embark--associated-file` to inject targets as file paths.
- Bind "/", "x" and "$" uniformly to `embark-dired-jump`, `embark-open-externally` and `eshell` respectively.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -359,7 +359,10 @@ indicate that for files at the prompt of the `delete-file' command,
     (xref-find-definitions embark--ignore-target)
     (xref-find-references embark--ignore-target)
     (sort-regexp-fields embark--ignore-target)
-    (align-regexp embark--ignore-target))
+    (align-regexp embark--ignore-target)
+    ;; Treat target as file
+    (embark-dired-jump embark--as-file)
+    (embark-open-externally embark--as-file))
   "Alist associating commands with post-injection setup hooks.
 For commands appearing as keys in this alist, run the
 corresponding value as a setup hook after injecting the target
@@ -3836,14 +3839,6 @@ with command output.  For replacement behavior see
 (declare-function bookmark-prop-get "bookmark")
 (declare-function bookmark-completing-read "bookmark")
 
-(defun embark-bookmark-open-externally (bookmark)
-  "Open BOOKMARK in external application."
-  (interactive (list (bookmark-completing-read "Open externally: ")))
-  (embark-open-externally
-   (or (bookmark-prop-get bookmark 'location)
-       (bookmark-prop-get bookmark 'filename)
-       (user-error "Bookmark `%s' does not have a location" bookmark))))
-
 (defun embark-bury-buffer (buf)
   "Bury buffer BUF."
   (interactive "bBuffer: ")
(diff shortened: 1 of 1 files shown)


minad:
Okay, I kept `dired-jump` on `j` and moved `bookmark-jump` to `b`. Let's see if Omar agrees.

oantolin:
OK, after looking at the code I understand what this is about. I also noticed now that while the description accompanying the PR is very mysterious, @minad's first comment explains what it does, I could have sworn that comment wasn't there before (I only remember the comment's starting from @hmelman's).

oantolin:
How about this for your `F`, @minad?

```elisp
(defun embark-act-on-file (file)
  "Run `embark-act' on FILE."
  (interactive "FFile to act on: ")
  (with-temp-buffer
    (insert file)
    (let ((embark-target-finders '(embark-target-file-at-point)))
      (embark-act))))

(setf (alist-get 'embark-act-on-file embark-target-injection-hooks)
      '(embark--as-file))

(keymap-set embark-bookmark-map "F" #'embark-act-on-file)
```

(Here written as outside configuration, I'd of course but this in the right places in `embark.el`.)

oantolin:
I'm not sure my `embark-act-on-file` is such a good idea. An alternative approach to all of this would be to add additional file targets that triggers in case of bookmarks or other things we want to treat as files. I can't remember exactly if a transformer could do that: maybe rewriting `embark--as-file` as transformer works, but I can't remember if that can add an extra target or only replace one.

oantolin:
I'll merge this as is, but we should keep talking about `F`, which I think is a good idea, whether or not it is implemented via the above `embark-act-on-file`.

minad:
For me the "jump" and "open" operations (eshell, dired, xdgopen) are most important. The F menu would be nice to have, but it is not really crucial. Your implementation is quite a hack though, with the temporary buffer...

Regarding the technicalities here - I tried to use an around hook to transform to files, but they cannot transform the target, since the target injection uses the original target (is this a bug or missing feature?). I also tried transformers, but they transform the target in the very beginning and not per action, right? Then I went with a custom file injection hook which performs the transformation.

oantolin:
> For me the "jump" and "open" operations (eshell, dired, xdgopen) are most important. The F menu would be nice to have, but it is not really crucial. The F menu would be nice to have, but it is not really crucial.

Agreed.

> Your implementation is quite a hack though, with the temporary buffer...

I know, that's why I expressed hesitation about it. 😛

> Regarding the technicalities here - I tried to use an around hook to transform to files, but they cannot transform the target, since the target injection uses the original target (is this a bug or missing feature?).

I'm not sure I understand the exact issue here. How did you try to use an around hook for this?

>I also tried transformers, but they transform the target in the very beginning and not per action, right? Then I went with a custom file injection hook which performs the transformation.

Yes, that's right, they are not per action.

Also, completely unrelated to any of this: I'm finally using both Corfu and Cape! It didn't take long to get used to Corfu (i.e., to stop glancing down at the minibuffers when I ask for completions). Also, I thought I'd miss Embark while corfuing, but M-h covers 90% of what I wanted anyway. I occasionally miss embark-export but I can live without it. Also, Cape finally stuck with me, I'm not sure what was different about this attempt. I'm even using it instead of the TeX input method, which I was extremely used to.

PR #766, opened 2025-07-06
Prevent Tramp slowdown in Eldoc functions

minad:
Eldoc functions are called when moving point around. Some target finders use ffap which call file name handlers. This can lead to unpredictable latencies due to Tramp, which is not acceptable for point movement.


oantolin:
Huh, I've never noticed this! I guess I just don't have Tramp paths lying around in my buffers.

minad:
The problem occurs mainly when you edit remote buffers, since then ffap will search the remote `default-directory`. I am not sure if it happens if you have Tramp paths in your buffer, but it is likely. In any case, it is the right thing to completely disable Tramp for these Eldoc functions. If the reported targets are slightly wrong, not much is lost.

PR #769, opened 2025-07-26
Add target for dates

lmq-10:
First of all, thank you very much for the package! Here I add a simple target for dates following YYYY-MM-DD format. I find it very useful, specially as a starting point for custom functions, and I thought it could be of value for other people, too. I already assigned copyright to the FSF.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -133,6 +133,7 @@
     (environment-variables embark-file-map) ; they come up in file completion
     (url embark-url-map)
     (email embark-email-map)
+    (date embark-date-map)
     (buffer embark-buffer-map)
     (tab embark-tab-map)
     (expression embark-expression-map)
@@ -174,6 +175,7 @@ or a list of such symbols."
     embark-target-smerge-at-point
     embark-target-package-at-point
     embark-target-email-at-point
+    embark-target-date-at-point
     embark-target-url-at-point
     embark-target-file-at-point
     embark-target-buffer-at-point
@@ -936,6 +938,16 @@ centered at point that includes the current line)."
                (cons ,(or target '(match-string 0))
                      ,(or bounds
                           '(cons (match-beginning 0) (match-end 0)))))))))
+
+(defvar embark-iso-date-regexp
+  (rx (group (= 4 digit)) "-"
+      (group (= 2 digit)) "-"
+      (group (= 2 digit)))
+  "Regexp matching an ISO 8601 date string.")
+
+(embark-define-regexp-target date embark-iso-date-regexp nil
+                             (match-string-no-properties 0)
+                             nil 10)
 
 (defun embark--identifier-types (identifier)
   "Return list of target types appropriate for IDENTIFIER."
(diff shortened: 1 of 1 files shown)


hmelman:
I think I like this.  I'd add a command to compute and print the difference from the date to now (in a human readable form) and maybe the day of the week.  Perhaps a command to put the date into calc for more advanced calculations.  Currently I use shift-number to increment/decrement numbers, perhaps keys (with a repeat-map) to change Y, M, D values.

lmq-10:
@hmelman Thanks for the feedback! I just added functions for computing day difference and printing the date in a custom format. The other two suggestions are more complex. They can be done, but I'm sure they would imply more coding. We'll have to think whether they are better suited for embark.el or as suplementary code somewhere else.

lmq-10:
Hmm... Actually, I think this does the trick for Calc:

```elisp
(defun embark-send-date-to-calc (date)
  "Insert DATE (an ISO 8601 date string) into a calc window."
  (interactive "sDate (YYYY-MM-DD): ")
  (when (string-match embark-iso-date-regexp date)
    (require 'calc)
    (require 'calendar)
    (let ((abs (calendar-absolute-from-gregorian (embark--iso-date-to-calendar date))))
      (calc)
      (calc-push-list `((date ,abs))))))
```

But I ask for feedback because I never wrote code for it.

minad:
Hmm, I like the idea too but I am critical of the fact that we seemingly have to write all the actions ourselves here. This makes the addition feel a little bit ad-hoc. For many (most?) Embark target types we can simply reuse builtin Emacs commands as actions. The problem here is that the ISO date format is not recognized by default in text buffers in Emacs, unlike for example urls or mail addresses, which are supported by the thingatpt Emacs library. Maybe because there are too many competing date formats? Otoh jumping to the agenda or the calendar from a date is a natural thing to do.

1. Are there other date manipulation package we could rely on, or which could be integrated with Embark via a separate package? In particular the actions suggested by @hmelman feel like a separate package, and not like we are integrating existing functionality.
2. Would it be possible to reuse Org commands for date actions, such that we would have to write fewer custom Embark commands?
3. Maybe all the date-related functionality could be moved to an embark-date.el feature (or at least the embark-date-* namespace). It could still be part of the Embark package on ELPA.

@oantolin What do you think?

embark.el
```diff
@@ -4060,6 +4072,69 @@ It assumes the URL was encoded in UTF-8."
...
+(declare-function calendar-date-string "calendar")
+
+(defun embark-pretty-print-date (date)
+  "Echo a pretty representation for DATE.
+Format is defined by the variable `calendar-date-display-form'."
+  (interactive "sDate (YYYY-MM-DD): ")
+  (when (string-match embark-iso-date-regexp date)
+    (require 'calendar)
+    (message (calendar-date-string (embark--iso-date-to-calendar date)))))
+
```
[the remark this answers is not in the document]
lmq-10: Good catch. Done.

embark.el
```diff
@@ -4060,6 +4072,71 @@ It assumes the URL was encoded in UTF-8."
...
+(defun embark-date--to-calendar (date)
+  "Convert ISO 8601 DATE to calendar internal format."
+  (when (string-match embark-date--regexp date)
+    (let ((day (match-string 3 date))
+          (month (match-string 2 date))
+          (year (match-string 1 date)))
+      (list
+       (string-to-number month)
+       (string-to-number day)
+       (string-to-number year)))))
```
minad, line 4084: This function can also be simplified I think:

~~~ elisp
(calendar-gregorian-from-absolute (date-to-day date))
~~~

embark.el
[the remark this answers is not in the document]
minad: Thanks.

embark.el
```diff
@@ -4060,6 +4072,71 @@ It assumes the URL was encoded in UTF-8."
...
+  (require 'diary-lib)
+  (diary-check-diary-file)
+  (diary-list-entries (embark-date--to-calendar date) 1))
+
+(declare-function days-between "time-date")
+
+(defun embark-date-echo-difference (date)
+  "Echo day difference between current date and DATE."
+  (interactive (list (embark-date--read)))
+  (let ((day-difference (days-between date (format-time-string "%F"))))
```
minad, line 4123: The difference can be computed more efficiently as follows:

~~~ elisp
(- (date-to-day date) (time-to-days nil))
~~~

Not that it matters, but the code is also a little bit shorter...

~~~ elisp
(- (date-to-day date) (time-to-days nil))
(days-between date (format-time-string "%F")))
~~~

embark.el
lmq-10: It's also slightly less efficient though. What about this?

```elisp
(mapcar #'string-to-number (list month day year))
```

lmq-10: Done

embark.el
lmq-10: Done

lmq-10:
FWIW I just created a library for all things related to ISO dates:

- https://github.com/lmq-10/iso-date

Just in case you prefer these definitions outside this repo!

lmq-10:
@hmelman Fixed, thank you!

oantolin:
It gives me pause that all of the actions had to be written for this, though not too much pause since they are just a bit of plumbing to call built-in functions. The stuff that ships with Embark focuses on target types where most of the actions are built-in Emacs commands, usable as is or with small hooks. These date actions do seem very useful, but on balance I think they should probably be in a separate package. Are all the actions here also in your iso-date package? If so, I think the best thing would be to either add the embark configuration to that package or to make a separate embark-iso-date package.

PR #773, opened 2025-08-24
Handle new project-buffer category

minad:
https://github.com/emacs-mirror/emacs/commit/bb0ede711eb630e3cc4b02316bf76079ce760de6 https://github.com/minad/marginalia/commit/30e6813c8142ef8cb45e6f9bdd23ead1c80b9b2e


oantolin:
Shouldn't we also associate embark-buffer-map to project-buffer?

oantolin:
Sorry! I didn't think things through, in particular I didn't realize that `embark--uniquify-orig-buffer` always uses `buffer` as the target type; and thus I didn't realize this was enough.

PR #782, opened 2025-11-17
Fix export of apropos data

lwf:
The sort override was missing a few arguments. Side-step the whole process by adding varargs.


lwf:
I should maybe add a little more context for why this is necessary.

If you call describe-function/describe-variable/etc. interactively and attempt to export candidates from the minibuffer, the code in master will yield the following error:

```emacs-lisp
apply: Wrong number of arguments: #[(list _pred) ((nreverse list)) (t)], 5
```

The patch works around this by ignoring arguments irrelevant to our monkeypatch.

Swallowing additional arguments using varargs is likely to be more resistant to future Emacs changes. Allowing this looseness seems like a reasonable trade-off, given the hacky nature of relying on implementation details like this in the first place.

I have had this fix sitting in my config for close to a year and figured it was time to try to upstream it.

PR #788, opened 2025-12-23
Completion update

### LINKED ISSUE #787: Emacs 31: Completion list candidate can be selected even from minibuffer
Hello @oantolin!

If `minibuffer-visible-completions` is enabled, a candidate can be selected in the completions buffer directly from the minibuffer. This means that `embark-target-completion-list-candidate` should be adjusted accordingly. See https://github.com/minad/consult/commit/7be0d6bf556b608b0b38d006a4f66be849d721e0 for the related Consult commit. I can provide a patch if you like.

Merry christmas and all the best,
Daniel

minad:
Fixes #787. Please test carefully. You need `minibuffer-visible-completions` set to `t` in order to see the effect. This settings allows you to "remote" control the Completions buffer from the minibuffer. You can navigate with up and down keys from the minibuffer.

### DIFF
--- a/embark-consult.el
+++ b/embark-consult.el
@@ -90,7 +90,10 @@ category `consult-line'."
   (let ((buf (generate-new-buffer "*Embark Export Occur*"))
         (mouse-msg "mouse-2: go to this occurrence")
         (inhibit-read-only t)
+        (affixator (embark--get-affixator 'consult-location))
         last-buf)
+    ;; Run affixator for lazy highlighting
+    (setq lines (mapcar #'car (funcall affixator lines)))
     (with-current-buffer buf
       (dolist (line lines)
         (pcase-let*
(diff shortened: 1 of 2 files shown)


oantolin:
Oh, is this something I need to test on the current master? I'm using the released Emacs 30.1. Since I have now lost most of my fear of compiling Emacs from source I guess I could test this.

minad:
@oantolin Did you find time to look at this? As mentioned, the main question is if you think that the reordering of the target finder is correct. I have tested this for a while and I am confident that it is okay.

oantolin:
The `@comma{}`s are really irritating me for some reason. 😅 I assume the TeXinfo Org exporter changed in Emacs 31, and decided to just play it safe and always export a `,` as `@comma{}` because in some rare instances it's necessary? I don't like it. 😅

oantolin:
Initial testing seems good! I'll give it a little more time and try a few thing related to the target order.

 I do have one question: why only target the highlighted completion in the `*Completions*` when `minibuffer-visible-completions` is `t`? If I understand correctly it only affects key bindings, right? You can always navigate the `*Completions*` buffer from the minibuffer, when `minibuffer-visible-completions` is `nil` you use `M-<up>`, when it is `t` you use `<up>`. Why should embark care about that difference?

PR #804, opened 2026-05-02
Use Compat 31

minad:
compat-31.0.0.0 is available since 2026-May-02

https://elpa.gnu.org/packages/compat.html


oantolin:
Thank you so much! I'm glad they're adding `completion-table-with-metadata` (and to a lesser extent also `completion-list-candidate-at-point`) to Emacs.

minad:
> I'm glad they're adding completion-table-with-metadata (and to a lesser extent also completion-list-candidate-at-point) to Emacs.

Yes, I added both of these functions :)

PR #808, opened 2026-05-21
Treat computed space better in embark--display-string

minad:
See also https://github.com/minad/vertico/commit/9132ab8a7e294a04d411cbcd67902052210eb14d.

It would be good to keep these in sync.


minad:
Hmm, I have to rethink this a little more.

PR #809, opened 2026-05-21
Treat computed space better in embark--display-string

minad:
See also https://github.com/minad/vertico/commit/14cc82fc56b34482b33fb0175ec3044ab9a50003

It would be good to keep these in sync.

@oantolin I found this when working with computed space in Elfeed buffers. The problem is that align-to is relative to the left border by default, and since the candidates obtained by consult-line are reformatted with line numbers in front, the align-to space might collapse to a zero width space. Vertico and the Embark snapshot buffer are affected by this. Therefore always add a single space to make sure that the candidates do not look compressed. Compute space with :width does not have that problem. See https://github.com/emacs-elfeed/elfeed/commit/b92434b0b18b94402cb6dc1cbbcf92704f2e402e for reference.

I believe I had such zero space compression in other buffers with alignment before, when invoking consult-line. But I don't remember which mode was affected. So it seems good to do this, even if it is a bit of a hack.

### DIFF
--- a/embark.el
+++ b/embark.el
@@ -3100,6 +3100,7 @@ For non-minibuffers, assume candidates are of given TYPE."
                 (add-face-text-property
                  0 (length disp) face t (setq disp (concat disp))))
               (setq pos nextd chunks (cons disp chunks)))
+          (pcase disp (`(space :align-to . ,_) (push " " chunks)))
           (while (< pos nextd)
             (let ((nexti
                    (next-single-property-change pos 'invisible str nextd)))


oantolin:
I'm a little confused about why this affects embark collect, it uses tabulated-list for alignment. Isn't that different enough from Vertico (which, if I recall, has to assemble the entire candidate line) that it wouldn't be an issue here? Is there a simple completing-read call for which embark-collect shows the problem?

minad:
If the candidates contain align-to spaces the space may collapse to zero space due to the way the candidates are displayed in tabulated-list-mode or in Vertico. The patch only makes sure that we insert at least a single space for each displayed space. You can execute consult-line (with Vertico and then embark-collect) in elfeed-search buffers to see the difference. Also you can try Vertico stable vs the commit linked above.

oantolin:
Ah, great, elfeed did illustrate the problem right away, thanks.

minad:
> Ah, great, elfeed did illustrate the problem right away, thanks.

Maybe I can even win you back as an Elfeed user ;)

Thanks for merging. embark/vertico--display-string are just using a heuristic anyway, so it shouldn't hurt.

oantolin:
> Maybe I can even win you back as an Elfeed user ;)

Yeah, installing elfeed to see this problem really made me nostalgic for my elfeed-user days. As I think I've told you I only stopped using it because I am almost always away from my computer when I want to read RSS feed, which is why I prefer an RSS reader that works well on my phone. Oh, and I also remember it being a little awkward to sync the elfeed database between two computers (I remember syncing problems and having to manually fix the databases on rare occasions).

PR #810, opened 2026-05-30
Make embark-prefix-help-command work under which-key

LemonBreezes:
Ok. I have been using this code change for over a year in my config. It makes it so `C-h` in `which-key` brings up the Embark prefix help. Very useful and nice!

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,6 +1,13 @@
 #+title: Embark changelog
 
 * Development version
+- When you use =embark-prefix-help-command= as your =prefix-help-command=
+  and also have the =which-key= package on, the help key now invokes
+  Embark even while the =which-key= popup is showing, instead of
+  =which-key='s own paging menu. This also works when =which-key= was
+  summoned explicitly (e.g. via =which-key-show-top-level= or
+  =which-key-show-keymap=), in which case the keymap =which-key= is
+  displaying is described.
 - The target finders for outline headings now also work with heading
   determined by ~outline-search-function~, not just ~outline-regexp~ as
   previously. A new heading target finder specifically for eww buffers
--- a/embark.el
+++ b/embark.el
@@ -1948,6 +1948,71 @@ type @ and the key binding (without the prefix)."
                          (format " under %s" (key-description prefix)))))
       (embark-bindings-in-keymap keymap))))
 
+;;; Integration with which-key
+
+(declare-function which-key--popup-showing-p "ext:which-key")
+(declare-function which-key--current-key-string "ext:which-key")
+(declare-function which-key-C-h-dispatch "ext:which-key")
+(declare-function which-key--show-keymap "ext:which-key")
+(declare-function which-key-show-top-level "ext:which-key")
+
+(defvar embark--which-key-keymap nil
+  "Keymap that `which-key' is currently displaying, if any.
+Recorded by `embark--which-key-record-keymap' and
+`embark--which-key-record-top-level' so that
+`embark-which-key-C-h-dispatch' can describe that keymap when
+`which-key' is summoned explicitly rather than from an incomplete
+key sequence.")
+
+(defun embark--which-key-record-keymap (_keymap-name keymap &rest _)
+  "Record KEYMAP as the one `which-key' is currently displaying.
+Intended as `:before' advice for `which-key--show-keymap'."
+  (setq embark--which-key-keymap keymap))
+
+(defun embark--which-key-record-top-level (&rest _)
+  "Record the global map as the keymap `which-key' is displaying.
+Intended as `:before' advice for `which-key-show-top-level', which
+shows the top-level bindings via `which-key--create-buffer-and-show'
+and so never goes through `embark--which-key-record-keymap'."
+  (setq embark--which-key-keymap (current-global-map)))
+
+(defun embark-which-key-C-h-dispatch (oldfun &rest args)
+  "Make `embark-prefix-help-command' take effect under `which-key'.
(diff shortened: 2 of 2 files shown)


oantolin:
I'm not 100% sure but I think that the mere fact that the commit is coauthored by an LLM means I cannot accept it, since Embark is in GNU ELPA and they won't take any LLM-generated code. This is for copyright reasons: it seems likely that LLM code cannot be copyrighted, but anything in GNU ELPA needs to be copyrightable and have its copyright assigned to the FSF. I believe GNU hasn't made a final decision on this, but in the meantime GNU Emacs specifically is not taking any LLM-generated code contributions.

LemonBreezes:
> I'm not 100% sure but I think that the mere fact that the commit is coauthored by an LLM means I cannot accept it, since Embark is in GNU ELPA and they won't take any LLM-generated code. This is for copyright reasons: it seems likely that LLM code cannot be copyrighted, but anything in GNU ELPA needs to be copyrightable and have its copyright assigned to the FSF. I believe GNU hasn't made a final decision on this, but in the meantime GNU Emacs specifically is not taking any LLM-generated code contributions.

Oh, the LLM did not write the code. I had it in my config for over a year, before LLM was mainstream. I just asked the LLM to make a draft PR out of the code.

So the LLM did not write any code, it updated the README and copy-pasted my code then added whitespace, comments, and function documentation.

oantolin:
I don't know if that changes the situation, I'll ask around.

oantolin:
Okay, so the advice I received is that since you wrote the code, I can review the pull request in the usual way, but I should probably rewrite the change log entry to be safe.

minad:
FWIW I am not fond of adding a which-key integration here. I don't even have it installed in my Emacs. Any reason why this code cannot go as well into the wiki? Embark even offers a better alternative to which-key.

oantolin:
I used to feel the same way, @minad, but which-key comes with Emacs now, so I think which-key integration is reasonable (for people who don't recognize the superiority of embark-prefix-help-command 🤷). If which-key were not included with Emacs I would have immediately said "this should go in the wiki".

oantolin:
> I don't even have it installed in my Emacs.

How did you pull that off? 🙂

LemonBreezes:
I figured instead of PRing the shim into Doom Emacs that 99% of people using Emacs enable `which-key` so a little glue for an edge case made sense. Basically, if you call which-key with a command explicitly then type C-h to bring up Embark, you need this code.

oantolin:
Upon further reflection, it does feel a little odd to add support to switch from which-key to embark-prefix-help-command, which is an alternative to which-key. It feels similar to adding markdown syntax support to org mode or a command to vertico to switch to icomplete, which sound like strange things to do.

Maybe it would be better to make an embark-which-key package? You could also add the which-key indicator that's on the embark wiki to that package.

LemonBreezes:
> Upon further reflection, it does feel a little odd to add support to switch from which-key to embark-prefix-help-command, which is an alternative to which-key. It feels similar to adding markdown syntax support to org mode or a command to vertico to switch to icomplete, which sound like strange things to do.
> 
> Maybe it would be better to make an embark-which-key package? You could also add the which-key indicator that's on the embark wiki to that package. 

Maybe as a separate file like `embark-which-key` in this repo? Those which-key commands like `which-key-show-major-mode` I do use and I use them with embark + grid view too.

oantolin:
Yes, a separate embark-which-key.el file might make sense. I'll think about it a bit. What I use instead of which-key-show-major-mode is embark-bindings, which by default shows only commands bound in the local map and minor mode maps (i.e., commands from the major mode and enabled minor modes).

minad:
> How did you pull that off?

I use a stripped-down Emacs.

> I used to feel the same way, @minad, but which-key comes with Emacs now, so I think which-key integration is reasonable.

Well, I maintain my opinion that not every addition to Emacs is or was reasonable. The amount of old cruft in Emacs is staggering.

Given that we have a better alternative here I don't see why one should not promote and recommend that instead.

In particular I disagree with the way the addition is made here via globally installed advices. This does not look like a clean addition to me.

minad:
Note that there is also https://github.com/oantolin/embark/wiki/Additional-Configuration#use-which-key-like-a-key-menu-prompt. I think all of the which-key related code should go into a separate package, and not only minor integrations like this one. It could be part of this repository as embark-which-key.el, but then, Omar is not using which-key, so it won't go maintained, I guess?

LemonBreezes:
> Note that there is also https://github.com/oantolin/embark/wiki/Additional-Configuration#use-which-key-like-a-key-menu-prompt. I think all of the which-key related code should go into a separate package, and not only minor integrations like this one. It could be part of this repository as embark-which-key.el, but then, Omar is not using which-key, so it won't go maintained, I guess?

I can maintain it. I'm really awesome and I'm probably gonna use Emacs for another 60 years.

minad:
> I can maintain it. I'm really awesome and I'm probably gonna use Emacs for another 60 years.

That's good. But then the which-key integration should be "complete" and also contain this indicator stuff from the wiki. And it should be a separate file at least, or maybe a separate package.

To be honest, I had hoped that we could avoid the which-key discussion for a little longer, given that Embark still supports Emacs 29, so we would have to add a dependency here for which-key or make sure that it also compiles fine without which-key or distribute it as separate package or even put it in a separate repository.

oantolin:
I think I would prefer if you made a separate embark-which-key package, @LemonBreezes, for integration. I just added a which-key-like `embark-auto-prefix-help-mode`, so even more than before it feels like Embark is an alternative to which-key and thus an integration package between the two is sort of a niche concern.

PR #812, opened 2026-06-05
Support non-numeric prefix keys in `embark-prefix-help-command`

Stebalien:
Without this, prefix keys such as `<menu>` will signal an error.


oantolin:
Thank you!

